PR #16, opened 2015-04-08
TOOLS-703: MongoExport - exporting documents with double datatypes

andromediusEx:
Checks situation when float equals to int and prints zero signs after
point
This guarantee that json to bson conversion will be done correctly.
For example after bson to json conversion 1.0 should remain 1.0


sr527:
Hi Andrew, 
Thank you for submitting this pull request. Previously, mongoexport did not add a trailing .0 to whole numbers that are float64, but behavior was changed in https://github.com/mongodb/mongo-tools/commit/59191d86dfa45034a04aee5fcf19f01a6a0c6206. Now mongoexport exports a float64 with either a decimal point or as scientific notation, even with whole numbers.

PR #17, opened 2015-04-15
TOOLS-681 Added ability to import int values

Antonytm:
For some reason in Mongo 3+ behavior of mongoimport was changed. It
stopped to allow importing of int values, all values that are imported
became float64. This commit backs previous behavior and allows to import
either int or float.

### DIFF
--- a/common/bsonutil/converter.go
+++ b/common/bsonutil/converter.go
@@ -40,7 +40,7 @@ func ConvertJSONValueToBSON(x interface{}) (interface{}, error) {
 		}
 		return v, nil
 
-	case string, float64:
+	case string, float64, int:
 		return v, nil // require no conversion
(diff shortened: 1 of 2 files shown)


sr527:
Hi Anton,
Thank you for bringing up this issue. We have fixed this regression in mongoimport, which previously parsed any number that is not wrapped in an NumberInt() or NumberLong() as a float64. The new behavior is that un-wrapped numbers will parse as a float64 if there is a decimal point after the number or if it is in scientific notation, and as an int32 or int64 otherwise, depending on the size of the number. The commit where these changes were made are in https://github.com/mongodb/mongo-tools/commit/59191d86dfa45034a04aee5fcf19f01a6a0c6206.

PR #20, opened 2015-05-28
TOOLS-750: add the final count of dumped documents with mongodump

ramnes:
This does not resolve the issue since it doesn't explain why the first count is different than the final number of dumped documents, but at least it shows the difference and avoid user to run a `bsondump --objcheck` to know how many documents where dumped.

### DIFF
--- a/mongodump/mongodump.go
+++ b/mongodump/mongodump.go
@@ -48,6 +48,7 @@ type MongoDump struct {
 	authVersion     int
 	archive         *archive.Writer
 	progressManager *progress.Manager
+	count		int
 }
 
 // ValidateOptions checks for any incompatible sets of options.
@@ -417,6 +418,7 @@ func (dump *MongoDump) DumpIntent(intent *intents.Intent) error {
 
 	}
 
+	dump.count = 0
 	if dump.useStdout {
 		log.Logf(log.Always, "writing %v to stdout", intent.Namespace())
 		return dump.dumpQueryToWriter(findQuery, intent)
@@ -444,7 +446,8 @@ func (dump *MongoDump) DumpIntent(intent *intents.Intent) error {
 		return nil
 	}
 
-	log.Logf(log.Always, "done dumping %v", intent.Namespace())
+	log.Logf(log.Always, "done dumping %v (%v documents dumped)",
+		intent.Namespace(), dump.count)
 	return nil
 }
 
@@ -513,6 +516,7 @@ func (dump *MongoDump) dumpIterToWriter(
 			return fmt.Errorf("error writing to file: %v", err)
 		}
 		progressCount.Inc(1)
+		dump.count += 1
 	}
 
 	return nil
--- a/mongodump/mongodump.go
+++ b/mongodump/mongodump.go
@@ -48,7 +48,6 @@ type MongoDump struct {
 	authVersion     int
 	archive         *archive.Writer
 	progressManager *progress.Manager
-	count		int
 }
 
 // ValidateOptions checks for any incompatible sets of options.
@@ -418,7 +417,6 @@ func (dump *MongoDump) DumpIntent(intent *intents.Intent) error {
 
 	}
 
-	dump.count = 0
 	if dump.useStdout {
 		log.Logf(log.Always, "writing %v to stdout", intent.Namespace())
 		return dump.dumpQueryToWriter(findQuery, intent)
@@ -446,8 +444,7 @@ func (dump *MongoDump) DumpIntent(intent *intents.Intent) error {
 		return nil
 	}
 
-	log.Logf(log.Always, "done dumping %v (%v documents dumped)",
-		intent.Namespace(), dump.count)
+	log.Logf(log.Always, "done dumping %v", intent.Namespace())
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


mongodump/mongodump.go
```diff
@@ -513,6 +516,7 @@ func (dump *MongoDump) dumpIterToWriter(
 			return fmt.Errorf("error writing to file: %v", err)
 		}
 		progressCount.Inc(1)
+		dump.count += 1
```
ramnes: I would have prefered to make something like `_, dump.count = progressCount.Progress()` just below rather than incrementing two values at the same time, but the `Progress` method isn't accessible with a `progress.Updateable`. I actually wonder: why is there is two different `progress` interfaces @3rf ?

ultrabug:
:+1: thanks !

btall:
thanks :+1:

Lujeni:
GoGo ! Thx

mongodump/mongodump.go
```diff
@@ -48,6 +48,7 @@ type MongoDump struct {
 	authVersion     int
 	archive         *archive.Writer
 	progressManager *progress.Manager
+	count		int
```
3rf: Unfortunately, making this a struct field for the entire tool won't work. Mongodump dumps collections concurrently, and having a single variable is not thread safe. To scope this more safely to a single goroutine, let's:

1. Make some changes to the progress bar interfaces (see below).
2. Change `func (dump *MongoDump) dumpQueryToWriter` to return a count of documents dumped along with an error.
3. Grab that final count from the progress tracker.

mongodump/mongodump.go
3rf: You're right that this could be better. The two interfaces are left over from a design plan that ended up changing. At this point, I'd recommend merging both interfaces into one (just keep the name "Progressor"). This might require some very minor updates for the other tools as well.

3rf:
Guillaume,
We appreciate the work you've done on this and would be happy to accept this as a community fix. I've left a few comments to help make this change more complete. Let me know if you have any questions!

ramnes:
That's really not much. I'll make the modifications as you commented on the code as soon as possible. Thanks!

ramnes:
Hi Kyle,

I've updated the code as requested. I've tried to keep it as much clean as possible so I've decided to print the result directly inside `func (dump *MongoDump) dumpIterToWriter` rather than making it return multiple values.

I've also deleted the `Updateable` interface, and it doesn't seem that it was used anywhere else in the repository.

Keep me up to date!

3rf:
Thanks for getting this updated so quickly and for dealing with the interface change! I like that you're trying to keep things simple, but I think we might have to change the `dumpIterToWriter` return values in other to print the count next to its collection. Let me know if I can help with that any further.

-Kyle

ramnes:
I don't know much about Go and even less about its concurrency, but as I can read the code it is supposed to return in chain between the first and the second message; I don't really see how the two messages could be separated from each other. Could you explain me more?

Also, an other solution would be to give `intent` to `dumpIterWriter` so that we can access `Namespace` and print everything here. What do you think?

Hokutosei:
thanks :+1:

3rf:
First off, I'd like to apologize for a typo above. I meant to say we should modify `dumpQueryToWriter` to return the number written (i.e. `(written int, err error)`), which we could do by grabbing the count of the progressor. Modifying `dumpIterWriter` bumps up on our special case for repair cursor iterators, which we do not need the intent data for. Additionally, we use these helpers for restoring auth and oplog collections, which would add lots of unneeded and confusing log output. It would be better to just bubble the count up and only log it for regular collections.

Second, the issue with concurrency is that each collection is restored in a separate goroutine (you can think about this as a thread). In practice, execution order is pretty predictable, but we can run into some weird race conditions if we make the logs separate messages. For instance, say we are restoring 2 collections in parallel, in thread A and B. We could have a situation where:
```
Thread A writes "dumped 10000 documents" then sleeps
Thread B writes "dumped 500 documents"
Thread B writes "finished dumping test.threadB"
Thread A wakes up and writes "finished dumping test.threadA"
```
The log output would look like
```
dumped 10000 documents
dumped 500 documents
finished dumping test.threadB
finished dumping test.threadA
```
Looking at the above output, it is impossible to tell which thread has which count without further context.

ramnes:
Hi Kyle,

This last commit should answer your concerns about possible race conditions.

I've put the logging into `dumpQueryToWriter` because it seemed simpler to handle to me than changing every return values when the function is called. I hope you won't see anything wrong about it.

3rf:
Hi Guillaume,
Unfortunately I had to implement this ticket myself last week due to time constraints. We were running up on two weeks without a reply and needed this feature fixed urgently. 

I'd like to sincerely apologize--not communicating the time constraints and not closing this pull request before starting my work was a huge lack of oversight, and I'm sorry for wasting your time with the recent patch set. In the future we will be much more proactive about updating GitHub issues and pull requests.

I really appreciate your interest in contributing to the tools project. If you're still interested in working on something, feel free to browse through our current open tickets ( https://jira.mongodb.org/browse/TOOLS ) and see what looks feasible!

Sorry again,
Kyle

ramnes:
No big deal, but yeah, it would have been better to communicate a bit more. :+1: 

Out of curiosity, where can I see your implementation?

PR #22, opened 2015-05-28
TOOLS-663 TOOLS-667 support compressed files in dump directories

gabrielrussell:


mongorestore/filepath.go
```diff
@@ -23,13 +24,21 @@ const (
...
+}
+
 // realBSONFile implements the intents.file interface. It lets intents read from real BSON files
 // ok disk via an embedded os.File
 // The Read, Write and Close methods of the intents.file interface is implemented here by the
 // embedded os.File, the Write will return an error and not succeed
 type realBSONFile struct {
-	*os.File
+	io.ReadCloser
+	errorWriter
```
deafgoat: i'd add documentation explaining why we need this.

PR #28, opened 2015-07-21
TOOLS-834: Return a context-specific error for invalid escape sequences.

visemet:
Makes it so that the error "invalid character 'o' in string escape code" is returned instead of "invalid escape character" when given the regular expression `/f\o\o/`.

https://jira.mongodb.org/browse/TOOLS-834


mpobrien:
This is merged

PR #30, opened 2015-07-31
TOOLS-853: exposing signals handler and progress manager for customisation/removal in mongodump

htch:
See https://jira.mongodb.org/browse/TOOLS-853


zachjs:
https://jira.mongodb.org/browse/TOOLS-853 has been resolved.

PR #35, opened 2015-08-29
TOOLS-896: Add --headerline option for mongoexport

coding-hello:
Add an option --headerline to control whether the csv file export with field list at the first line.

### DIFF
--- a/mongoexport/mongoexport.go
+++ b/mongoexport/mongoexport.go
@@ -308,9 +308,11 @@ func (exp *MongoExport) exportInternal(out io.Writer) (int64, error) {
 	log.Logf(log.Always, "connected to: %v", connURL)
 
 	// Write headers
-	err = exportOutput.WriteHeader()
-	if err != nil {
-		return 0, err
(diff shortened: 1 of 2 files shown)


gabrielrussell:
ZhuZiqiang,

Thank you for this work. It's a good start.

There are two problems that I see with this solution. The first of which is that this changes default behavior. The second of which is that it disables the call to the WriteHeader() function when the output format is JSON. The JSON WriteHeader function is actually used for something, so we still need to call it.

A solution for this will requite these things:

* Don't change default behavior.
* Don't break non-csv output.
* Error if someone specifies the --HeaderLine flag with any format other then csv.
* lastly, unit tests will need to updated to check for correct functionality.

It may be much easier and simpler if you were to implement this as a --noHeaderLine flag which would disable calling WriteHeader only when the output mode is csv.

Let me know if you need any guidance.

Thank you,
Gabriel Russell

coding-hello:
Hi gabrielrussell,

Thank your for your suggestion. It is very useful for me.  I'll implement an --noHeaderLine flag and commit again.

PR #36, opened 2015-08-31
Support precise oplog replay

hustchensi:
precise `ts` GTE tailling is needed for data migrating


gabrielrussell:
I'm not entirely sure of what problem you're trying to solve here, but any changes to how oplog replay work would need to go through a more rigorous process. If there is a problem that this solves, could you please open a ticket up on https://jira.mongodb.org/ and describe the problem in detail there.

PR #52, opened 2015-12-23
TOOLS-1028 mongostat json has wrong mmapv1-specific fields

yomegui:
The json formatter doesn't display these fields when the engine is WiredTiger:
* flushes
* ar|aw
* qr|qw

But these fields are present in grid formatter and [documentation](https://docs.mongodb.org/manual/reference/program/mongostat/) doesn't mention that these fields are mmapv1 specific.

This patch fixes that.


mpobrien:
Change looks good, merged this and also backporting to the 3.0 branch as well. Thank you @yomegui !

Issue #56, opened 2016-01-22
Gopass changed a method


SommerEngineering:
The library gopass has changed the return value of a method:
https://github.com/howeyc/gopass/blob/master/pass.go

`getPasswd()` returns now the data and an error. Thus, the mongo-tools does not compile any longer, caused by `common/password/pass_util.go`'s `GetPass()`.

SommerEngineering:
I have created a pull request with the fix.

gabrielrussell:
Since we vender Gopass, this will get fixed when we revender it.

gabocic:
Hi gabrielrussell,

Do you have an estimation of when this issue will get fixed?

thanks in advance

gabrielrussell:
Gabocic,

Are you currently having trouble building a version of the tools? I don't know of any actual problem currently with building any released version of the tools.

gabocic:
gabrielrussel,

Still seeing the same problem:

(myserver:~/gopath) go get github.com/mongodb/mongo-tools/common/password
# github.com/mongodb/mongo-tools/common/password
src/github.com/mongodb/mongo-tools/common/password/pass_util.go:20: multiple-value gopass.GetPasswd() in single-value context

gabrielrussell:
I can't really tell what you're trying to do there. If you want to build the tools, you should follow the instructions here: https://github.com/mongodb/mongo-tools#setup, and here: https://github.com/mongodb/mongo-tools#building-tools

ys-roposo:
gabriellrussell,
  [multiple-value gopass.GetPasswd() in single-value context when installing #36](https://github.com/ParsePlatform/flashback/issues/36)

  building flashback fails, as mentioned by tredman.

gabrielrussell:
Applications often have dependencies. Generally not all versions of the software is compatible with all versions of dependencies. They way that we solve this in our project is to vendor in the dependencies that we know our application is compatible with.

The application you're trying to build needs a way to manage its dependencies on versions of its dependencies.

If you have trouble building _our_ application, using the provided instructions then feel free to let me know.

gabrielrussell:
mongo-tools ships with a version of gopass that its compatible with. We build a GOPATH with a script in the root of the project called set_gopath.sh, that lets us build against the included dependencies. While the application that you're tying to build may not be able to run set_gopath.sh directly, you it may be able to incorporate the knowledge withing and build against the correct vendored dependencies.

ys-roposo:
[...]

PR #57, opened 2016-01-22
Change regard new gopass release

SommerEngineering:
Change to meet the new gopass release. gopass.GetPasswd() now returns
data and error. Fixes issue #56!

### DIFF
--- a/common/password/pass_util.go
+++ b/common/password/pass_util.go
@@ -17,5 +17,9 @@ func IsTerminal() bool {
 }
 
 func GetPass() string {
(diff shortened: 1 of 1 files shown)


gabrielrussell:
Since we vendor github.com/howeyc/gopass, this change won't be needed until we revender the new version.
As it is now, this patch causes a compile error:
````
.gopath/src/github.com/mongodb/mongo-tools/common/password/pass_util.go:20: assignment count mismatch: 2 = 1
````

I suspect that we'll go through the process of revendoring all of the dependencies at some point before the spring.

PR #59, opened 2016-03-13
add %dirty and %used cache for json in WTOnly

garnaud:
% dirty and % used are present in GridLineFormatter but not in JSONLineFormatter.


lucasem:
This went away when we refactored mongostat in 330310c.

PR #61, opened 2016-03-29
TOOLS-1120 Add option targetDb to mongorestore when use option archive

ArthurHlt:
When we use the option archive it could be nice to de-archive just one database inside another database which have a different name.

The actual use case is this one:
1. I have a mongo database created by a SaaS, this service give one database
2. I dump this database as an archive (Archive is used because of the necessity to dump it as a stream using stdout)
3. I want to restore my database to a new created database created from SaaS wich have a different database name than the original one
4. I'm blocked cause it's not possible with `--archive` option when restoring, it wants to create a database with the name from the original database which is not possible to use

Behavior of the option `--targetDb` in mongorestore:
- this can only be used when using `--archive` option
- This option make mongorestore to restore just one database at once
- If you have more than one database in archive and user use `--targetDb` it will fail, in this case it will ask to set also the option `--db`
- else mongorestore will take the only database set in archive and restore it in the database set in `--targetDb`

For the test, i didn't any test cause I don't see ad to add them, it seems to have no canvas to make this test. It should be in `mongorestore_test.go` but there is no relevant place for this because `--archive` option doesn't seem to be tested

### DIFF
--- a/mongorestore/filepath.go
+++ b/mongorestore/filepath.go
@@ -234,6 +234,9 @@ func (restore *MongoRestore) CreateAllIntents(dir archive.DirLike, filterDB stri
 	if err != nil {
 		return fmt.Errorf("error reading root dump folder: %v", err)
 	}
+	if filterDB == "" && restore.OutputOptions != nil && restore.OutputOptions.TargetDB != "" && len(entries) > 1 {
+		return fmt.Errorf("You can use --targetDb without --db only when there is one database in archive. Please use --db option.")
+	}
 	for _, entry := range entries {
 		if entry.IsDir() {
 			if err = util.ValidateDBName(entry.Name()); err != nil {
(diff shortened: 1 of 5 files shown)


bompus:
+1 to have this fixed. Didn't fully review the PR, but the core problem needs to be solved either way.

lucasem:
We've implemented a solution to this, via [TOOLS-1234](https://jira.mongodb.org/browse/TOOLS-1234) documented by [DOCS-8118](https://jira.mongodb.org/browse/DOCS-8118). This added namespace inclusion, exclusion, and renaming support. These new CLI options are prefixed with `ns` because they take a namespace pattern (where asterisks must be escaped for literal use). For example:

```sh
# restore everything, but rename prod db to staging
$ mongorestore --nsFrom 'prod.*' --nsTo 'staging.*' --archive='my_archive'
# restore only prod db, and rename it
$ mongorestore --nsInclude 'prod.*' --nsFrom 'prod.*' --nsTo 'staging.*' --archive='my_archive'
# restore everything but the prod db
$ mongorestore --nsExclude 'prod.*' --archive='my_archive'
```

These changes are committed at 77d72c8.

PR #73, opened 2016-07-28
TOOLS-69 add new option "--mode=merge"; change --upsert to --mode=upsert.

yinrong:


zachjs:
Hello and thank you for your pull request. We had a discussion regarding your changes and the original feature request. We'd like to add these changes without complicating our command line interface. Ideally `--upsert` could be modified to take an argument, but that would not be backwards compatible. So a better option may be to add a new, more ubiquitous argument `--mode=<opt>`. `--mode=insert` would be the default, `--mode=upsert` would provide the existing upsert behavior, and `--mode=merge` would trigger the merge import functionality you have added. Use of the `--upsert` argument should then be deprecated and alias to `--mode=upsert`.
As a point of style, statements in the form of
```
if (...) {
    ...
} else {
    if (...) {
        ...
    } else {
        ...
    }
}
```
may be more clearly written as
```
if (...) {
    ...
} else if (...) {
    ...
} else {
    ...
}
```
Finally, please take a look at `test/qa-tests` — our workflow requires that tests be written for any new features.

Thanks again for your proposal and let us know if we can be of any help!

yinrong:
thanks for your comment. I would like to change the code as you say

yinrong:
@zachjs 
sorry I don't how to get qa-tests to run:

```
18:46:09-~/dev/mongo-tools/test/qa-tests (master)$ python buildscripts/setup_multiversion_mongodb.py ./install ./link "OSX/x86_64" "2.4" "2.2" "2.6"
Traceback (most recent call last):
  File "buildscripts/setup_multiversion_mongodb.py", line 290, in <module>
    main()
  File "buildscripts/setup_multiversion_mongodb.py", line 285, in main
    downloader.download_version(version)
  File "buildscripts/setup_multiversion_mongodb.py", line 55, in download_version
    url, full_version = self.gen_url(version)
  File "buildscripts/setup_multiversion_mongodb.py", line 172, in gen_url
    % (version, self.links))
Exception: Cannot find a link for version 2.4, versions {} found.
18:46:16-~/dev/mongo-tools/test/qa-tests (master)$ python buildscripts/smoke.py bson export files import oplog restore stat top
Traceback (most recent call last):
  File "buildscripts/smoke.py", line 1451, in <module>
    main()
  File "buildscripts/smoke.py", line 1358, in main
    set_globals(options, tests)
  File "buildscripts/smoke.py", line 1109, in set_globals
    raise Exception("no mongod found in this directory.")
Exception: no mongod found in this directory.
```

zachjs:
Unless you are running certain tests for `mongorestore`, you should be able to run the tests without running the multiversion script. Also note that on `master` we now primarily use the `resmoke.py` script to run our tests, rather than the older `smoke.py`.

Before running the tests, be sure to run `./build.sh` inside of the root of the mongo-tools repo. Then, inside of `test/qa-tests`, run 
```
PATH=/path/to/mongo-tools/bin:$PATH python buildscripts/resmoke.py --executor=core --mongod=/path/to/mongod --mongo=/path/to/mongo jstests/import/<test>.js
```
(replacing the `/path/to`s with the appropriate paths).

Thanks, and please let me know if you need further assistance!

yinrong:
tested with command:
```sh
for f in jstests/import/*.js; do PATH=../../bin:$PATH python buildscripts/resmoke.py --executor=core --mongod=~/local/bin/mongod --mongo=~/local/bin/mongo $f || break; done
```

and all passed.

mongoimport/mongoimport.go
```diff
@@ -502,8 +522,10 @@ func (up *upserter) Insert(doc interface{}) error {
 	var err error
 	if selector == nil {
 		err = up.collection.Insert(document)
-	} else {
+	} else if up.imp.IngestOptions.Mode == MODE_UPSERT {
 		_, err = up.collection.Upsert(selector, document)
+	} else { // mode == 'merge'
+		_, err = up.collection.Upsert(selector, bson.M{"$set": document})
```
yinrong: please help check if bson.M is correct

test/qa-tests/jstests/import/upsert.js
```diff
@@ -14,49 +14,75 @@
...
+    assert.eq(db.c.count(), 2);
+
+    var one_expect = {_id: "one", a: "unicorns", b: "zebras"};
+    if (option === "--mode=merge") {
+      one_expect["x"] = "unique field";
+    }
+    assert.docEq(db.c.findOne({_id: "one"}), one_expect);
+  }
+
+  testWithUpsertFields("--upsert"); // deprecated option
```
yinrong: six test cases in this file

zachjs:
Thank you very much for your work thus far. This is definitely in the right direction. I do have a few small comments which I will post on the diff itself. Besides those things, great work!

mongoimport/mongoimport.go
```diff
@@ -27,6 +27,13 @@ const (
 	JSON = "json"
 )
 
+// Modes accepted by mongoimport.
+const (
+	MODE_INSERT = "insert"
```
zachjs: Standard constants like this should be formatted in camel case to follow the convention in Go's libraries. (e.g. `modeInsert = ...`)

test/qa-tests/jstests/import/mode.js
```diff
@@ -0,0 +1,88 @@
...
+
+    assert.docEq(doc2_1, doc2_1_expect);
+    assert.docEq(doc2_2, doc2_2_expect);
+    assert.eq(ret, 0);
+  }
+
+  function testWithoutUpsertFields(option) {
+    // This works by applying the update using _id
+    db.c.drop();
+    db.c.insert({_id: "one", a: "original value"  , x: "unique field"});
```
zachjs: The two spaces before the second comma on this line cause our JS linter to fail. Our linting settings are specified in `.eslintrc` in the root of this repo. While they do make the code look better, they should just be removed.

mongoimport/mongoimport_test.go
```diff
@@ -229,7 +229,7 @@ func TestMongoImportValidateSettings(t *testing.T) {
 			So(err, ShouldBeNil)
 			imp.InputOptions.HeaderLine = true
 			imp.InputOptions.Type = CSV
-			imp.IngestOptions.Upsert = true
+			imp.IngestOptions.Mode = "upsert"
```
zachjs: You should use `modeUpsert` here and in the lines bellow. Once they constants are defined, they should be used wherever possible.

mongoimport/mongoimport.go
zachjs: In this case, use `bson.D{{"$set", document}}` (with the extra curly braces).

zachjs:
Even though `--mode=insert` is the default, it would be nice if an execution that explicitly specifies `--mode=insert` was added to one of the jstests.

mongoimport/options.go
```diff
@@ -55,8 +55,12 @@ type IngestOptions struct {
 	// Forces mongoimport to halt the import operation at the first insert or upsert error.
 	StopOnError bool `long:"stopOnError" description:"stop importing at first insert/upsert error"`
 
-	// Modifies the import process to update existing objects in the database if they match --upsertFields.
-	Upsert bool `long:"upsert" description:"insert or update objects that already exist"`
+	// Modifies the import process,
+	// "upsert": replace existing objects in the database if they match --upsertFields.
+	// "modify": similar as "upsert', but modify existing object rather than replace them
+	Mode string `long:"mode" value-name:"<mode>" default:"insert" description:"upsert: insert or replace existing objects. merge: modify existing objects rather than replace them. defaults to insert"`
```
zachjs: Use `choice:"insert" choice:"upsert" choice:"merge"` to restrict this option to those choices.

mongoimport/options.go
```diff
@@ -55,8 +55,12 @@ type IngestOptions struct {
...
+
+	Upsert bool `long:"upsert" description:"(deprecated; same as --mode=upsert) insert or update objects that already exist"`
```
zachjs: Given that this is deprecated, you should add `hidden:"true"` so it isn't show in the help text.

mongoimport/options.go
zachjs: Also, if you use `choice`s you should remove the `value-name`.

mongoimport/options.go
```diff
@@ -55,8 +55,12 @@ type IngestOptions struct {
 	// Forces mongoimport to halt the import operation at the first insert or upsert error.
 	StopOnError bool `long:"stopOnError" description:"stop importing at first insert/upsert error"`
 
-	// Modifies the import process to update existing objects in the database if they match --upsertFields.
-	Upsert bool `long:"upsert" description:"insert or update objects that already exist"`
+	// Modifies the import process,
+	// "upsert": replace existing objects in the database if they match --upsertFields.
```
zachjs: Also document the behavior of the default insert mode.

mongoimport/mongoimport.go
```diff
@@ -154,17 +161,30 @@ func (imp *MongoImport) ValidateSettings(args []string) error {
...
+	} else if !(imp.IngestOptions.Mode == MODE_INSERT ||
+		imp.IngestOptions.Mode == MODE_UPSERT ||
+		imp.IngestOptions.Mode == MODE_MERGE) {
+		return fmt.Errorf("invalid --mode argument: %v", imp.IngestOptions.Mode)
+	}
+
 	if imp.IngestOptions.UpsertFields != "" {
-		imp.IngestOptions.Upsert = true
+		if imp.IngestOptions.Mode == MODE_INSERT {
+			imp.IngestOptions.Mode = MODE_UPSERT
```
gabrielrussell: This should be actually be an error. Something along the lines of "--upsertFields requires a --mode=upsert or --mode=merge"

yinrong: --upsertFields can be used alone before. are you sure to throw error here?

if throws error here, someone should encounter a failure if they use --upsertFields alone.

mongoimport/mongoimport.go
yinrong: if I change the source to 
```go
_, err = up.collection.Upsert(selector, bson.D{{"$set": document}})
```

compilication error is
```sh
./mongoimport.go:530: invalid field name "$set" in struct initializer
```

I don't quite understand. bson.M to wrap {$set:...} is the only way I found working right now

yinrong: http://stackoverflow.com/questions/24904129/golang-mongodb-mgo-driver-upsert-upsertid-documentation

mongoimport/mongoimport.go
gabrielrussell: Ok, fair enough. 

The problem with the code as it is is that if you were to specify --upsertFields "foo" --mode=insert, then you'd silently ignore their explicit request to do inserts.

I think what you need to do is to remove the default value for the mode option, so that you can detect the difference between not specifying the option, and specifying "insert"

yinrong:
@zachjs done

mongoimport/mongoimport.go
```diff
@@ -502,8 +526,12 @@ func (up *upserter) Insert(doc interface{}) error {
 	var err error
 	if selector == nil {
 		err = up.collection.Insert(document)
-	} else {
+	} else if up.imp.IngestOptions.Mode == modeInsert {
+		// skip if document exists
```
zachjs: If we are in insert mode, the selector must have been nil, so this block is unnecessary. It may make sense to change the else block to an else if that checks if it's in merge mode.

zachjs:
You should also rebase your commit on the latest changes we've pushed to master.

yinrong:
go test mongoimport/

resmoke qa-tests/import

eslint qa-tests/import

all passed, except qa-tests/import/parse_grace.js fails before I make this change:

```
diff --git a/test/qa-tests/jstests/import/parse_grace.js b/test/qa-tests/jstests/import/parse_grace.js
index 90c1bb6..ea96a36 100644
--- a/test/qa-tests/jstests/import/parse_grace.js
+++ b/test/qa-tests/jstests/import/parse_grace.js
@@ -57,7 +57,7 @@
   // parseGrace=fail should cause a failure
   var ret = toolTest.runTool.apply(toolTest, ["import", "--file",
     "jstests/import/testdata/parse_grace.csv",
-    "--writeConcern", "{w: 1, j: 1, fsync: 1}",
+    "--writeConcern", "{w: 1, fsync: 1}",
     "--type", "csv",
     "--db", "testdb",
     "--collection", "testcoll",
@@ -72,7 +72,7 @@
   // with an un-coercable field
   ret = toolTest.runTool.apply(toolTest, ["import", "--file",
     "jstests/import/testdata/parse_grace.csv",
-    "--writeConcern", "{w: 1, j: 1, fsync: 1}",
+    "--writeConcern", "{w: 1, fsync: 1}",
     "--type", "csv",
     "--db", "testdb",
     "--collection", "testcoll",
@@ -89,7 +89,7 @@
   // of the row
   ret = toolTest.runTool.apply(toolTest, ["import", "--file",
     "jstests/import/testdata/parse_grace.csv",
-    "--writeConcern", "{w: 1, j: 1, fsync: 1}",
+    "--writeConcern", "{w: 1, fsync: 1}",
     "--type", "csv",
     "--db", "testdb",
     "--collection", "testcoll",
@@ -106,7 +106,7 @@
   // parseGrace=autoCast should import the un-coercable field
   ret = toolTest.runTool.apply(toolTest, ["import", "--file",
     "jstests/import/testdata/parse_grace.csv",
-    "--writeConcern", "{w: 1, j: 1, fsync: 1}",
+    "--writeConcern", "{w: 1, fsync: 1}",
     "--type", "csv",
     "--db", "testdb",
     "--collection", "testcoll",
```

however my commit does not include this.

mongoimport/mongoimport.go
```diff
@@ -500,10 +524,12 @@ func (up *upserter) Insert(doc interface{}) error {
 	document := doc.(bson.D)
 	selector := constructUpsertDocument(up.imp.upsertFields, document)
 	var err error
-	if selector == nil {
+	if selector == nil { // modeInsert || doc-not-exist
```
yinrong: I add comment here

mongoimport/mongoimport.go
```diff
@@ -500,10 +524,12 @@ func (up *upserter) Insert(doc interface{}) error {
...
 		err = up.collection.Insert(document)
-	} else {
+	} else if up.imp.IngestOptions.Mode == modeUpsert {
 		_, err = up.collection.Upsert(selector, document)
+	} else { // modeMerge
```
yinrong: This is unsafe alone, but I add unit-tests & qa-tests

mongoimport/mongoimport_test.go
```diff
@@ -223,36 +223,55 @@ func TestMongoImportValidateSettings(t *testing.T) {
 			So(imp.ValidateSettings([]string{}), ShouldBeNil)
 		})
 
+		Convey("an error should be thrown if --mode is incorrect", func() {
```
yinrong: new test

test/qa-tests/jstests/import/mode.js
```diff
@@ -0,0 +1,147 @@
...
+    default:
+      throw new Error();
+    }
+    assert.docEq(docs, docExpects);
+  }
+
+  // argument-1: mode expected
+  // argument-2: command argument for mongoimport
+
+  testWithUpsertFields("error", "--mode=wrong");
```
yinrong: new test

test/qa-tests/jstests/import/mode.js
```diff
@@ -0,0 +1,147 @@
...
+  testWithUpsertFields("error", "--mode=insert");
+  testWithUpsertFields("upsert", "");
+  testWithUpsertFields("upsert", "--upsert"); // deprecated cmdArg
+  testWithUpsertFields("upsert", "--mode=upsert");
+  testWithUpsertFields("merge", "--mode=merge");
+
+  testWithoutUpsertFields("error", "--mode=wrong");
```
yinrong: new test

test/qa-tests/jstests/import/mode.js
```diff
@@ -0,0 +1,147 @@
...
+    default:
+      throw new Error();
+    }
+    assert.docEq(docs, docExpects);
+  }
+
+  // argument-1: expected behavior
+  // argument-2: command argument for mongoimport
+
+  testWithUpsertFields("error", "--mode=wrong");
```
yinrong: new test

yinrong:
@zachjs please help review again

mongoimport/mongoimport.go
```diff
@@ -154,17 +161,34 @@ func (imp *MongoImport) ValidateSettings(args []string) error {
...
 		imp.upsertFields = []string{"_id"}
 	}
 
-	if imp.IngestOptions.Upsert {
+	if imp.IngestOptions.Mode == "" {
+		imp.IngestOptions.Mode = modeInsert
+	} else if !(imp.IngestOptions.Mode == modeInsert ||
+		imp.IngestOptions.Mode == modeUpsert ||
+		imp.IngestOptions.Mode == modeMerge) {
+		return fmt.Errorf("invalid --mode argument: %v", imp.IngestOptions.Mode)
```
lucasem: This condition is unreachable, because of the go-flags choice restriction, and can safely be removed.

mongoimport/options.go
```diff
@@ -55,11 +55,18 @@ type IngestOptions struct {
...
+	// "insert": Insert only, skip exisiting documents.
+	// "upsert": Insert new documents or replace existing ones.
+	// "merge": Insert new documents or modify existing ones; Preserve values in the database that are not overwritten.
+	Mode string `long:"mode" choice:"insert" choice:"upsert" choice:"merge" description:"insert: insert only. upsert: insert or replace existing documents. merge: insert or modify existing documents. defaults to insert"`
+
+	Upsert bool `long:"upsert" hidden:"true" description:"(deprecated; same as --mode=upsert) insert or update objects that already exist"`
 
 	// Specifies a list of fields for the query portion of the upsert; defaults to _id field.
-	UpsertFields string `long:"upsertFields" value-name:"<field>[,<field>]*" description:"comma-separated fields for the query part of the upsert"`
+	UpsertFields string `long:"upsertFields" value-name:"<field>[,<field>]*" description:"comma-separated fields for the query part when --mode=upsert/merge"`
```
lucasem: Because `--mode=upsert/merge` is not actually valid, it may be preferable to say `--mode is set to 'upsert' or 'merge'`

lucasem:
Other than two small nits, this looks good to me (and @zachjs). Those test failures are unrelated to your changes (you can see the status of tests on the `master` branch [here](https://evergreen.mongodb.com/waterfall/mongo-tools)). If you rebase again, those tests should be passing, but you don't really need to — I'll test your changes on our supported platforms before merging.

yinrong:
@lucasem done

lucasem:
Thanks for your contribution @yinrong!

PR #74, opened 2016-08-30
TOOLS-1419: add support for noCursorTimeout flag

ar7z1:
[JIRA ticket](https://jira.mongodb.org/browse/TOOLS-1419).

Do I need to create a PR for [docs repo](https://github.com/mongodb/docs)?

### DIFF
--- a/common/db/command.go
+++ b/common/db/command.go
@@ -12,6 +12,7 @@ const (
 	Snapshot = 1 << iota
 	LogReplay
 	Prefetch
+	NoCursorTimeout
 )
 
 type NodeType string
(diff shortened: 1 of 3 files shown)


gabrielrussell:
No, you just mark that the jira ticket "needs documentation", and then add draft documentation to the ticket.

gabrielrussell:
It was decided in https://jira.mongodb.org/browse/TOOLS-1419 that we won't be solving this problem this way. But instead we'll try to fix it in https://jira.mongodb.org/browse/SERVER-6036.

PR #81, opened 2016-12-13
[TOOLS-1555] record EOF when connection closes

robertclancy:
- send EOF recordedOp when connection closes.
- send EOF only for non-empty streams.

### DIFF
--- a/mongoreplay/mongo_op_handler.go
+++ b/mongoreplay/mongo_op_handler.go
@@ -361,5 +361,18 @@ func (bidi *bidi) streamOps() {
 		// inform the tcpassembly that we've finished with the reassemblies.
 		stream.done <- nil
 	}
+	var lastOpTimeStamp time.Time
+	if bidi.streams[0].opTimeStamp.After(bidi.streams[1].opTimeStamp) {
+		lastOpTimeStamp = bidi.streams[0].opTimeStamp
+	} else {
+		lastOpTimeStamp = bidi.streams[1].opTimeStamp
+	}
+	if !lastOpTimeStamp.IsZero() {
+		bidi.opStream.unorderedOps <- RecordedOp{
+			Seen:              &PreciseTime{lastOpTimeStamp.Add(time.Nanosecond)},
+			SeenConnectionNum: bidi.connectionNumber,
+			EOF:               true,
+		}
+	}
 	bidi.logvf(Info, "Connection %v: finishing", bidi.connectionNumber)
 }
--- a/mongoreplay/play.go
+++ b/mongoreplay/play.go
@@ -240,7 +240,7 @@ func Play(context *ExecutionContext,
 	repeat int,
 	queueTime int) error {
 
-	sessionChans := make(map[string]chan<- *RecordedOp)
+	sessionChans := make(map[int]chan<- *RecordedOp)
 	var playbackStartTime, recordingStartTime time.Time
 	var connectionID int64
 	var opCounter int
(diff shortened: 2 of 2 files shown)


robertclancy:
I believe this fixes the referenced issues, bar the fact that the opcode is used to determine an identifier for the connection:
https://github.com/mongodb/mongo-tools/blob/df7e8e24e63441f37285be714fa1c7b30d7fa1e5/mongoreplay/play.go#L280-L284

Is there a reason for not using `op.SeenConnectionNum`? or maybe `op.SeenConnectionNum` and `op.Generation` to identify the connection?

gabrielrussell:
Robert,

I suspect that you've identified the cause of TOOLS-1555, but I think that TOOLS-1499 may not be completely solved by this.

Typically, a bug and fix of this magnitude would merit a corresponding test.

I don't quite understand how connection identifiers used in play.go relate to the patch. Can you explain? 

Gabriel

mongoreplay/mongo_op_handler.go
```diff
@@ -312,20 +312,37 @@ func (bidi *bidi) handleStreamStateOutOfSync(stream *stream) {
...
+		case reassemblies, ok = <-ch1:
 			reassembliesStream = 1
+			if !ok {
+				ch1 = nil
+			}
 		}
+		stream := bidi.streams[reassembliesStream]
 		if !ok {
-			break
+			if !stream.opTimeStamp.IsZero() {
```
gabrielrussell: I think there should be just one EOF per bidirectional connection, so you can replace the break and move this EOF op generation to after the loop block.

robertclancy:
I agree this needs a test 👍 .

The question about the connection identifiers arises because the resulting replay records do not have an op attached, this means that play.go does not know whether the operation came from the server, or whether it came from the client. play.go uses the op to determine whether to use the connection string, or the reversed connection string to identify the connection in a map from string -> session. I was wondering was there a reason why the "connection string" was used in that map to identify a connection and not the `seenConnectionNumber`?

mongoreplay/mongo_op_handler.go
```diff
@@ -312,20 +312,37 @@ func (bidi *bidi) handleStreamStateOutOfSync(stream *stream) {
 // stream's and tries to create whole protocol messages from them.
 func (bidi *bidi) streamOps() {
 	bidi.logvf(Info, "starting")
-	for {
+	ch0 := bidi.streams[0].reassembled
+	ch1 := bidi.streams[1].reassembled
+	for ch0 != nil || ch1 != nil {
```
robertclancy: good point. I'll remove this change.

mongoreplay/mongo_op_handler.go
robertclancy: The reason I have explicit closes going both ways right now is to get around the fact that these "close" recordedops always use the connectionString (and not the reversed connection string) in play.go. Having closes for both directions means that it is guaranteed to close the connection.

However I'd be happy to make this just one "close" recordedop, if play.go used something like the connectionNumber to identify the session?

gabrielrussell:
Ok, I understand that issue now. I'll look at it again.

mongoreplay/mongo_op_handler.go
robertclancy: actually I remember now why this is necessary. It's so that we do generate two EOFs for the two directions. This is related to the play issue.

robertclancy:
Also, I misread TOOLS-1499. I don't think its explained by this.

gabrielrussell:
I agree with your idea that using op.SeenConnectionNum and op.Generation to look up the session would be better and would make the EOF generation and handling simpler. Feel free to update this patch with that change.

gabrielrussell:
If you're interested in writing the tests, I could offer you some guidance or direction, but if you're not interested in writing the test, I'd likely have some time next week to write it.

robertclancy:
@gabrielrussell I've updated the play command to use the connection number. I didn't include the generation because the play command only sends through the EOF for the final generation.

I had a look at writing a test but it wasn't clear to me where I should be making changes. Some guidance on how to write a test asserting that an EOF should be seen in the op stream would be great.

billybanfield:
The commit in this PR is used here: 02386dcb5b25d1908fcfc24df604444c6a081b3c

PR #84, opened 2016-12-30
TOOLS-1684: Mongodump: Don't create or .Close() the SessionProvider if one is already provided

timvaillancourt:
In this (merged) commit mongodump.MongoDump now exposes the field 'SessionProvider': https://github.com/mongodb/mongo-tools/commit/5ae956b366c0f9fe78302ac8f95857f8f1ac0dac#diff-ba031cdd7da2cf37217b3529bc49a3da, however if you provide one, the .Init() function will overwrite it.

This commit moves to making this logic conditional, ie: if SessionProvider is defined then we DO NOT overwrite it, similar to 'OutputWriter'. There is also a boolean flag to disabled the deferred .Close() on the SessionProvider if the session provider was provided from outside.

EDIT: if it's cleaner, I could always just move the SessionProvider code up to https://github.com/mongodb/mongo-tools/blob/master/mongodump/main/mongodump.go, like mongorestore has done - passing the provider 'in' instead of creating it. No preference


billybanfield:
Hey Tim, this change looks really good. I prefer your suggested method of creating the session provider object in main and closing it there as well, similar to how mongorestore does it. 

Also, please file a ticket against the [mongodb tools jira project](https://jira.mongodb.org/browse/TOOLS/?selectedTab=com.atlassian.jira.jira-projects-plugin:summary-panel) and prefix this PR with the JIRA issue number, in the form TOOLS-1234 <MESSAGE>

timvaillancourt:
Thanks @billybanfield. https://jira.mongodb.org/browse/TOOLS-1684 created.

timvaillancourt:
Would it make sense to update mongodump/main.go to use this new approach or is it best to make two sets of changes @billybanfield?

billybanfield:
Thanks @timvaillancourt ! Perfectly fine to make the other set of changes here.

timvaillancourt:
Great @billybanfield. Last one: now that I think of it, I made this change to support both approches but if I move mongodump/main.go to use the new approach perhaps the backwards compatibility is just extra complexity. Would it make sense to only support the "new" approach of passing SessionProvider in?

billybanfield:
@timvaillancourt Seems totally reasonable to me to remove the backwards compatibility

billybanfield:
@timvaillancourt Let me know if you need any other input on this issue.

timvaillancourt:
Sorry for the delay @billybanfield. I've updated this to work exactly like mongorestore here:

https://github.com/mongodb/mongo-tools/blob/master/mongorestore/main/mongorestore.go#L68-L73

I've copied the error text mongodump was using for session-creation errors up to main/mongodump.go but I've moved it to a log.Logvf() instead of fmt.Errorf() to match mongorestore. I'm not sure if that is ok to move from log->fmt.

xdg:
Closing: this no longer applies after the switch to the new Go driver.

PR #96, opened 2017-05-18
TOOLS-936: Add option to allow mongorestore to build indexes in the foreground

sindbach:
Default to always build background indexes. Specify option `--allowForegroundIndexBuild` to build indexes according to the `*.metadata.json` file.


jaraco:
The title of this PR doesn't match the description. Suggest s/force/allow/.

gabrielrussell:
Do you need this backported to 3.4, or just in 3.6? Because this includes a change of default behavior, I wouldn't want to backport it to 3.4. If you need background index builds in 3.4, the default should still be to do foreground builds.

jaraco:
As long as 3.6 tools can work with older but recent releases of mongod, I see no issue with only including this with 3.6.

sindbach:
@gabrielrussell Just in 3.6 would be great. Thanks.

H-Max:
I think it would be great to add an option to build indexes in the background prior to 3.6, and make it defaults starting in 3.6.

(Just my 2 cents)

PR #104, opened 2017-10-05
TOOLS-1806: Don't include default values in option descriptions since they are echoed by default

stennie:


tfogo:
Hmmm... I've messed this up somehow, I'm just going to open a new PR with your changes: #170

PR #110, opened 2018-02-20
TOOLS-1954 enhanced csv array notation

calebHankins:
These changes add support for the csv array notation that mongoexport uses.  A header of foo.bar.0, value of 'baz' should now become `"foo": { "bar": ["baz"] } `

The live version will create: 
` "foo": { "bar": { "0": "baz" }   }`

There was also an issue when upserting using csv files where getUpsertValue would always return nil. The addition of the type switch on line 224 should address this issue. 

Additional examples can be found here: https://jira.mongodb.org/browse/TOOLS-1954

Please feel free to reach out to me with any questions or comments. Thanks!

### DIFF
--- a/mongoimport/common.go
+++ b/mongoimport/common.go
@@ -221,11 +221,14 @@ func getUpsertValue(field string, document bson.D) interface{} {
 	if subDoc == nil {
 		return nil
 	}
-	subDocD, ok := subDoc.(bson.D)
-	if !ok {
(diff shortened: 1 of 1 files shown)


tfogo:
Hi @calebHankins! Thank you for this PR also. Sorry we weren't able to use this one either. But TOOLS-1954 has been implemented in PR https://github.com/mongodb/mongo-tools/pull/175.

PR #111, opened 2018-02-21
TOOLS-1956 Add Bulk Upsert and Remove Modes and increase batch size limit

calebHankins:
The below changes were implemented after consulting with our Mongo rep Anant Srivastava to meet internal implementation needs. I will be opening a pull request shortly with our changes for review in case some/all of these changes want to be rolled into the product.

Bulk upserts
Enable bulk upsert operations. In the live version of mongoimport, running in upsert mode limits to 1 insertion worker process and an effective batch size of 1. This results in performance that unfortunately rendered mongoimport not viable for our volumes. With the addition of bulk, multi-worker upserts, we are seeing a 400-700X performance boost. With this performance tweak, mongoimport became a viable tool for our update process.

--bulkUpdate command line option added. When toggled on, upserts can be executed in bulk and in multiple worker processes. This option was added to limit the impact to existing processes using mongoimport. There is some debate on whether this flag is necessary or if 'bulkUpdate' mode should be 'on' by default and toggled 'off' via the --maintainInsertionOrder option

The change for 'bulkUpdate' upsert mode was implemented through disabling maintainInsertionOrder, removing the restriction for 1 insertion worker and adding new method to BufferedBulkInserter to support bulk Upsert operations.

Remove mode
--mode remove option added. Will construct bson selectors using records from input file and --upsertFields to remove matching documents. Each selector will remove only a single matching document. Implemented through adding new method to BufferedBulkInserter to support bulk Remove operations.

--upsertFields are required when specifying this option.

batchSize limit increased from 1k to 100k
With the MongoDB 3.6 batch size limit changes, the --batchSize option's maximum was raised to 100k documents. Mongoimport and mongo driver code (gopkg.in/mgo.v2) were patched to support this. Specifying a batch size larger than 1000 and targeting MongoDB <3.6 results in operations being batched driver side in chunks of 1000. The driver was also patched to split write operations >16MB into separate writeOpCommand calls for *insertOp, bulkUpdateOp, and bulkDeleteOp operation types.

https://docs.mongodb.com/manual/reference/limits/#Write-Command-Batch-Limit-Size

See https://jira.mongodb.org/browse/TOOLS-1956 for more details.

### DIFF
--- a/common/db/buffered_bulk.go
+++ b/common/db/buffered_bulk.go
@@ -8,6 +8,7 @@ package db
 
 import (
 	"fmt"
+
 	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 )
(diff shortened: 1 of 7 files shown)


tfogo:
Hi @calebHankins! Thank you for submitting this PR. Since some time passed before we started working on this issue, your PR got out of sync with master so we were unable to use it. [TOOLS-1956](https://jira.mongodb.org/browse/TOOLS-1956) and [TOOLS-2268](https://jira.mongodb.org/browse/TOOLS-2268) were resolved by PR https://github.com/mongodb/mongo-tools/pull/151 and PR https://github.com/mongodb/mongo-tools/pull/172 respectively.

PR #113, opened 2018-04-06
TOOLS-1997 mongostat needs to handle large counters encoded as floats

mzraly:
This is a minor change to display differences and rates computed from counter values larger than 1<<30, which get stored as doubles instead of signed integers.

### DIFF
--- a/mongostat/status/readers.go
+++ b/mongostat/status/readers.go
@@ -8,6 +8,7 @@ package status
 
 import (
 	"fmt"
+	"math"
 	"regexp"
 	"sort"
 	"time"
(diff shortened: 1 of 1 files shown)


xdg:
Could you please add a check for the `float64` case to ensure that the pre-conversion value fits within `math.MinInt64` and `math.MaxInt64`?  If, for some reason, the float doesn't fit, there should be an error rather than a cast.

mzraly:
Certainly, will try to get to that this evening.

xdg:
Thanks.  Just FYI, I'm not going to get a chance to merge this for a while as I'm going away to a conference and then have some urgent ticket to address before the end of the month.

tfogo:
I'm going through stale PRs to try to clean them up. Sorry this PR has been stale for such a long time.

Unfortunately this fix is not a priority for the Tools team at the moment. I appreciate the work you put into this PR, but I think we're going to have to close it since we don't have any current plans to fix this.

PR #120, opened 2018-12-05
TOOLS-2166: outputs the correct connected address for --uri

sindbach:
Update for --uri to output connection to : 

```
connected to: replicasetName/host1,host2,host3
```
Or, if it's standalone:

```
connected to: host1
```


TedTuckman:
@sindbach It looks good. Before we can merge it in, can you write a test or two to make sure its getting the output you're expecting? There's an example of how to redirect the logger to a buffer to check in common/log/tool_logger_test.go.

sindbach:
@TedTuckman This time updated with tests. I extracted the logic into a method and test that instead checking buffer output.

TedTuckman:
Looks good on evergreen, going to merge this in.

sindbach:
Thanks @TedTuckman

PR #122, opened 2019-02-05
V3.4 master

godevop:
couldn't find the set_gopath.sh file in the clone.
How i could install it in ubuntu 18.04?
screenshot from 2019-02-05 17-06-40


xdg:
Not sure how you wound up submitting a PR along with a question.  I'm going to close this PR so it's off  my PR screen, but I'm happy to help diagnose the problem you're having.

Looks like you're having problems with the change in the build system.  Our build process has changed to allow us to standardize across operating systems.  A couple things you need:

* Have the source checked out into a path that resembles a GOPATH -- i.e. `src/github.com/mongodb/mongo-tools`
* Run `build.sh` -- this will build all the tools into a `bin` directory.  You can pass tags like 'ssl' to `build.sh` to add ssl support.
* If you want to build individual tools only, look at how `build.sh` does it.  In particular, source `set_goenv.sh`, then call `set_goenv`.  Then look at how the `go build` line in `build.sh` sets `ldflags` and tags using helper functions.

I hope that helps!

PR #124, opened 2019-02-21
TOOLS-1840 Migrate mongofiles to the new Go driver

patrickfreed:
[TOOLS-1840](https://jira.mongodb.com/browse/TOOLS-1840)

This PR ports mongofiles to the new go driver. It currently is WIP, since it needs some bug fixes that are only present in later versions of the go driver.

The options parsing will change a lot before this is merged due to Go driver API changes in later versions. That said, in this patch when parsing options that are both present in the URI and command line, preference is given to the URI (this is how the version of the go driver mongofiles is built with behaves). Personally, I think it would make most sense for the command line args to prevail over URI parameters (despite what the specifications say for drivers). Prior to this port, mongofiles would error if competing options were set, which I think we should definitely do away with.

Another change is that the md5 hash of the file is no longer computed. This is a deprecated part of the GridFS spec that was still being done by mgo but has no presence in the new go driver. To preserve old functionality, we could compute it in mongofiles, but that would require buffering all the data, which the go driver does anyways.

I believe there was a bug where Write Concern settings in the URI were ignored that has been fixed as part of this. A similar one was recently found in mongoexport.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/mongo-tools/124)


mongofiles/options.go
```diff
@@ -22,6 +31,53 @@ Possible commands include:
...
+	// set WriteConcern
+	if storageOpts.WriteConcern != "" {
+		opts.WriteConcern, err = db.PackageWriteConcernOptsToObject(storageOpts.WriteConcern, nil)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error parsing --writeConcern: %v", err)
+		}
+	}
+
+	// set ReadPreference
+	var readPref *readpref.ReadPref
```
kmahar: nit: is there any reason to use a separate variable rather than just directly setting opts.ReadPreference? (like you did above with WriteConcern.)

mongofiles/mongofiles.go
```diff
@@ -57,185 +62,281 @@ type MongoFiles struct {
...
-// ValidateCommand ensures the arguments supplied are valid.
-func (mf *MongoFiles) ValidateCommand(args []string) error {
-	// make sure a command is specified and that we don't have
-	// too many arguments
-	if len(args) == 0 {
-		return fmt.Errorf("no command specified")
-	}
+// Struct representing the metadata associated with a GridFS files collection document.
+type gfsFileMetadata struct {
+	ChunkSize   int           	   `bson:"chunkSize"`
```
kmahar: nit: line up the bson labels

mongofiles/mongofiles.go
```diff
@@ -57,185 +62,281 @@ type MongoFiles struct {
...
-		if len(args) == 1 {
-			mf.FileName = ""
-		} else {
-			mf.FileName = args[1]
-		}
-	case Search, Put, Get, Delete:
-		if len(args) > 2 {
-			return fmt.Errorf("too many positional arguments")
+// Write data to GridFS Upload Stream. If this file has not been written to before, this function will open a new stream that must be closed.
+// Note: if this file already exists, the chunks written here will be orphaned when close is called and an error will be returned.
```
kmahar: so the write call will succeed but close will later fail if the file already exists? that seems a bit unexpected. is this just a limitation of how the GridFS API works?

kmahar: > I just checked to see what the current mongofiles behavior is, and it seems it actually _deletes_ the old chunks after failing to insert into the chunks collection, which is a serious bug.

by the old chunks, do you mean the ones from the file that already exists? 😱 

I don't think it is worth checking ahead of time for duplicates. from the user perspective, the behavior is really the same, it's just that we would potentially identify the problem like three lines earlier in `MongoFiles.put`.

mongofiles/mongofiles.go
```diff
@@ -57,185 +62,282 @@ type MongoFiles struct {
...
 
-	if mf.StorageOptions.GridFSPrefix == "" {
-		return fmt.Errorf("--prefix can not be blank")
+	return file.downStream.Read(buf)
+}
+
+// Deletes the corresponding GridFS file in the database and its chunks.
+// Note: this file must be closed if it had been written to before being deleted. Any download streams will be closed as part of this deletion.
+func (file *gfsFile) Delete() error {
+	if file.upStream != nil {
```
kmahar: this would only happen if we ourselves forget to close it, right?

mongofiles/mongofiles.go
kmahar: is there a test checking behavior when you try to write to an existing file without using the `replace` option? given the bug you described before, I think it would be worth adding one.

kmahar:
regarding options that are present in both the command line and the URI, is it that crazy to error if an option is set in both places? as a user I'm really not sure what I would expect to happen if I set a conflicting write concern.

mongofiles/mongofiles.go
patrickfreed: >by the old chunks, do you mean the ones from the file that already exists? 😱

Yeah. I filed [TOOLS-2229](https://jira.mongodb.com/browse/TOOLS-2229) for that.

>is there a test checking behavior when you try to write to an existing file without using the `replace` option? given the bug you described before, I think it would be worth adding one.

I added one and in doing so discovered why the docstring was written the way it was. The go driver actually has a bug where the `{files_id: 1, _id: 1}` index is not _unique_, as the spec requires. So we actually are getting orphan chunks in this case. Filed [GODRIVER-856](https://jira.mongodb.com/browse/GODRIVER-856) for that one.

I'm going to hold off on merging the test into this patch though, as I'll likely implement it when I fix TOOLS-2229 (which will come before this port is done).

mongofiles/mongofiles.go
patrickfreed: Yep.

mongofiles/mongofiles.go
kmahar: all sounds good 👍

kmahar:
ohhh, I see. I was imagining a case where you put w:1 in the URI and w:2 in the options field or something. I agree that is annoying.

mongofiles/options_test.go
```diff
@@ -7,76 +7,64 @@
...
-		opts.URI.AddKnownURIParameters(options.KnownURIOptionsWriteConcern)
+			storageOpts := &StorageOptions{}
+			inputOpts := &InputOptions{}
+			_, opts, err := ParseOptions(args, storageOpts, inputOpts)
 
-		Convey("Parsing with no value should leave write concern empty", func() {
-			_, err := opts.ParseArgs([]string{})
 			So(err, ShouldBeNil)
-			So(storageOpts.WriteConcern, ShouldEqual, "")
-			Convey("and building write concern object, WMode should be majority", func() {
```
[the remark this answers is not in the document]
patrickfreed: Related question: should we be explicitly setting majority as the default? Or should we be using the server's default when the user doesn't supply anything? I know in the drivers we're not supposed to send a write concern if the user doesn't supply one, but does that guidance extend to the tools?

>The string "majority" only exists on the left, delete code. Where are we testing that the default write concern is majority?

I've added an integration test to replace the deleted unit one that verifies the default behavior, and I did discover that in this patch, we are in fact ignoring write concern options set in the URI. I've made a slight (temporary) modification so that they aren't ignored any more, but that causes the default case to not send a write concern and use the MongoDB default instead (i.e. {w: 1}). Once we bump the go driver dependency, I can use the new options merging api to set "majority" as the forced default if that is what's desired.

Because we're no longer parsing the URI ourselves, the write concern information will be on the client/db/collection as set by the go driver rather than in our options types, hence why it must be an integration test.

Also, so that the arguments parsing is more easily testable, I added a `New` constructor function to `mongofiles` and introduced a new `Options` struct that encompasses all the different options types that can be used to configure mongofiles. This way, our integration tests can run by passing the same command-line arguments that a user would instead of manually constructing options types. I've found that the old integration tests combined with the unit tests are not really effective ways of testing the options parsing. The new test I added is an example of the new style. If you think it's the right approach, I will convert the rest of the integration tests, either in a later patch for this PR or in follow-up work. (I've already implemented a similar design in my wip [mongorestore port](https://github.com/patrickfreed/mongo-tools/blob/ffda33c3551f548bb6d924a82c4b30f1df4507f7/mongorestore/mongorestore_test.go#L91))

mongofiles/mongofiles.go
```diff
@@ -57,185 +62,282 @@ type MongoFiles struct {
...
-	Id          bson.ObjectId `bson:"_id"`
-	ChunkSize   int           `bson:"chunkSize"`
-	Name        string        `bson:"filename"`
-	Length      int64         `bson:"length"`
-	Md5         string        `bson:"md5"`
-	UploadDate  time.Time     `bson:"uploadDate"`
-	ContentType string        `bson:"contentType,omitempty"`
+// Struct representing a GridFS files collection document.
+type gfsFile struct {
+	Id         interface{}        `bson:"_id"`
```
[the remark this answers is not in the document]
patrickfreed: Done

mongofiles/options_test.go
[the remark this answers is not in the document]
xdg: The tool is documented as defaulting to majority, so I think we need to do that.

mongofiles/mongofiles.go
```diff
@@ -57,185 +62,282 @@ type MongoFiles struct {
...
-		} else {
-			mf.FileName = args[1]
-		}
-	case Search, Put, Get, Delete:
-		if len(args) > 2 {
-			return fmt.Errorf("too many positional arguments")
+// Write data to GridFS Upload Stream. If this file has not been written to before, this function will open a new stream that must be closed.
+// Note: the go driver buffers data until it hits a chunk size before writing to the database, so if the amount of data is written < chunkSize,
+// the actual write to the database will occur in Close.
+func (file *gfsFile) Write(p []byte) (int, error) {
```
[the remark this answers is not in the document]
xdg: When I read `getTargetGFSFile()`, I don't read that as "open".  Look at how Go's [File](https://golang.org/pkg/os/#File) type uses `Open` and `OpenFile` as constructors.  Here's something similar I would find more natural/obvious:

```
file, err := openTargetGFSFileForWrite()
// err check here
defer file.Close()
```

In particular, that also allows handling any error opening distinctly from the write operation.

mongofiles/mongofiles.go
```diff
@@ -57,185 +62,282 @@ type MongoFiles struct {
...
-// ValidateCommand ensures the arguments supplied are valid.
-func (mf *MongoFiles) ValidateCommand(args []string) error {
-	// make sure a command is specified and that we don't have
-	// too many arguments
-	if len(args) == 0 {
-		return fmt.Errorf("no command specified")
-	}
+// Struct representing the metadata associated with a GridFS files collection document.
+type gfsFileMetadata struct {
+	ChunkSize   int                `bson:"chunkSize"`
```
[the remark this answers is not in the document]
xdg: Including at the top level is great. Thanks.

mongofiles/mongofiles.go
```diff
@@ -57,185 +62,282 @@ type MongoFiles struct {
...
-		}
-	case Search, Put, Get, Delete:
-		if len(args) > 2 {
-			return fmt.Errorf("too many positional arguments")
+// Write data to GridFS Upload Stream. If this file has not been written to before, this function will open a new stream that must be closed.
+// Note: the go driver buffers data until it hits a chunk size before writing to the database, so if the amount of data is written < chunkSize,
+// the actual write to the database will occur in Close.
+func (file *gfsFile) Write(p []byte) (int, error) {
+	if file.upStream == nil {
+		rawBSON, err := bson.Marshal(file.Metadata)
```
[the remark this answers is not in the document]
xdg: OK.

mongofiles/mongofiles.go
```diff
@@ -57,185 +62,282 @@ type MongoFiles struct {
...
-	cursor := gfs.Find(query).Iter()
-	defer cursor.Close()
+	return nil
+}
 
-	var file GFSFile
-	for cursor.Next(&file) {
-		display += fmt.Sprintf("%s\t%d\n", file.Name, file.Length)
+// Closes the file, assigning any error returned to finalErr. Use this to catch errors in deferred closing.
+func (file *gfsFile) SafeClose(finalErr *error) {
```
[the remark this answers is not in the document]
xdg: I'd prefer to keep the original error on the grounds that it was the one that interrupted the program flow and thus is the one that needs to be reported.  After an initial error, the Close error is second-order during cleanup, so less important.

I think it's fine to write helpers.  Here's an example of a DeferredCloser with the properties I mentioned: https://play.golang.org/p/8TSA6yoK0AO

mongofiles/options.go
```diff
@@ -6,13 +6,21 @@
...
+)
+
 var Usage = `<options> <command> <filename or _id>
 
 Manipulate gridfs files using the command line.
 
 Possible commands include:
-	list      - list all files; 'filename' is an optional prefix which listed filenames must begin with
-	search    - search all files; 'filename' is a substring which listed filenames must contain
+	list      - list all files; 'filename' is an optional regex which listed filenames must match
```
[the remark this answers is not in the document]
xdg: ack

mongofiles/mongofiles.go
xdg: In such a case, it might be worth calling that out in a comment.

mongofiles/gfsfile.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	return nil
+}
+
+// Closes any opened Download or Upload streams.
+func (file *gfsFile) Close() error {
+	if file.downStream != nil {
+		if err := file.downStream.Close(); err != nil {
+			return fmt.Errorf("could not close download stream: %v", err)
```
xdg: Note that if there ever were a download and upload stream at the same time, this error would cause the upStream not to get closed.

mongofiles/gfsfile.go
```diff
@@ -0,0 +1,152 @@
...
+	out.mf = mf
+
+	return &out, nil
+}
+
+// Write data to GridFS Upload Stream. If this file has not been written to before, this function will open a new stream that must be closed.
+// Note: the go driver buffers data until it hits a chunk size before writing to the database, so if the amount of data is written < chunkSize,
+// the actual write to the database will occur in Close.
+func (file *gfsFile) Write(p []byte) (int, error) {
+	if file.upStream == nil {
```
xdg: The more I reflect on this, the more I think we shouldn't be combining "open" with "write".  Already the "new..." functions are differentiated for reading and writing contexts.  I get that you're trying to anticipate future usage, like with Delete() checking for an upload stream, but I think it winds up confusing comprehension of how the code is used currently.  I think it would be better to have the gfsFile produce upload/download streams as needed and leave responsibility for closing them to the caller rather than wrapping Read/Write.  I understand that on Swift there's a 1-10 scale for -- on this I'm about a 7 out of 10 for being more explicit and less magical.

mongofiles/options.go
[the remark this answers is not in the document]
xdg: Looking more closely, I now see that "list" constructs a string prefix search -- it does use a regular expression behind the scenes, but it quotes the metacharacters of the input. So 'search' is the one that should be documented as taking a PCRE, and 'list' should stick with saying "prefix"

vendor/github.com/mongodb/mongo-tools-common/db/db.go
```diff
@@ -184,9 +183,6 @@ func configureClient(opts options.ToolOptions) (*mongo.Client, error) {
 	}
 	if opts.WriteConcern != nil {
 		clientopt.SetWriteConcern(opts.WriteConcern)
-	} else {
-		// If no write concern was specified, default to majority
-		clientopt.SetWriteConcern(writeconcern.New(writeconcern.WMajority()))
```
xdg: If all the tools default to majority, then I think we need to keep this.

mongofiles/mongofiles.go
```diff
@@ -57,19 +60,40 @@ type MongoFiles struct {
 	// filename in GridFS
 	FileName string
 
-	//ID to put into GridFS
-	Id string
+	// ID to put into GridFS
+	ID string
```
xdg: process nit: Renaming "Id" to "ID" is fine, but should happened as a minor commit that this PR gets rebased on top of.  It would eliminate of ton of minor diff noise.

mongofiles/mongofiles_test.go
```diff
@@ -8,25 +8,26 @@ package mongofiles
...
+	"github.com/mongodb/mongo-tools-common/testtype"
+	"github.com/mongodb/mongo-tools-common/testutil"
+	"github.com/mongodb/mongo-tools-common/util"
 	. "github.com/smartystreets/goconvey/convey"
-	"gopkg.in/mgo.v2"
 )
 
 var (
-	testDB     = "mongofiles_test_db"
+	testDBName = "mongofiles_test_db"
```
xdg: This is another minor renaming that possibly should be part of a different pull request.  Sometimes renaming as you go is good for clarity -- as this is in abstract -- but seeing this or "data" vs "Data" below adds to the cognitive burden of code review.  The goal should be simplifying CR, so anything unrelated can/should be part of a separate commit.

mongofiles/mongofiles.go
```diff
@@ -130,228 +154,287 @@ func (mf *MongoFiles) ValidateCommand(args []string) error {
...
+	file, err := mf.getTargetGFSFile()
 	if err != nil {
-		return fmt.Errorf("error opening GridFS file '%s': %v", mf.FileName, err)
+		return err
 	}
-	defer gFile.Close()
-	if err = mf.writeFile(gFile); err != nil {
+	defer func() {
+		closeErr := file.Close()
+		if closeErr != nil {
```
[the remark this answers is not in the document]
patrickfreed: This is actually an outdated patch-I've since changed this (and all other `defer` + `Close`s) to use the `DeferredCloser` example you sent in your previous review. Is that still valid, or would you prefer the simpler approach you outlined here?

mongofiles/gfsfile.go
patrickfreed: In a more recent commit than the one this comment is on, I added `OpenForWriting` and `OpenForReading` that must occur outside of the `Write` and `Read` functions respectively. Given that, do you think I should still do away with the `Writer`/`Reader` implementation and just provide the streams as needed, or leave as-is?

xdg: Great!  I think you could trivially have `OpenForWriting/Reading` return the associated stream rather than caching it in the object and then you won't need `Read` and `Write` wrappers.

mongofiles/mongofiles_test.go
patrickfreed: Reverted both this and the "data" one (I believe the latter was a mistake).

I will keep that in mind going forward and reign in some of the more trivial refactors / changes in later ports / CRs.

vendor/github.com/mongodb/mongo-tools-common/db/db.go
patrickfreed: Reverting this change, but before this PR is merged I will have to do more changes here to ensure URI options aren't ignored (once go driver bump happens).

mongofiles/gfsfile.go
patrickfreed: Done. 

This actually leaves `gfsfile.go` a little bare. Is it still worth providing the `Delete`, `OpenForX` functions on the `gfsFile` type, or should I just make them functions of `MongoFiles` instead?

mongofiles/gfsfile.go
patrickfreed: `Close` is no longer a function of `gfsFile`.

patrickfreed:
@xdg I've rebased this PR off of master. If my git operations were done right, the only new changes should be in the "go driver 1.0.0" and later commits.

This will need to wait to be merged until the fixes I made to the go driver are released. They're in the vendor directory right now so that the tests pass on evg.

patrickfreed:
@xdg @kmahar this should be ready for a look. The only commits since David's approval are "go driver 1.0.0" and later.

As I mentioned above, there are some fixes that I vendored in here that are currently on master of the go driver. Should I fix the revision that we depend on, or just keep this PR around until go driver 1.0.1?

PR #125, opened 2019-02-22
TOOLS-2222 Mongodump should ignore config.transactions and config.transaction_coordinators

patrickfreed:
[TOOLS-2222](https://jira.mongodb.com/browse/TOOLS-2222)

### DIFF
--- a/mongodump/prepare.go
+++ b/mongodump/prepare.go
@@ -149,7 +149,7 @@ func shouldSkipSystemNamespace(dbName, collName string) bool {
 			return true
 		}
 	case "config":
-		if collName == "transactions" || collName == "system.sessions" {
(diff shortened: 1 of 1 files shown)


mongodump/prepare.go
```diff
@@ -149,7 +149,7 @@ func shouldSkipSystemNamespace(dbName, collName string) bool {
 			return true
 		}
 	case "config":
-		if collName == "transactions" || collName == "system.sessions" {
+		if collName == "transactions" || collName == "system.sessions" || collName == "transaction_coordinators" {
```
xdg: Past 3 options, I think a switch/case block will be clearer, but this is fine for now.  :-)

PR #127, opened 2019-02-27
Add bsonutil.FindStringValueByKey

kmahar:
@xdg [pointed out](https://github.com/10gen/mongomirror/pull/133/files#r260497575) a couple of places in a mongomirror PR where we had duplicated logic to call `bsonutil.FindValueByKey` and then cast the return value to a string. 
searching through the codebase I found several more places we do this, so I've added a helper to do both at once. 

[evg patch](https://evergreen.mongodb.com/version/5c76ae160305b922178f28ee)

### DIFF
--- a/common/bsonutil/bsonutil.go
+++ b/common/bsonutil/bsonutil.go
@@ -21,6 +21,7 @@ import (
 )
 
 var ErrNoSuchField = errors.New("no such field")
+var ErrNoSuchString = errors.New("field present, but is not a string")
 
 // ConvertJSONDocumentToBSON iterates through the document map and converts JSON
 // values to their corresponding BSON values. It also replaces any extended JSON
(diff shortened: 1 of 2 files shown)


common/bsonutil/bsonutil.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 var ErrNoSuchField = errors.New("no such field")
+var ErrNoSuchString = errors.New("field present, but is not a string")
```
patrickfreed: It might be worth allowing this to be formatted with the actual value that was expected to be a string.

e.g.
`field present, but it is not a string: 1234`.

kmahar:
given @xdg's suggestion that I just add a local helper in mongomirror, not sure if we want to merge this now at all. if we do want to merge it, should copy over the subdocument helper I added in 10gen/mongomirror#135. will wait to see what he says. if we do want to merge it, I will update the error message to include the value

xdg:
Let's close this without merging for now and reconsider when we're done with Go driver migration.

PR #132, opened 2019-03-22
TOOLS-1838 Migrate Mongorestore

patrickfreed:
[TOOLS-1838](https://jira.mongodb.org/browse/TOOLS-1838)


mongorestore/main/mongorestore.go
```diff
@@ -8,47 +8,23 @@
...
-)
 
-const (
-	progressBarLength   = 24
-	progressBarWaitTime = time.Second * 3
+	"os"
 )
 
 func main() {
-	// initialize command-line opts
```
patrickfreed: most of the options parsing was moved to `ParseArgs` in `options.go` and other logic to `New` in `mongorestore.go`.

mongorestore/mongorestore_test.go
```diff
@@ -27,10 +28,19 @@ func init() {
 	})
 }
 
-var (
-	testServer = "localhost"
-	testPort   = db.DefaultTestPort
-)
+func getRestoreWithArgs(additionalArgs ...string) (*MongoRestore, error) {
```
patrickfreed: All integration tests were rewritten to go through the entire options parsing path. It makes them easier to read and also enables testing of the actual user experience.

mongorestore/options.go
```diff
@@ -16,6 +25,27 @@ or use -d and -c to restore a single collection from a single .bson file.
...
+type Options struct {
+	*options.ToolOptions
+	*InputOptions
+	*NSOptions
+	*OutputOptions
+	TargetDirectory string
+}
+
+// InputOptions command line argument long names
+const (
```
patrickfreed: Because we're using these in the tests, I added constants for all the command line arguments. These will need to be kept up to date with the labels in the various options types. If they're tested properly (which not all of them are), tests will break when they become out of date.

mongorestore/options.go
```diff
@@ -76,3 +136,78 @@ type NSOptions struct {
...
+		return Options{}, fmt.Errorf("error parsing write concern: %v", err)
+	}
+	opts.WriteConcern = wcOpts.ToMongoWriteConcern()
+
+	return Options{opts, inputOpts, nsOpts, outputOpts, targetDir}, nil
+}
+
+// getTargetDirFromArgs handles the logic and error cases of figuring out
+// the target restore directory.
+func getTargetDirFromArgs(extraArgs []string, dirFlag string) (string, error) {
```
patrickfreed: moved from `main/mongorestore.go`

test/qa-tests/jstests/restore/users_and_roles_admin.js
```diff
@@ -0,0 +1,193 @@
+// Tests that mongodump and mongorestore properly handle access control information when doing
```
patrickfreed: This was a test in legacy30, but the stuff it was testing wasn't covered anywhere in the new jstests or integration tests. As such, I copied it here and updated parts of it.

vendor/github.com/mongodb/mongo-tools-common/db/write_concern.go
```diff
@@ -40,10 +40,9 @@ type WriteConcernOptions struct {
 	WTimeout   time.Duration
```
patrickfreed: the changes here and in `db/db.go` are just re-vendored in. the other changes in `vendor` are parts of this PR.

mongorestore/main/mongorestore.go
xdg: Thanks.  That will help make [TOOLS-2195](https://jira.mongodb.org/browse/TOOLS-2195) easier, as well.

mongorestore/metadata.go
```diff
@@ -337,6 +337,33 @@ func createCollectionCommand(intent *intents.Intent, options bson.D) bson.D {
 // RestoreUsersOrRoles accepts a users intent and a roles intent, and restores
 // them via _mergeAuthzCollections. Either or both can be nil. In the latter case
 // nothing is done.
+//
+// _mergeAuthzCollections is an internal server command implemented specifically for mongorestore. Instead of inserting
```
xdg: Thank you for documenting this!

mongorestore/options.go
xdg: ack

patrickfreed:
@xdg so this was actually a draft PR that I was still working on a bit. I put it up before I decided to remove the mgo bson library dependency, which I've since added. This leaked into mongodump only slightly, so I think it's okay to include as part of the port.

(this review is no longer a draft, btw)

vendor/github.com/mongodb/mongo-tools-common/db/bson_stream.go
```diff
@@ -77,6 +79,21 @@ func (dbs *DecodedBSONSource) Next(result interface{}) bool {
 	return true
 }
 
+// NextGBSON unmarshals the next BSON document into result using the official go driver. Returns true if no errors are
+// encountered and false otherwise. This function does NOT zero out the result before writing to it.
+func (dbs *DecodedBSONSource) NextGBSON(result interface{}) bool {
```
patrickfreed: This is added so we can easily replace it with a generic `Next` once we drop mgo bson support in mongo-tools-common. The slack conversation re: zeroing in `Unmarshal` is related to this function.

vendor/github.com/mongodb/mongo-tools-common/db/db.go
```diff
@@ -86,14 +87,14 @@ type ApplyOpsResponse struct {
 
 // Oplog represents a MongoDB oplog document.
 type Oplog struct {
-	Timestamp bson.MongoTimestamp `bson:"ts"`
+	Timestamp primitive.Timestamp `bson:"ts"`
```
xdg: I'm a bit worried how this will impact mongodump. Did you test it?

patrickfreed: I did test it and it passes our current tests (integration and js). 

I had to make some slight [changes](https://github.com/mongodb/mongo-tools/blob/ac61328529fa098692aec9eb05123bc01156cf58/mongodump/oplog_dump.go) to mongodump as part of this, though. 

Mongo mirror uses its own oplog type, so it is unaffected by this change.

PR #135, opened 2019-04-01
TOOLS-1836 Migrate mongoexport to the new Go driver

divjotarora:


mongoexport/options.go
```diff
@@ -83,3 +87,61 @@ func (inputOptions *InputOptions) GetQuery() ([]byte, error) {
...
+// ParseOptions reads command line arguments and converts them into options that can be used to configure mongoexport.
+func ParseOptions(rawArgs []string) (Options, error) {
+	// initialize command-line opts
+	opts := options.New("mongoexport", Usage,
+		options.EnabledOptions{Auth: true, Connection: true, Namespace: true, URI: true})
+	outputOpts := &OutputFormatOptions{}
+	opts.AddOptions(outputOpts)
+	inputOpts := &InputOptions{}
+	opts.AddOptions(inputOpts)
+	opts.AddKnownURIParameters(options.KnownURIOptionsReadPreference)
```
divjotarora: This is part of the fix for https://jira.mongodb.org/browse/TOOLS-2217. Adding a known URI param will stop it from logging readPreference as an unsupported param, but it doesn't fix the fact that read pref isn't set on the ToolOptions. I think to get that behavior, common will need to be changed to have a method like NewMongoWriteConcernOpts for read preferences to take both a read pref string and a connstring.

mongoexport/options_test.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (C) MongoDB, Inc. 2014-present.
```
divjotarora: There will be more tests here to test specifying a read preference in the URI once we figure out how to do TOOLS-2217.

mongoexport/mongoexport.go
```diff
@@ -378,6 +367,27 @@ func (exp *MongoExport) exportInternal(out io.Writer) (int64, error) {
...
+	}
+
+	coll := session.Database(exp.ToolOptions.Namespace.DB).Collection(exp.ToolOptions.Namespace.Collection)
+	collInfo, err := db.GetCollectionInfo(coll)
+	if err != nil {
+		return 0, err
+	}
+
+	// If the collection doesn't exist, GetCollectionInfo will return nil
+	if collInfo == nil {
```
divjotarora: Before the port, we weren't checking if the collection exists up front. I added this logic because I didn't want to add nil checks every time we call GetCollectionInfo in the code.

mongoexport/main/mongoexport.go
```diff
@@ -91,36 +73,34 @@ func main() {
 		inputOpts.ReadPreference = "nearest"
 	}
 
-	var mode mgo.Mode
-	if opts.ReplicaSetName != "" || isMongos {
-		mode = mgo.Primary
-	} else {
-		mode = mgo.Nearest
```
divjotarora: Talked to @xdg about this. Setting the default read preference to nearest was actually trying to make a direct connection to the server. This is already taken care of in the new common library (https://github.com/mongodb/mongo-tools-common/blob/master/options/options.go#L472) so this isn't needed.

mongoexport/main/mongoexport.go
```diff
@@ -65,23 +64,6 @@ func main() {
...
 
-	provider, err := db.NewSessionProvider(*opts)
-	if err != nil {
-		log.Logvf(log.Always, "%v", err)
-		os.Exit(util.ExitError)
-	}
-	defer provider.Close()
-
-	// temporarily allow secondary reads for the isMongos check
-	provider.SetReadPreference(mgo.Nearest)
```
divjotarora: This is already taken care of by isMongos (https://github.com/mongodb/mongo-tools-common/blob/master/db/command.go#L125)

test/qa-tests/jstests/export/type_case.js
```diff
@@ -33,7 +33,7 @@
     '--type="foobar"',
     '--fields', 'a']
     .concat(commonToolArgs));
-  assert.eq(3, ret);
+  assert.eq(1, ret);
```
divjotarora: This was changed because MongoExport.validateSettings is now taken care of by MongoExport.New, so failures from that method return with util.ExitError instead of util.ExitBadOptions.

mongoexport/options.go
divjotarora: I added a comment to TOOLS-2217 to track this.

mongoexport/mongoexport.go
divjotarora: I considered it but decided against it to minimize diff noise. I'm open to doing this in another commit on this ticket or on a different ticket entirely if we think it's a good idea.

mongoexport/main/mongoexport.go
```diff
@@ -62,83 +36,12 @@ func main() {
...
-	isMongos, err := provider.IsMongos()
+	exporter, err := mongoexport.New(opts)
 	if err != nil {
 		log.Logvf(log.Always, "%v", err)
 		os.Exit(util.ExitError)
 	}
-
-	provider.SetFlags(db.DisableSocketTimeout)
```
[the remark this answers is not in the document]
divjotarora: Based on the new driver code and talking to Kris, I think connections will have no explicit socket timeout if we don't explicitly set one when create the client, so this line isn't needed.

mongoexport/options.go
```diff
@@ -83,3 +87,61 @@ func (inputOptions *InputOptions) GetQuery() ([]byte, error) {
...
+
+	// verify URI options and log them
+	opts.URI.LogUnsupportedOptions()
+
+	if inputOpts.SlaveOk {
+		if inputOpts.ReadPreference != "" {
+			return Options{}, fmt.Errorf("--slaveOk can't be specified when --readPreference is specified")
+		}
+
+		log.Logvf(log.Always, "--slaveOk is deprecated and being internally rewritten as --readPreference=nearest")
```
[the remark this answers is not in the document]
divjotarora: Can you clarify? Are you proposing changing this to an error or just changing the error message?

mongoexport/mongoexport.go
```diff
@@ -224,82 +279,61 @@ func (exp *MongoExport) getCount() (c int, err error) {
...
 	// or if we cannot because  we are querying, sorting, or if the collection is a view
 	if !exp.InputOpts.ForceTableScan && len(query) == 0 && exp.InputOpts != nil && exp.InputOpts.Sort == "" && !collInfo.IsView() && !collInfo.IsSystemCollection() {
-		flags = flags | db.Snapshot
+		findOpts.SetSnapshot(true)
 	}
 
-	skip := 0
 	if exp.InputOpts != nil {
-		skip = exp.InputOpts.Skip
+		findOpts.SetSkip(int64(exp.InputOpts.Skip))
```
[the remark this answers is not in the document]
divjotarora: Thought about doing this but getCount() returns an int and in 2 cases, returns the limit and skip so we'd need to cast in those cases instead. I considered making a larger change to have getCount() return int64 as well, but decided against it to have a smaller diff. Like my comment above, I'm open to doing this in a separate commit if you think it'll help.

mongoexport/mongoexport.go
```diff
@@ -345,18 +378,21 @@ func (exp *MongoExport) exportInternal(out io.Writer) (int64, error) {
...
+			return docsCount, err
+		}
+
 		err := exportOutput.ExportDocument(result)
 		if err != nil {
 			return docsCount, err
 		}
 		docsCount++
-		if docsCount % watchProgressorUpdateFrequency == 0 {
+		if docsCount%watchProgressorUpdateFrequency == 0 {
```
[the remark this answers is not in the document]
divjotarora: it's just something that gofmt does automatically. I don't mind leaving it how it is, but it'll change every time we run gofmt, so I'm inclined to remove the spacing.

mongoexport/mongoexport.go
divjotarora: Added a cached version of CollectionInfo in the latest commit.

mongoexport/mongoexport.go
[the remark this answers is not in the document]
divjotarora: Made the larger change in a new commit.

mongoexport/mongoexport.go
```diff
@@ -69,9 +81,55 @@ type ExportOutput interface {
 	Flush() error
 }
 
-// ValidateSettings returns an error if any settings specified on the command line
+// New constructs a new MongoExport instance from the provided options.
+func New(opts Options) (*MongoExport, error) {
+	exporter := &MongoExport{
+		ToolOptions: *opts.ToolOptions,
```
patrickfreed: I know this is existing code, but I think it's worth updating now while you're tweaking the options: most of the other tools store ToolOptions as a pointer, probably to avoid copying the entire struct. For consistency and/or performance we should probably store a ToolOptions pointer as well in mongoexport.

mongoexport/mongoexport.go
```diff
@@ -378,6 +408,27 @@ func (exp *MongoExport) exportInternal(out io.Writer) (int64, error) {
 // of documents successfully exported, and a non-nil error if something went wrong
 // during the export operation.
 func (exp *MongoExport) Export(out io.Writer) (int64, error) {
+	session, err := exp.SessionProvider.GetSession()
```
patrickfreed: the docstring for `exportInternal` claims it is used directly in testing which would bypass this new logic. We then may want to move this code into `exportInternal` so existing and new tests written that way will exercise it.

mongoexport/options_test.go
```diff
@@ -0,0 +1,57 @@
...
+		So(opts.InputOptions.ReadPreference, ShouldEqual, inputRp)
+		if toolsRp == nil {
+			So(opts.ToolOptions.ReadPreference, ShouldBeNil)
+		} else {
+			So(opts.ToolOptions.ReadPreference.Mode(), ShouldEqual, toolsRp.Mode())
+		}
+	}
+}
+
+func TestOptionsParsing(t *testing.T) {
```
patrickfreed: This test (and the helper) are only really testing read preference behavior, so it should probably be named less generically to indicate that (e.g. `TestReadPreferenceOptionsParsing`).

mongoexport/options_test.go
patrickfreed: In the meantime, you can add an integration test for read preference behavior if you want. Given the way common is structured, it should still be getting set on the client even if it's absent from `ToolOptions`.

mongoexport/options.go
patrickfreed: I'd say go ahead and implement it to unblock this PR. (@xdg is it okay to bump the priority on TOOLS-2217?)

xdg: Sure.

mongoexport/mongoexport.go
divjotarora: Done in new commits.

mongoexport/mongoexport.go
divjotarora: Moved to exportInternal.

mongoexport/options.go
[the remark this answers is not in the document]
kmahar: sorry, I meant "log message", not "error message". I think we should just change the message.

mongoexport/options.go
divjotarora: I'll put TOOLS-2217 in progress and block 1836 on it then.

mongoexport/options_test.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func TestReadPreferenceParsing(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	Convey("With ToolOptions and InputOptions", t, func() {
+		Convey("Parsing with no values should leave read pref empty",
+			validateReadPreferenceParsing([]string{}, true, "", nil))
+
+		Convey("Parsing with value only in command line opts",
```
kmahar: nit: update string to add "should [do X]", like the other descriptions have.

mongoexport/options_test.go
divjotarora: I'm starting 2217 now so I'll add the tests as soon as that's done.

mongoexport/options_test.go
divjotarora: Changed the names to include read preference in latest commits.

test/qa-tests/jstests/export/type_case.js
kmahar: it seems like we might still want the original error code if it's possible, since it better conveys the problem.

mongoexport/main/mongoexport.go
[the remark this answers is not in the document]
divjotarora: There's likely no value to keeping this in. There will need to be some follow-up work to remove unused functions in common after all of the ports are done, but I think that should be done as a separate commit/ticket rather than part of this port.

mongoexport/mongoexport.go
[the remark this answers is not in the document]
divjotarora: Reverted this in new commits. I agree we should do one big PR for all of the violations later.

mongoexport/options.go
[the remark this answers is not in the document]
divjotarora: Changed log msg in new commits.

mongoexport/options_test.go
divjotarora: Done.

test/qa-tests/jstests/export/type_case.js
divjotarora: I agree the original error code was better. The problem was that validateSettings should be called by New, but if New returns an err, it's not possible to know where that error came from. I've added a SetupError type in the new commits to convey what the error code should be when exiting.

mongoexport/mongoexport.go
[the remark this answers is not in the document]
kmahar: can you open a ticket so we remember to do it, please?

mongoexport/main/mongoexport.go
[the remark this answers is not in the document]
kmahar: sounds good to me. is there a ticket for that?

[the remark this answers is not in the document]
divjotarora: https://jira.mongodb.org/browse/TOOLS-2248

mongoexport/mongoexport.go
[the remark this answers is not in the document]
divjotarora: https://jira.mongodb.org/browse/TOOLS-2247

kmahar:
no further comments on the existing code - all lgtm 👍 just re-request a review once you've updated based on 2217 and I will look at the new stuff

mongoexport/options_test.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	}
+}
+
+func TestReadPreferenceParsing(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	Convey("With ToolOptions and InputOptions", t, func() {
+		Convey("Parsing with no values should leave read pref empty",
+			validateReadPreferenceParsing([]string{}, true, "", nil))
```
xdg: There aren't that many cases here, but even still I'd normally prefer to see table-driven testing.  For *new* tests -- particularly unit tests that don't need test fixture setup, I'd prefer normal go testing without GoConvey.

mongoexport/mongoexport.go
```diff
@@ -32,11 +36,16 @@ const (
...
+	progressBarLength   = 24
+	progressBarWaitTime = time.Second
+)
+
 // MongoExport is a container for the user-specified options and
 // internal state used for running mongoexport.
 type MongoExport struct {
 	// generic mongo tool options
-	ToolOptions options.ToolOptions
+	ToolOptions *options.ToolOptions
```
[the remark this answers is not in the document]
patrickfreed: This was something I suggested, rationale here: https://github.com/mongodb/mongo-tools/pull/135#discussion_r271459910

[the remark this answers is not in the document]
xdg: Ack.  I agree for consistency. Given that this object is created once and passed around a handful of times, I'm not sold on the 'performance optimization argument' until someone actually benchmarks it.

mongoexport/options_test.go
divjotarora: I also prefer the normal table driven tests over GoConvey. I rewrote them as table driven. One possible thing to consider in the future is using some sort of assertion library (e.g. testify) to simplify comparison logic instead of having conditionals and t.Fatalf everywhere.

PR #136, opened 2019-04-02
TOOLS-2173 Migrate mongo-tools-common to Go driver bson compatibility

patrickfreed:
[TOOLS-2173](https://jira.mongodb.org/browse/TOOLS-2173)

This PR updates mongo-tools-common to use the new go driver's bson library (see vendor/). The corresponding changes to the tools are also included here as well (this includes updating mongodump to go driver bson compatibility). The meat of the changs are in vendor, so make sure to check those out. Once this is approved, I'll just make a commit to common without a PR.

NOT INCLUDED here are changes to import/export, which will be incorporated after the respective ports are completed. I've put this up for review early as I expect the changes in those tools to be pretty minimal. 

As part of this change, I've refactored all the `bsonutil` functions dealing with JSON to explicitly mention that they're using the legacy extended json parser so as to differentiate them from the driver's. Going forward, I think we should directly use the driver's marshaling capabilities for extJSON purposes, only really relying on `bsonutil` for legacy parsing.


vendor/github.com/mongodb/mongo-tools-common/testutil/auth.go
```diff
@@ -23,54 +20,6 @@ var (
 	CreatedUserPasswordEnv = "TOOLS_TESTING_AUTH_PASSWORD"
 )
 
-// Initialize a user admin, using the already-connected session passed in.
```
patrickfreed: these functions were not used by any of the tools, so I removed them

vendor/github.com/mongodb/mongo-tools-common/db/namespaces.go
```diff
@@ -45,48 +43,13 @@ func (ci *CollectionInfo) GetUUID() string {
...
-			}
 		default:
 			log.Logvf(log.DebugHigh, "unknown UUID BSON type '%T'", v)
 		}
 	}
 	return ""
 }
 
-// IsNoCmd reeturns true if err indicates a query command is not supported,
-// otherwise, returns false.
```
patrickfreed: these functions were not used by any of the tools, so I removed them

vendor/github.com/mongodb/mongo-tools-common/bsonutil/bsonutil.go
```diff
@@ -324,91 +319,35 @@ func ParseSpecialKeys(special interface{}) (interface{}, error) {
...
-				binary.Kind = kind[0]
+				binary.Subtype = kind[0]
 
 			default:
 				return nil, errors.New("expected $type field to have string value")
 			}
 			return binary, nil
 		}
-
-		if jsonValue, ok := doc["$ref"]; ok {
```
patrickfreed: the go driver has no type associated with a dbRef (and neither does bson, for that matter), so I just removed all the parsing for it. It should go through the parser as a document, which is basically what it is anyways.

vendor/github.com/mongodb/mongo-tools-common/util/timestamp.go
```diff
@@ -0,0 +1,26 @@
...
+			return -1
+		} else {
+			return 0
+		}
+	} else if lhs.T > rhs.T {
+		return 1
+	} else {
+		return -1
+	}
+}
```
kmahar: nit: newline at end of file

vendor/github.com/mongodb/mongo-tools-common/db/namespaces.go
```diff
@@ -120,16 +77,6 @@ func GetCollections(database *mongo.Database, name string) (*mongo.Cursor, error
 	return cursor, nil
 }
 
-func getCollectionsPre28(database *mgo.Database, name string) (*mgo.Iter, error) {
```
kmahar: this was deleted because GODRIVER-492 has been implemented, right? should we also delete the comment above on `GetCollections`?

vendor/github.com/mongodb/mongo-tools-common/db/namespaces.go
```diff
@@ -97,17 +60,11 @@ func GetIndexes(coll *mongo.Collection) (*mongo.Cursor, error) {
 	return coll.Indexes().List(context.Background())
 }
 
-func getIndexesPre28(coll *mgo.Collection) (*mgo.Iter, error) {
```
kmahar: similar Q to below: this was deleted because GODRIVER-279 has been implemented, right? should we also delete the comment above on GetIndexes?

vendor/github.com/mongodb/mongo-tools-common/db/namespaces.go
```diff
@@ -45,48 +43,13 @@ func (ci *CollectionInfo) GetUUID() string {
 			if x.Subtype == 4 {
 				return hex.EncodeToString(x.Data)
 			}
-		case bson.Binary:
```
kmahar: why was this removed?

patrickfreed: `bson.Binary` is an mgo type, `primitive.Binary` is the go driver one. this function apparently handled both (you can see if you expand the diff), so I just removed the mgo one.

vendor/github.com/mongodb/mongo-tools-common/db/namespaces.go
patrickfreed: I removed them simply because they're not called anywhere. I'm not really sure what specifically they were used for in the past. I do think I can remove the comment from `GetIndexes` though, so I've done that.

vendor/github.com/mongodb/mongo-tools-common/db/namespaces.go
patrickfreed: ditto comment above

vendor/github.com/mongodb/mongo-tools-common/util/timestamp.go
patrickfreed: fixed

mongodump/oplog_dump.go
```diff
@@ -86,7 +85,7 @@ func (dump *MongoDump) checkOplogTimestampExists(ts bson.MongoTimestamp) (bool,
 }
 
 func oplogDocumentFilter(in []byte) ([]byte, error) {
-	var rawD bson.RawD
+	var rawD bson.D
```
divjotarora: Nit: the variable name doesn't match the type anymore because we're using a bson.D instead of bson.Raw.

mongodump/oplog_dump.go
```diff
@@ -104,13 +103,14 @@ func oplogDocumentFilter(in []byte) ([]byte, error) {
...
 	}
 
-	for i := range rawD {
-		if rawD[i].Name == "o" {
-			var rawO bson.RawD
-			err = bson.Unmarshal(rawD[i].Value.Data, &rawO)
-			for j := range rawO {
-				if rawO[j].Name == "renameCollection" {
-					return nil, fmt.Errorf("cannot dump with oplog while renames occur")
+	for _, e := range rawD {
```
divjotarora: I know this is copy/paste code, but this might be cleaner if we decode into a bson.M instead of a bson.D so we can do lookups instead of nested for loops.

patrickfreed: Yeah, this for-looping code is pretty bad. I never took a step back to look at it. I've taken your suggestion one step further and just updated the struct that we're already unmarshalling to with an "o.renameCollection" field.

mongodump/oplog_dump.go
patrickfreed: Removed as part of changes re: your other comment.

mongodump/metadata_dump.go
```diff
@@ -46,7 +46,7 @@ func (dump *MongoDump) dumpMetadata(intent *intents.Intent, buffer resettableOut
 	// The collection options were already gathered while building the list of intents.
 	// We convert them to JSON so that they can be written to the metadata json file as text.
 	if intent.Options != nil {
-		if meta.Options, err = bsonutil.ConvertBSONValueToJSON(intent.Options); err != nil {
+		if meta.Options, err = bsonutil.ConvertBSONValueToLegacyExtJSON(intent.Options); err != nil {
```
xdg: I think we ought to convert to canonical Extended JSON -- I don't see a lot of value in using the legacy format for dumping metadata.  Anyone disagree?  A counterargument I would accept is that this ticket is all about getting rid of mgo/bson and we think we should make a change to a new format a separate ticket.   I guess I've talked myself into that point of view.  I'll open a ticket for it.

mongodump/oplog_dump.go
```diff
@@ -44,40 +43,40 @@ func (dump *MongoDump) determineOplogCollectionName() error {
...
 	}
-	err = gbson.Unmarshal(tempBSON, &oldestOplogEntry)
+	err = bson.Unmarshal(tempBSON, &oldestOplogEntry)
 	if err != nil {
 		return false, err
 	}
 
 	log.Logvf(log.DebugHigh, "oldest oplog entry has timestamp %v", oldestOplogEntry.Timestamp)
-	if bsonutil.ConvertTimestampToMongoTimestamp(oldestOplogEntry.Timestamp) > ts {
+	if util.CompareTimestamps(oldestOplogEntry.Timestamp, ts) > 0 {
```
xdg: Maybe make a `util.TimestampGreaterThan(left, right)` function?  It's a little clearer than knowing how "compare" relates to zero.

mongodump/oplog_dump.go
```diff
@@ -86,16 +85,13 @@ func (dump *MongoDump) checkOplogTimestampExists(ts bson.MongoTimestamp) (bool,
...
-	}
-
 	var nsD struct {
 		NS string `bson:"ns"`
+		O *struct {
+			RenameCollection interface{} `bson:"renameCollection"`
+		} `bson:"o"`
 	}
-	err = bson.Unmarshal(in, &nsD)
+	err := bson.Unmarshal(in, &nsD)
```
xdg: I find this approach of defining a struct that doesn't apply to all oplog entries and unmarshalling to see what worked to be a bit odd/confusing.  I think a better approach in this functions is to convert the `in` parameter to a `bson.Raw` and use `Lookup`/`LookupErr` to spot-check the conditions we care about without unmarshaling to struct.  (It's possibly higher performance, too.)

mongodump/oplog_dump.go
```diff
@@ -104,23 +100,16 @@ func oplogDocumentFilter(in []byte) ([]byte, error) {
...
-					return nil, fmt.Errorf("cannot dump with oplog while renames occur")
-				}
-			}
-		}
+	if nsD.O != nil && nsD.O.RenameCollection != nil {
+		return nil, fmt.Errorf("cannot dump with oplog while renames occur")
 	}
-	return bson.Marshal(rawD)
+
+	return append([]byte(nil), in...), nil
```
xdg: Rather than relying on append for the copy, I'd prefer an explicit `make` followed by `copy`.

I'm kind of annoyed that we're copying here, because it's not clear why it needs to be a copy.  I've filed [TOOLS-2256](https://jira.mongodb.org/browse/TOOLS-2256) about it.

mongorestore/oplog.go
```diff
@@ -136,7 +136,7 @@ func (restore *MongoRestore) TimestampBeforeLimit(ts primitive.Timestamp) bool {
 		// always valid if there is no --oplogLimit set
 		return true
 	}
-	return ts.T < restore.oplogLimit.T || (ts.T == restore.oplogLimit.T && ts.I < restore.oplogLimit.I)
+	return util.CompareTimestamps(ts, restore.oplogLimit) < 0
```
xdg: Same point as earlier about a more explicit comparison function.

divjotarora:
Question unrelated to the code: is there a ticket to track the changes necessary to import/export? I remember there was some stuff in import that could be changed to use the new BSON lib once common had been converted.

patrickfreed:
There are no tickets for it yet. I'll make what changes make sense for it in this PR and ticket out any leftovers as necessary.

mongodump/metadata_dump.go
patrickfreed: hah, I had the exact same sequence of thoughts on this.

mongodump/oplog_dump.go
patrickfreed: This has since been changed to using `Raw`'s `LookupErr` instead of partial unmarshalling.

vendor/github.com/mongodb/mongo-tools-common/util/timestamp.go
```diff
@@ -9,18 +9,6 @@ package util
...
-		} else {
-			return 0
-		}
-	} else if lhs.T > rhs.T {
-		return 1
-	} else {
-		return -1
-	}
+func TimestampGreaterThan(lhs, rhs primitive.Timestamp) bool {
+	return lhs.T > rhs.T || lhs.T == rhs.T && lhs.I > rhs.I
```
xdg: It's now out of sync with the comment.  :-)

patrickfreed: Oh, crap. Fixed.

kmahar:
everything up to this point LGTM. please just re-request a review once you've added import/export changes and I will come check those out

patrickfreed:
This has been rebased off of the import/export port and updated to remove the last mgo references. As part of that, I discovered a couple of errors in my original common update with handling code, code w/ scope, and datetimes, which have since been fixed.

mongoimport/mongoimport_test.go
```diff
@@ -21,9 +21,9 @@ import (
 	"github.com/mongodb/mongo-tools-common/testtype"
 	"github.com/mongodb/mongo-tools-common/testutil"
 	. "github.com/smartystreets/goconvey/convey"
-	gbson "go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson"
 	driverOpts "go.mongodb.org/mongo-driver/mongo/options"
-	"gopkg.in/mgo.v2/bson"
+
```
divjotarora: nit: this newline shouldn't be here

PR #137, opened 2019-04-03
TOOLS-2217 Use command line and URI arguments to construct read preference

divjotarora:
This blocks TOOLS-1836 (porting mongoexport). I decided to do this as a separate PR because the changes to mongofiles and mongodump seemed out of scope for the mongoexport PR. Once this is approved, I'll make the changes to common and revendor without doing another code review.

### DIFF
--- a/mongodump/mongodump.go
+++ b/mongodump/mongodump.go
@@ -132,15 +132,9 @@ func (dump *MongoDump) Init() error {
 		dump.OutputWriter = os.Stdout
 	}
 
-	var pref *readpref.ReadPref
-	if dump.InputOptions.ReadPreference != "" {
-		pref, err = db.ParseReadPreference(dump.InputOptions.ReadPreference)
-		if err != nil {
-			return fmt.Errorf("error parsing --readPreference : %v", err)
-		}
-	} else {
-		pref = readpref.Primary()
-		// Direct connection is configured elsewhere if replica set name is not given
+	pref, err := db.ParseReadPreference(dump.InputOptions.ReadPreference, dump.ToolOptions.URI.ParsedConnString())
+	if err != nil {
+		return fmt.Errorf("error parsing --readPreference : %v", err)
 	}
 	dump.ToolOptions.ReadPreference = pref
 
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -67,6 +67,7 @@ func simpleMongoDumpInstance() *MongoDump {
 			Connection: connection,
 			Auth:       &auth,
 			Verbosity:  &options.Verbosity{},
+			URI:        &options.URI{},
 		}
 	}
 
--- a/mongofiles/options.go
+++ b/mongofiles/options.go
@@ -11,7 +11,6 @@ import (
 	"github.com/mongodb/mongo-tools-common/db"
 	"github.com/mongodb/mongo-tools-common/log"
 	"github.com/mongodb/mongo-tools-common/options"
-	"go.mongodb.org/mongo-driver/mongo/readpref"
 )
 
 // Usage string printed as part of --help
@@ -65,16 +64,9 @@ func ParseOptions(rawArgs []string) (Options, error) {
 	opts.WriteConcern = wc
 
 	// set ReadPreference
-	if inputOpts.ReadPreference != "" {
-		opts.ReadPreference, err = db.ParseReadPreference(inputOpts.ReadPreference)
-		if err != nil {
-			return Options{}, fmt.Errorf("error parsing --readPreference: %v", err)
-		}
-	} else {
-		opts.ReadPreference, err = readpref.New(readpref.NearestMode)
-		if err != nil {
-			return Options{}, fmt.Errorf("error setting default read preference: %v", err)
-		}
+	opts.ReadPreference, err = db.ParseReadPreference(inputOpts.ReadPreference, opts.URI.ParsedConnString())
+	if err != nil {
+		return Options{}, fmt.Errorf("error parsing --readPreference: %v", err)
 	}
 
 	return Options{opts, storageOpts, inputOpts, args}, nil
--- a/vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
+++ b/vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
@@ -12,6 +12,7 @@ import (
 	"github.com/mongodb/mongo-tools-common/json"
 	"go.mongodb.org/mongo-driver/mongo/readpref"
 	"go.mongodb.org/mongo-driver/tag"
+	"go.mongodb.org/mongo-driver/x/network/connstring"
 )
 
 type readPrefDoc struct {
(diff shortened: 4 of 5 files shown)


mongodump/mongodump_test.go
```diff
@@ -67,6 +67,7 @@ func simpleMongoDumpInstance() *MongoDump {
 			Connection: connection,
 			Auth:       &auth,
 			Verbosity:  &options.Verbosity{},
+			URI:        &options.URI{},
```
divjotarora: I added this because the mongodump tests panicked because opts.URI.ParsedConnString() doesn't check if the URI is nil.

mongofiles/options.go
```diff
@@ -65,16 +64,9 @@ func ParseOptions(rawArgs []string) (Options, error) {
 	opts.WriteConcern = wc
 
 	// set ReadPreference
-	if inputOpts.ReadPreference != "" {
-		opts.ReadPreference, err = db.ParseReadPreference(inputOpts.ReadPreference)
-		if err != nil {
-			return Options{}, fmt.Errorf("error parsing --readPreference: %v", err)
-		}
-	} else {
-		opts.ReadPreference, err = readpref.New(readpref.NearestMode)
```
divjotarora: I think the nearest read preference was being used to mock a direct connection. Afaik, the default read preference should be primary. The logic for creating a direct connection is handled in the new common code.

vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
```diff
@@ -24,15 +25,22 @@ const (
 		"connection to mongos may produce inconsistent duplicates or miss some documents."
 )
 
-func Primary() *readpref.ReadPref          { return readpref.Primary() }
```
divjotarora: Removed these because callers can use `readpref` directly.

vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
```diff
@@ -24,15 +25,22 @@ const (
...
-func PrimaryPreferred() *readpref.ReadPref { return readpref.PrimaryPreferred() }
-func Nearest() *readpref.ReadPref          { return readpref.Nearest() }
+// ParseReadPreference takes a string (command line read preference argument) and a ConnString (from the command line
+// URI argument) and returns a ReadPref. If both are provided, preference is given to the command line argument. If
+// both are empty, a default read preference of nearest will be returned.
```
patrickfreed: comment should be updated to reflect the new default of Primary.

mongofiles/options.go
```diff
@@ -65,16 +64,9 @@ func ParseOptions(rawArgs []string) (Options, error) {
...
-		if err != nil {
-			return Options{}, fmt.Errorf("error setting default read preference: %v", err)
-		}
+	opts.ReadPreference, err = db.ParseReadPreference(inputOpts.ReadPreference, opts.URI.ParsedConnString())
+	if err != nil {
+		return Options{}, fmt.Errorf("error parsing --readPreference: %v", err)
```
patrickfreed: Because an error could occur as a result of parsing the URI now, I think this error (and similar ones) should be updated to just say "error parsing read preference...", omitting the "--".

mongodump/mongodump.go
```diff
@@ -132,15 +132,9 @@ func (dump *MongoDump) Init() error {
...
-	var pref *readpref.ReadPref
-	if dump.InputOptions.ReadPreference != "" {
-		pref, err = db.ParseReadPreference(dump.InputOptions.ReadPreference)
-		if err != nil {
-			return fmt.Errorf("error parsing --readPreference : %v", err)
-		}
-	} else {
-		pref = readpref.Primary()
-		// Direct connection is configured elsewhere if replica set name is not given
+	pref, err := db.ParseReadPreference(dump.InputOptions.ReadPreference, dump.ToolOptions.URI.ParsedConnString())
```
patrickfreed: to match the write concern counterpart, we might want to refactor `ParseReadPreference` to `NewReadPreference` (similar to `NewWriteConcern`).

mongodump/mongodump_test.go
patrickfreed: Yeah, for things like this is why I rewrote the mongorestore tests to go through the whole command line arg parsing rather than manually setting the options like this. Maybe sometime in the future we'll get around to updating the rest of our integration tests.

vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
```diff
@@ -24,38 +25,62 @@ const (
...
-	var tagSet tag.Set
-	if rp == "" {
-		return readpref.Nearest(), nil
+	if len(cs.ReadPreferenceTagSets) == 0 {
+		// no tag sets
+		return doc
 	}
+
+	// take the last tag set
+	doc.Tags = cs.ReadPreferenceTagSets[len(cs.ReadPreferenceTagSets) - 1]
```
patrickfreed: We shouldn't be discarding extra tag sets. It's totally fine for the user to specify more than one. Instead, update the `readPrefDoc` to have a slice of tag sets and use `NewTagSetsFromMaps` instead.

vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
```diff
@@ -24,38 +25,62 @@ const (
...
+		return nil, err
+	}
+
+	mode, err := readpref.ModeFromString(doc.Mode)
+	if err != nil {
+		return nil, err
+	}
+
+	tagSet := tag.NewTagSetFromMap(doc.Tags)
+	return readpref.New(mode, readpref.WithTagSets(tagSet))
```
patrickfreed: We should also be parsing `maxStalenessMS`. See the [server selection spec](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#components-of-a-read-preference) for more info.

vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
divjotarora: Based on a conversation with Seth, we've decided to keep command line read preference parsing the way it is right now to avoid breaking changes and add support for multiple tag sets later. The new commit separates out the connection string read pref parsing to account for both this and maxStaleness.

vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
```diff
@@ -28,59 +28,53 @@ const (
...
-	return doc
-}
-
-func readPrefDocFromString(rp string) (readPrefDoc, error) {
-	var doc readPrefDoc
-
-	if rp[0] != '{' {
-		doc.Mode = rp
-		return doc, nil
+	if cs.MaxStaleness != 0 {
```
patrickfreed: we can use `cs.MaxStalenessSet` here

vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
patrickfreed: Sounds good. I think we might want to add a TODO comment describing what's going on and mentioning the ticket associated with command line rationalization somewhere in the docstring of `NewReadPreference` (so we don't forget).

vendor/github.com/mongodb/mongo-tools-common/db/read_preferences.go
divjotarora: done

PR #139, opened 2019-04-09
TOOLS-2257 Output canonical extended JSON in bsondump

divjotarora:


kmahar:
code change looks fine, but do we care about allowing people to get relaxed extJSON out at all? should there be an option?

divjotarora:
Talked to @xdg about that and he mentioned that the main use case of bsondump is taking data from a tool like mongodump and getting it into a human-readable format. For this case, canonical is probably best so users can see type information. It's possible that users might want relaxed JSON for documents that only have strings/numbers so that they can feed results into some external tool? @xdg - thoughts?

patrickfreed:
I believe the go change is all that's required in the tools themselves, but `all_types_json.js` is still expecting the relaxed format and will need to be updated.

divjotarora:
I've updated the all_types_json.js test to expect canonical ext JSON and those changes pass on Evergreen. The output_file.js test is failing because it uses bsondump to generate JSON files and then attempts to put them in MongoDB via mongoimport, which doesn't currently parse the new ext JSON format. For the time being, I'm going to block TOOLS-2257 on TOOLS-1610.

divjotarora:
TOOLS-1610 has been pushed now, so I've requested another review on this because it's been a while since this went up.

PR #140, opened 2019-04-10
TOOLS-2247

divjotarora:
There are lots of changes here from running gofmt that can be ignored. The only file that needs to be reviewed is common.yml, which has the changes for adding a format task to Evergreen.


common.yml
```diff
@@ -740,6 +742,17 @@ functions:
         set -e
         GOPATH=$PWD go get github.com/10gen/ftdc-utils/cmd/ftdc
 
+  "fetch gometalinter" :
```
divjotarora: I chose to use gometalinter, which is a wrapper that can run many different linters/formatters, instead of running gofmt directly. This is because gofmt doesn't return a non-zero exit code if the files aren't correctly formatted, so figuring out if formatting failed required some extra work that didn't seem worth it.

Gometalinter is deprecated in favor of golangci-lint, but I talked to Sam Kleinman on Evergreen and he said gometalinter is fine for our purposes.

xdg: gometalinter repo says: "DEPRECATED: Use https://github.com/golangci/golangci-lint"  Maybe we should use the recommended tool instead.

common.yml
```diff
@@ -1010,6 +1023,23 @@ tasks:
...
+      type: test
+      params:
+        working_dir: src/github.com/mongodb/mongo-tools
+        script: |
+          set -x
+          set -v
+          set -e
+          . ./set_goenv.sh
+          GOROOT="" set_goenv || exit
+          PATH=$PATH:$PWD/bin gometalinter --disable-all --enable=gofmt --vendor --exclude=src ./...
```
divjotarora: This only runs gofmt, but we can extend this to run other linters later on if we think that would be helpful. The --exclude=src is because the "fetch gometalinter" task puts the "src" folder in the working directory and the gometalinter src has some gofmt errors.

xdg: That's ironic.  We could vendor it, though, like we do with 3rf/mongo-lint.

common.yml
xdg: I think "fine for our purposes" isn't the same as "a good choice for the future".  Starting with deprecated tools seems the wrong approach.

divjotarora: Fair point - I'll update it to golangci-lint

common.yml
divjotarora: No longer needed - we can install the golangci-lint binary into bin/ directly without running go get.

PR #141, opened 2019-04-11
TOOLS-1610 Add extended JSON v2 support to mongoimport/mongoexport

divjotarora:


mongoexport/json.go
```diff
@@ -89,27 +87,26 @@ func (jsonExporter *JSONExportOutput) ExportDocument(document bson.D) error {
...
 	} else {
-		extendedDoc, err := bsonutil.ConvertBSONValueToLegacyExtJSON(document)
+		extendedDoc, err := bson.MarshalExtJSON(document, jsonExporter.JSONFormat == canonical, false)
 		if err != nil {
 			return err
 		}
-		err = jsonExporter.Encoder.Encode(extendedDoc)
-		if err != nil {
+
+		extendedDoc = append(extendedDoc, '\n')
```
divjotarora: Appending a new line was originally handled by the encoder field, which is no longer necessary because bson.MarshalExtJSON returns the exact bytes we need to write out.

test/qa-tests/jstests/import/types.js
```diff
@@ -74,7 +74,7 @@
   var oid = ObjectId();
   var irregularObjects = {
     _id: ObjectId(),
-    a: DBPointer('namespace', oid),
+    a: DBRef('namespace', oid),
```
divjotarora: not sure if this change is correct - DBPointer was being marshalled differently by the driver and our results expect a DBRef.

divjotarora:
The exportimport_date.js test still fails. The second date (yr 3001) is roundtripped as year 1831 by the driver. I think this is because the driver's ext JSON library uses time.Parse followed by time.UnixNano(), which only works for years between 1678 and 2262, so some sort of wrap around happens. Any thoughts on how to address this? I'm not sure what the test is trying to do because the comment above the problematic date says "Insert a date that we cannot format as an ISODate string".

mongoexport/json.go
```diff
@@ -24,20 +22,20 @@ type JSONExportOutput struct {
 	ArrayOutput bool
 	// Pretty when set to true indicates that the output will be written in pretty mode.
 	PrettyOutput bool
-	Encoder      *json.Encoder
 	Out          io.Writer
 	NumExported  int64
+	JSONFormat   string
```
patrickfreed: It might be worth defining an enum representing the different possible cases and storing that rather than just storing a raw string here.

mongoexport/mongoexport.go
```diff
@@ -33,6 +33,8 @@ const (
 	CSV                            = "csv"
 	JSON                           = "json"
 	watchProgressorUpdateFrequency = 8000
+	canonical                      = "canonical"
```
patrickfreed: [nit] the comment above says these constants are the different output types, which isn't exactly true. I'd say either update the comment or break out the constants defined here into logical categories (e.g. output formats, extended json type, etc)

mongoexport/options.go
```diff
@@ -46,6 +46,9 @@ type OutputFormatOptions struct {
 
 	// NoHeaderLine, if set, will export CSV data without a list of field names at the first line.
 	NoHeaderLine bool `long:"noHeaderLine" description:"export CSV data without a list of field names at the first line"`
+
+	// JSONFormat specifies what extended JSON format to export (canonical or relaxed). Defaults to relaxed.
+	JSONFormat string `long:"jsonFormat" default:"relaxed" description:"the extended JSON format to output, either canonical or relaxed (defaults to 'relaxed')"`
```
patrickfreed: [nit] if you add a "value-name" similar to the other options, it will make the --help output a little nicer.

test/qa-tests/jstests/import/types.js
patrickfreed: So DBRef isn't actually a bson type; it's more of a convention. DBPointer _is_ defined by the bsonspec and should be round tripped correctly. The comment above saying it will turn into a DBRef is hopefully now wrong with the new driver update.

I'd say keep the DBRef testing in to ensure it still works (it should, a DBRef is just an embedded Document), but also keep DBPointer in there and ensure it also round trips.

divjotarora:
The reason import doesn't have a jsonFormat option is because the Go driver's ext JSON parser *will* parse canonical extended JSON even if canonical is false, so the only necessary distinction is between legacy and non-legacy.

mongoexport/json.go
divjotarora: done in new commit

mongoexport/mongoexport.go
divjotarora: added a jsonFormat type and split up the constants in new commit.

mongoexport/options.go
divjotarora: done in new commit.

patrickfreed:
That seems like unexpected behavior from the go driver to me...are we sure that isn't a bug?

Regardless, that seems like an implementation detail that a user would only know the reason for if they were familiar with the go driver's marshaling api. The problem is that we can't really provide a distinction if the driver doesn't, though.

divjotarora:
I remember talking to @xdg about the --legacy/--jsonFormat issue and I can't exactly recall how we came to this conclusion. Regardless, this is probably something we should all discuss when @xdg is back in the office.

divjotarora:
Current status - @patrickfreed and I talked to Matt about the driver's ext JSON parsing and it seems like the parsing behavior was not intentional. If @xdg agrees that this is a bug and that the driver should be stricter about the user's preference while parsing, we can file a GODRIVER ticket for it (the change is very simple) and then change mongoimport to also take a --jsonFormat flag with options for relaxed, canonical, and legacy. I'm not sure if this would be consider a backwards-breaking change for the driver, so we might also have to consider that side of things.

xdg:
Do we have a decision to take or should I review this as it stands?

divjotarora:
The decision for tools options depends on the decision made for [GODRIVER-963](https://jira.mongodb.org/browse/GODRIVER-963). The solutions I proposed on the ticket and the corresponding tools changes are:

1. Deprecate the `canonical` parameter to the `UnmarsahlExtJSON` function. This requires the fewest driver changes (documentation-only change), but means that import will have `--legacy` and export will have `--jsonFormat` because for import, we cannot specify canonical vs relaxed.

2. Fix the behavior of the `canonical` param (and possibly rename it to `canonicalOnly`). In this case, we can add a symmetric `--jsonFormat` option to import if we think that would be a better idea than having asymmetric flags.

When I last spoke to Jeff about it, he mentioned he preferred deprecating the `canonical` param.

divjotarora:
Spoke to Jeff about the next steps for GODRIVER-963 today. He mentioned that the ticket won't be done immediately, but we can assume that the changes in the ticket will preserve current driver behavior (i.e. the parser will accept both canonical and relaxed and will not require users to specify a format). In this case, I think the current code with the `--legacy` flag for mongoimport can be reviewed as-is.

mongoimport/json_test.go
```diff
@@ -23,21 +25,21 @@ func TestJSONArrayStreamDocument(t *testing.T) {
 		var jsonFile, fileHandle *os.File
 		Convey("an error should be thrown if a plain JSON document is supplied", func() {
 			contents := `{"a": "ae"}`
-			r := NewJSONInputReader(true, bytes.NewReader([]byte(contents)), 1)
+			r := NewJSONInputReader(true, true, bytes.NewReader([]byte(contents)), 1)
```
xdg: Pls add a comment to this test function that we're testing legacy extended JSON, because the next reader isn't going to realize what the second "true" argument means.

mongoimport/options.go
```diff
@@ -44,6 +44,9 @@ type InputOptions struct {
 
 	// Indicates that field names include type descriptions
 	ColumnsHaveTypes bool `long:"columnsHaveTypes" description:"indicated that the field list (from --fields, --fieldsFile, or --headerline) specifies types; They must be in the form of '<colName>.<type>(<arg>)'. The type can be one of: auto, binary, bool, date, date_go, date_ms, date_oracle, double, int32, int64, string. For each of the date types, the argument is a datetime layout string. For the binary type, the argument can be one of: base32, base64, hex. All other types take an empty argument. Only valid for CSV and TSV imports. e.g. zipcode.string(), thumbnail.binary(base64)"`
+
+	// Indicates that the legacy extended JSON format should be used to parse JSON documents. Defaults to false.
+	Legacy bool `long:"legacy" default:"false" description:"use the legacy extended JSON format (defaults to 'false')"`
```
xdg: True should be an error unless `--type=json`.  Please add code and tests for that case.

test/qa-tests/jstests/import/types.js
xdg: Agreed.   I think we should expect DBPointer to be handled accurately, even if deprecated.

mongoimport/options.go
divjotarora: Added in the new commit.

test/qa-tests/jstests/import/types.js
divjotarora: Added a DBPointer in the previous commit before your comments.

mongoimport/mongoimport.go
```diff
@@ -175,6 +175,9 @@ func (imp *MongoImport) validateSettings(args []string) error {
 		if _, err := ValidatePG(imp.InputOptions.ParseGrace); err != nil {
 			return err
 		}
+		if imp.InputOptions.Legacy {
```
xdg: The deeply nested code above this makes me want to cry. 😢

PR #143, opened 2019-04-12
TOOLS-2249 Check if collection exists in mongodump

divjotarora:


mongodump/mongodump.go
```diff
@@ -162,9 +162,31 @@ func (dump *MongoDump) Init() error {
...
+	return collInfo != nil, nil
+}
+
 // Dump handles some final options checking and executes MongoDump.
 func (dump *MongoDump) Dump() (err error) {
 	defer dump.SessionProvider.Close()
+
+	exists, err := dump.verifyCollectionExists()
+	if err != nil || !exists {
+		return err
```
patrickfreed: I think this should be split into two cases: 
- `err != nil` -> return an error message like "couldn't verify collection info" or something
- !exists -> log that the collection doesn't exist, then return nil.

It could be argued that dumping a non-existent collection should be an error case, but I'm not sure if that's true or desired.

divjotarora: I agree there's something to be decided here. I wasn't sure what the right behavior should be. mongoexport has a command line option for asserting that the collection exists and an error is thrown if it doesn't. For now, I'll update it to match your suggestion but I think this should be discussed more when the cmd opts rationalization project is done.

divjotarora:
Yep, `go test -run TestMongoDumpBSON -v` in master with the new test panics.

PR #144, opened 2019-04-12
Minor: improve mongofiles error reporting

patrickfreed:
When I ported mongofiles, I made some weird choices and omissions when handling errors during initialization. This PR fixes them.

### DIFF
--- a/mongofiles/main/mongofiles.go
+++ b/mongofiles/main/mongofiles.go
@@ -40,7 +40,7 @@ func main() {
 	mf, err := mongofiles.New(opts)
 	if err != nil {
 		log.Logv(log.Always, err.Error())
-		if setupErr, ok := err.(mongofiles.SetupError); ok {
+		if setupErr, ok := err.(util.SetupError); ok {
 			if setupErr.Code == util.ExitBadOptions {
 				log.Logvf(log.Always, "try mongofiles --help for more information")
 			}
--- a/mongofiles/main/mongofiles.go
+++ b/mongofiles/main/mongofiles.go
@@ -21,6 +21,7 @@ func main() {
 	opts, err := mongofiles.ParseOptions(os.Args[1:])
 	if err != nil {
 		log.Logv(log.Always, err.Error())
+		log.Logv(log.Always, "try 'mongofiles --help' for more information")
 		os.Exit(util.ExitBadOptions)
 	}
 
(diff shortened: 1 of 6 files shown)


mongofiles/mongofiles.go
```diff
@@ -66,14 +66,24 @@ type MongoFiles struct {
 	bucket *gridfs.Bucket
 }
 
+// SetupError is the error thrown by New to convey what error occurred and the appropriate exit code.
+type SetupError struct {
```
patrickfreed: This was copy/pasted from mongoexport; should it be moved to common? Or is it minor enough to just keep here?

mongofiles/main/mongofiles.go
```diff
@@ -39,6 +40,13 @@ func main() {
 	mf, err := mongofiles.New(opts)
 	if err != nil {
 		log.Logv(log.Always, err.Error())
+		if setupErr, ok := err.(mongofiles.SetupError); ok {
+			if setupErr.Code == util.ExitBadOptions {
+				log.Logvf(log.Always, "try mongofiles --help for more information")
```
divjotarora: We're ignoring the actual error from SetupError in this case. Can we make the log msg include setupErr.Error()?

patrickfreed: we're still logging the error message on line 42, even if the error is a `SetupError`. The additional checking is to see if we should tell the user to use `--help` or not before trying again.

mongofiles/mongofiles.go
patrickfreed: moved to common and filed TOOLS-2260

mongofiles/mongofiles.go
```diff
@@ -84,7 +94,7 @@ func New(opts Options) (*MongoFiles, error) {
 	}
 
 	if err := mf.ValidateCommand(opts.ParsedArgs); err != nil {
```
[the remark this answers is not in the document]
patrickfreed: I agree with you, but since this is intended to be a very minor change (it's not even ticketed), I'd prefer to keep the changes minimal and restricted to just improving the error reporting.

mongofiles/main/mongofiles.go
divjotarora: ah missed that - resolving.

patrickfreed:
TOOLS-2253 will change the error reporting across all of the tools, so I may just incorporate some of the changes in this PR into the implementation of that ticket instead.

patrickfreed:
Yep, so I'll just close this PR. All the changes here are incorporated in the TOOLS-2253.

PR #148, opened 2019-04-17
TOOLS-2259 Properly check for connection establishment before executing

patrickfreed:
This was already approved in the common repository, but there needed to be some changes in the tools that I didn't anticipate. Mainly, some of our unit tests were creating full instances of the tools they were testing, including setting up connection infrastructure. Now that we actually test that infrastructure properly, they were failing because there was no server to connect to. This patch fixes those tests. 

Additionally, mongostat was not checking for an error on connection establishment, which it now does.

### DIFF
--- a/mongofiles/mongofiles_test.go
+++ b/mongofiles/mongofiles_test.go
@@ -608,7 +608,7 @@ func TestDefaultWriteConcern(t *testing.T) {
 	})
 
 	Convey("with no URI and no write concern option", t, func() {
-		mf, err := getMongofilesWithArgs("get", "filename")
+		mf, err := getMongofilesWithArgs("get", "filename", "--port", "33333")
 		So(err, ShouldBeNil)
 		So(mf.SessionProvider.DB("test").WriteConcern(), ShouldResemble, writeconcern.New(writeconcern.WMajority()))
 	})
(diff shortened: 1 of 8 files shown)


mongofiles/mongofiles_test.go
```diff
@@ -137,6 +137,20 @@ func simpleMongoFilesInstanceWithFilenameAndID(command, fname, ID string) (*Mong
...
 }
 
+// simpleMockMongoFilesInstanceWithFilename gets an instance of MongoFiles with no underlying SessionProvider.
+// Use this for tests that don't communicate with the server (e.g. options parsing tests)
+func simpleMockMongoFilesInstanceWithFilename(command, fname string) *MongoFiles {
+	return &MongoFiles{
+		ToolOptions:     toolOptions,
+		InputOptions:    &InputOptions{},
+		StorageOptions:  &StorageOptions{GridFSPrefix: "fs", DB: testDB},
+		SessionProvider: nil,
```
divjotarora: nit: you don't need to initialize things to their zero-values. `SessionProvider` and `Id` can be skipped in the struct initialization.

mongostat/main/mongostat.go
```diff
@@ -237,7 +237,10 @@ func main() {
 	}
 
 	for _, v := range seedHosts {
-		stat.AddNewNode(v)
+		if err := stat.AddNewNode(v); err != nil {
```
divjotarora: Good catch. Can you file a ticket to add errcheck to our CI process? We shouldn't be forgetting errors anywhere.

patrickfreed: Filed https://jira.mongodb.org/browse/TOOLS-2266

mongofiles/mongofiles_test.go
patrickfreed: Fixed. Thanks for pointing this out (again). I'm used to swift where you need to explicitly define an initializer that allows you to omit certain properties.

PR #151, opened 2019-04-25
TOOLS-1956 Use the bulk write API in mongoimport

divjotarora:
The relevant changes in common are in `buffered_bulk.go`.

Evergreen patch: https://evergreen.mongodb.com/version/5cc0f2c79ccd4e1880ffc02d


xdg:
I can't look at this today.  Maybe add @kmahar as another reviewer?

mongoimport/mongoimport.go
```diff
@@ -512,49 +501,25 @@ readLoop:
...
-}
-
-func (imp *MongoImport) newUpserter(collection *mongo.Collection, bypassDocValidation bool) *upserter {
-	return &upserter{
-		imp:         imp,
-		collection:  collection,
-		insertOpts:  driverOpts.InsertOne().SetBypassDocumentValidation(bypassDocValidation),
-		replaceOpts: driverOpts.Replace().SetBypassDocumentValidation(bypassDocValidation).SetUpsert(true),
-		updateOpts:  driverOpts.Update().SetBypassDocumentValidation(bypassDocValidation).SetUpsert(true),
+func (imp *MongoImport) insertDocument(inserter *db.BufferedBulkInserter, document bson.D) error {
```
patrickfreed: I would suggest renaming this to something more generic since we're doing merges and upserts here too, not just inserts. Maybe `importDocument`?

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
```diff
@@ -12,85 +12,99 @@ import (
...
 func (bb *BufferedBulkInserter) resetBulk() {
-	bb.documents = make([]interface{}, 0, bb.docLimit)
-	bb.byteCount = 0
+	bb.writeModels = make([]mongo.WriteModel, 0, bb.docLimit)
 	bb.docCount = 0
 }
 
-// Insert adds a document to the buffer for bulk insertion. If the buffer is
-// full, the bulk insert is made, returning any error that occurs.
+// Insert adds a document to the buffer for bulk insertion. If the buffer is full, the bulk write is done, returning
```
patrickfreed: I think rewording here (and in other places with similar language) to "If the buffer _becomes_ full, the bulk write is _performed_". Note the first change accompanies a semantic change I mention below, and the second one is more of a nit.

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
```diff
@@ -12,85 +12,99 @@ import (
...
+
+func (bb *BufferedBulkInserter) SetUpsert(upsert bool) {
+	bb.upsert = upsert
 }
 
 // throw away the old bulk and init a new one
 func (bb *BufferedBulkInserter) resetBulk() {
-	bb.documents = make([]interface{}, 0, bb.docLimit)
-	bb.byteCount = 0
+	bb.writeModels = make([]mongo.WriteModel, 0, bb.docLimit)
```
patrickfreed: Since we're never resizing this slice, I think we can reuse the memory by doing something like `bb.writeModels = bb.writeModels[:0]`

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
```diff
@@ -12,85 +12,99 @@ import (
...
-	// coded, it should be based on the server's ismaster response.
-	if bb.docCount >= bb.docLimit || bb.byteCount+len(rawBytes) > MaxBSONSize {
+// the bulk write is done, returning any error that occurs.
+func (bb *BufferedBulkInserter) InsertRaw(rawBytes []byte) error {
+	return bb.addModel(mongo.NewInsertOneModel().SetDocument(rawBytes))
+}
+
+func (bb *BufferedBulkInserter) addModel(model mongo.WriteModel) error {
+	var err error
+	if bb.docCount > bb.docLimit {
```
patrickfreed: I think it would be better if we make this check (and subsequent flush) _after_ adding the model. That way, the write occurs as soon as the buffer is full, rather than waiting for the next insert or an explicit flush.

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
divjotarora: Done

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
divjotarora: Done for this and other functions.

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
divjotarora: done

mongoimport/mongoimport.go
divjotarora: done

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
```diff
@@ -12,85 +12,98 @@ import (
...
+	return bb.addModel(mongo.NewReplaceOneModel().SetFilter(selector).SetReplacement(replacement).SetUpsert(bb.upsert))
+}
+
+// InsertRaw adds a document, represented as raw bson bytes, to the buffer for bulk insertion. If the buffer becomes full,
+// the bulk write is performed, returning any error that occurs.
+func (bb *BufferedBulkInserter) InsertRaw(rawBytes []byte) error {
+	return bb.addModel(mongo.NewInsertOneModel().SetDocument(rawBytes))
+}
+
+func (bb *BufferedBulkInserter) addModel(model mongo.WriteModel) error {
```
kmahar: nit: while this is internal, this is where the interesting behavior described in docstrings above is actually happening. might be worth adding docstring here too.

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
```diff
@@ -12,85 +12,98 @@ import (
...
 
-// Insert adds a document to the buffer for bulk insertion. If the buffer is
-// full, the bulk insert is made, returning any error that occurs.
+// Insert adds a document to the buffer for bulk insertion. If the buffer becomes full, the bulk write is performed, returning
+// any error that occurs.
 func (bb *BufferedBulkInserter) Insert(doc interface{}) error {
 	rawBytes, err := bson.Marshal(doc)
 	if err != nil {
-		return fmt.Errorf("bson encoding error: %v", err)
+		return fmt.Errorf("bson encoding err: %v", err)
```
kmahar: changed for consistency, or?

mongoimport/mongoimport.go
```diff
@@ -466,18 +460,12 @@ func (imp *MongoImport) runInsertionWorker(readDocs chan bson.D) (err error) {
...
-
-		inserter = bufferedInserter
-	} else {
-		inserter = imp.newUpserter(collection, imp.IngestOptions.BypassDocumentValidation)
+	inserter := db.NewBufferedBulkInserter(collection, imp.IngestOptions.BulkBufferSize, !imp.IngestOptions.StopOnError)
+	inserter.SetBypassDocumentValidation(imp.IngestOptions.BypassDocumentValidation)
+	if !imp.IngestOptions.MaintainInsertionOrder {
+		inserter.Unordered()
 	}
+	inserter.SetUpsert(true)
```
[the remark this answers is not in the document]
divjotarora: In both the replace and update cases, we want to do upserts. It's ignored by the insert case because the insert command doesn't have an upsert option.

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
divjotarora: I don't remember changing this. Changed back to `error`.

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
divjotarora: done

vendor/github.com/mongodb/mongo-tools-common/db/buffered_bulk.go
```diff
@@ -67,30 +69,43 @@ func (bb *BufferedBulkInserter) Insert(doc interface{}) error {
...
 }
 
-// InsertRaw adds a document, represented as raw bson bytes, to the buffer for bulk insertion. If the buffer is full,
-// the bulk insert is made, returning any error that occurs.
-func (bb *BufferedBulkInserter) InsertRaw(rawBytes []byte) (err error) {
-	// flush if we are full
-	//
-	// XXX With OP_MSG the limit is larger; MaxBSONSize shouldn't be hard
-	// coded, it should be based on the server's ismaster response.
-	if bb.docCount >= bb.docLimit || bb.byteCount+len(rawBytes) > MaxBSONSize {
```
kmahar: do we not need to worry about this case anymore where we have too many bytes?

divjotarora: The driver's bulk write API should take care of batch splitting and ensuring each batch is appropriately sized. In theory, we could get rid of `docLimit` too and do one big bulk write at the end of mongoimport. I chose to leave in `docLimit` because this way, if there are unrecoverable errors, they'll be caught earlier rather than at the very end.

PR #155, opened 2019-05-04
minor: Update README set_goenv instructions

eplanet:
@xdg sorry :-(. You don't have to accept that PR :-D.


eplanet:
Actually there's another problem here: [missing dependencies](https://jira.mongodb.org/browse/TOOLS-1448). Maybe I missed something but it keeps failing because of that.

xdg:
Hi.  llmgo is private, but is vendored, so unless you're trying to update dependencies, I would think it would work.  Could you please give me an [SSCCE](http://sscce.org) so I can repro what you're seeing?

PR #163, opened 2019-07-08
TOOLS-2319 Sanitize URI logging

xdg:
Not all tools log the connection string.  Those that do now sanitize
the URI before logging.


xdg:
Thanks!

PR #165, opened 2019-07-10
TOOLS-2174 Fix dump/export error for autoIndexId:false

xdg:
Trivial fix; most of the CR is test code.

### DIFF
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -832,3 +832,44 @@ func TestMongoDumpOplog(t *testing.T) {
 
 	})
 }
+
+// Test dumping a capped collection with autoIndexId:false.  As of MongoDB 4.0,
+// this is only allowed on the 'local' database.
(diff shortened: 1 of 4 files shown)


mongodump/mongodump_test.go
```diff
@@ -832,3 +832,44 @@ func TestMongoDumpOplog(t *testing.T) {
 
 	})
 }
+
+// Test dumping a capped collection with autoIndexId:false.  As of MongoDB 4.0,
```
kmahar: are we testing with capped for a reason other than that this was the user example? i.e. did this problem exist for non-capped collections with autoIndexId:false, too? just wondering sure if there's some relationship b/w capped collections and autoIndexId that I'm not aware of.

xdg: I tested capped because that was the case, but I spot checked and it applies to non-capped, too.  I'll drop back to non-capped for testing for clarity.

PR #166, opened 2019-07-16
Fix Readme Setup instructions - missing set_goenv

TeNNoX:
It took me quite a while to figure out why it won't build... (Maybe I should read scripts I execute beforehand :yum: )

It could also be done the other way around - just put `set_goenv` at the bottom of the script...


xdg:
Amended and applied as 7f9a7d6a1e66fd9803f3b7fb44d97710fccd0c9e.  Thanks!

PR #168, opened 2019-08-14
Use `EstimatedDocumentCount` instead `CountDocuments` for progress bar render

bozaro:
I am sure that the exact number of records in the progress bar is not important enough to read the collection twice.

### DIFF
--- a/mongodump/prepare.go
+++ b/mongodump/prepare.go
@@ -20,7 +20,6 @@ import (
 	"github.com/mongodb/mongo-tools-common/intents"
 	"github.com/mongodb/mongo-tools-common/log"
 	"github.com/mongodb/mongo-tools-common/util"
-	"go.mongodb.org/mongo-driver/bson"
 )
 
 type NilPos struct{}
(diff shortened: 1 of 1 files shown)


tfogo:
Hi @bozaro, thank you for your contribution. The `intent.Size` isn't actually used in generating the progress bars. It's used to prioritize which collections to start dumping first. But you're still correct, it's not necessary to run `CountDocuments()` for this. 

When I looked into how the progress bars were counting, I found that they were also using `CountDocuments()`. The progress bars are initialized in [mongodump.go#L614-L623](https://github.com/mongodb/mongo-tools/blob/3bb0ea0aaa/mongodump/mongodump.go#L614-L623). The `MongoDump.getCount()` method uses a utility in [mongodb/mongo-tools-common](https://github.com/mongodb/mongo-tools-common) which does call `CountDocuments()`. Based on your suggestion, I've opened a PR to fix that (https://github.com/mongodb/mongo-tools-common/pull/11) and switch it to use `EstimatedDocumentCount()`.

Thanks for your help! 👍

PR #169, opened 2019-08-20
TOOLS-2350 Print the correct output location

tfogo:
Sets `intent.Location` in the default case when `--archive` isn't set. Also moves setting the `intent.Location` to `MongoDump.NewIntentFromOptions()` since it feels like it makes more sense for this to be done at initialization.


tfogo:
I looked into it and it doesn't seem to be a problem.

PR #172, opened 2019-09-11
TOOLS-2268 Add --mode=delete option

tfogo:
When mode is set to delete, mongoimport will delete existing documents in the database with matching documents from the import file. No operation will be performed with non-matching documents. By default, the `_id` field will be used to match documents. Use `--upsertFields` to specify the fields against which to match existing documents.

A couple of notes: 
- The original ticket requested "remove" mode. I thought that "delete" was more in line with the MongoDB API and also sounds a bit more serious (which is a good thing for something that will delete data).
- The `--upsertFields` option is now mis-named. I think `--matchFields` might be a better name.


tfogo:
When I ran `dep ensure`, it added a bunch of missing files to the go driver package. It looks like it didn't change the revision (`d42ea72`), just those parts were missing in the repo for some reason. Any reason that would have happened?

xdg:
I agree on the `--upsertField` comment.  Why don't you open a ticket about it and add it to the CLI rationalization Epic.  I'll look at the code tomorrow.

mongoimport/mongoimport.go
```diff
@@ -490,7 +492,7 @@ readLoop:
 
 func (imp *MongoImport) updateCounts(result *mongo.BulkWriteResult, err error) {
 	if result != nil {
-		atomic.AddUint64(&imp.insertionCount, uint64(result.InsertedCount)+uint64(result.ModifiedCount)+uint64(result.UpsertedCount))
+		atomic.AddUint64(&imp.insertionCount, uint64(result.InsertedCount)+uint64(result.ModifiedCount)+uint64(result.UpsertedCount)+uint64(result.DeletedCount))
```
xdg: I think we should rename `insertionCount`.  What about "successCount" to mirror the existing "failureCount"?

mongoimport/mongoimport_test.go
```diff
@@ -754,6 +765,43 @@ func TestImportDocuments(t *testing.T) {
 			}
 			So(checkOnlyHasDocuments(*imp.SessionProvider, expectedDocuments), ShouldBeNil)
 		})
+		Convey("CSV import with --mode=delete should succeed", func() {
+			// Fist import 3 documetns
```
xdg: Typos.

mongoimport/options.go
```diff
@@ -78,7 +78,7 @@ type IngestOptions struct {
 	// "insert": Insert only, skip existing documents.
 	// "upsert": Insert new documents or replace existing ones.
 	// "merge": Insert new documents or modify existing ones; Preserve values in the database that are not overwritten.
-	Mode string `long:"mode" choice:"insert" choice:"upsert" choice:"merge" description:"insert: insert only. upsert: insert or replace existing documents. merge: insert or modify existing documents. (default: insert)"`
+	Mode string `long:"mode" choice:"insert" choice:"upsert" choice:"merge" choice:"delete" description:"insert: insert only. upsert: insert or replace existing documents. merge: insert or modify existing documents. (default: insert)"`
```
xdg: Need to add "delete" into description as well.

mongoimport/mongoimport_test.go
```diff
@@ -754,6 +765,43 @@ func TestImportDocuments(t *testing.T) {
...
+			So(err, ShouldBeNil)
+
+			imp.InputOptions.Type = CSV
+			imp.InputOptions.File = "testdata/test_delete.csv"
+			fields = "_id,c,b"
+			imp.InputOptions.Fields = &fields
+			imp.IngestOptions.Mode = modeDelete
+			imp.IngestOptions.StopOnError = true
+			// Must specify upsert fields since option parsing is skipped in tests
+			imp.upsertFields = []string{"_id"}
```
xdg: Should add a test that uses a non-ID field to delete -- which gives me an interesting question in that case... are we doing "delete one" or are we doing "delete many"?  To be consistent with other modes, I think we should be intending this to be single-document -- and so we need to document that the update fields should uniquely identify a document.

tfogo: Yes it is "delete one".

mongoimport/mongoimport.go
```diff
@@ -511,6 +513,8 @@ func (imp *MongoImport) importDocument(inserter *db.BufferedBulkInserter, docume
 			result, err = inserter.Insert(document)
```
tfogo: There's a neat little bug where if `--ignoreBlanks` is set, then mongoimport could fallback on insert mode. This causes `--mode delete` to insert documents.

E.g. 

```
./bin/mongoimport -c bar --file ./mongoimport/testdata/test_blanks.csv --type csv --fields a,b,c --ignoreBlanks --mode delete --upsertFields b -v
```

The issue is the `if selector == nil` line. I think the expected behaviour would be to just skip documents where we can't construct the filter.

tfogo:
Will revendor before the merge to fix conflicts

mongoimport/mongoimport.go
```diff
@@ -369,7 +371,7 @@ func (imp *MongoImport) ImportDocuments() (uint64, uint64, error) {
 // importDocuments is a helper to ImportDocuments and does all the ingestion
 // work by taking data from the inputReader source and writing it to the
 // appropriate namespace
-func (imp *MongoImport) importDocuments(inputReader InputReader) (numImported uint64, numFailed uint64, retErr error) {
+func (imp *MongoImport) importDocuments(inputReader InputReader) (uint64, uint64, error) {
```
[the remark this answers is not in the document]
tfogo: Good point, done!

mongoimport/mongoimport.go
```diff
@@ -415,9 +419,9 @@ func (imp *MongoImport) importDocuments(inputReader InputReader) (numImported ui
 	}()
 
 	e1 := channelQuorumError(processingErrChan, 2)
```
tfogo: @xdg as an aside, why do we wait for 2 errors here and not 1?

Gopkg.toml
```diff
@@ -29,11 +29,11 @@ required = ["github.com/3rf/mongo-lint"]
 
 [[constraint]]
   name = "github.com/mongodb/mongo-tools-common"
-  branch = "TOOLS-2268"
+  version = "v1.0.1"
```
xdg: This should be "^v1.0.1" but master already has that so just don't lose it in the conflict resolution.  (No need to fix now as part of this PR.)

mongoimport/mongoimport_test.go
```diff
@@ -792,16 +802,94 @@ func TestImportDocuments(t *testing.T) {
...
+			So(numProcessed, ShouldEqual, 2)
 			So(numFailed, ShouldEqual, 0)
 
 			expectedDocuments := []bson.M{
 				{"_id": int32(3), "c": 5.4, "b": "string"},
 			}
 			So(checkOnlyHasDocuments(*imp.SessionProvider, expectedDocuments), ShouldBeNil)
 		})
+		Convey("CSV import with --mode=delete and --upsertFields should succeed", func() {
+			// Fist import 3 documents
```
xdg: typo

PR #175, opened 2019-10-10
TOOLS-1954 Add --useArrayIndexFields flag

tfogo:
When `--useArrayIndexFields` is set, natural numbers in fields are interpreted as array indexes. For example:

```
a.0,a.1
1,2
```

would result in the document `{a: [1, 2]}`. Indexes must increase sequentially. For example, fields `a.1,a,0` would fail. Fields `a.0,a.2` would also fail since an element is skipped. 

A blank value for a field is inserted into the array as an empty string. E.g.

```
a.0,a.1
,2
```

Would result in the document: `{a: ["",2]}`.

If the first part of a key is a natural number, it is interpreted as a document key since the top level of a BSON document is (of course) a document. E.g.

```
0,1
1,2
```

would result in `{"0": 1, "1": 2}`.

Field names are incompatible if they imply different types of value at a certain key. E.g. the following would be incompatible:

- `a.0` (`a` is an array) and `a.a` (`a` is a document)
-  `a` (`a` is a value) and `a.0` (`a` is an array). Note, this case was already covered by field validation before this PR since `a` is incompatible with `a.x`.


tfogo:
I asked Seth about the flag name. I'll change it if he suggests something else.

mongoimport/common.go
```diff
@@ -414,13 +570,36 @@ func validateFields(fields []string) error {
 				return fmt.Errorf("fields cannot be identical: '%v' and '%v'", field, latterField)
 			}
 		}
+
+		if useArrayIndexFields {
```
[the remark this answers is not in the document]
xdg: I also wonder if this should move into the loop above after 571 (possibly factored out into a function), so that each field only checks fields after it.

mongoimport/mongoimport_test.go
```diff
@@ -1001,9 +1001,142 @@ func TestImportDocuments(t *testing.T) {
...
+				nil,
+			),
+		)
+		Convey("With --useArrayIndexFields: Array fields skipping an index should throw an error",
+			nestedFieldsTestHelper(
+				"_id,a.0,a.2\n1,2,3",
+				nil,
+			),
+		)
+		Convey("With --useArrayIndexFields: Array field should thorw an error if value has already been set as document",
```
xdg: typo

mongoimport/mongoimport_test.go
```diff
@@ -1001,9 +1001,142 @@ func TestImportDocuments(t *testing.T) {
...
+			nestedFieldsTestHelper(
+				"_id,a.a,a.0\n1,2,3",
+				nil,
+			),
+		)
+		Convey("With --useArrayIndexFields: Document field should thorw an error if value has already been set as array",
```
xdg: typo

mongoimport/common.go
```diff
@@ -402,6 +555,9 @@ func validateFields(fields []string) error {
 		// NOTE: since fields is sorted, this check ensures that no field
 		// is incompatible with another one that occurs further down the list.
 		// meant to prevent cases where we have fields like "a" and "a.c"
+
+		// TODO: Add a check here for incompatible fields:
+		// a.b and a.0
```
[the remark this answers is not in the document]
tfogo: Covered below, I'll remove the TODO

mongoimport/common.go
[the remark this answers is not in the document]
tfogo: Yes I think we could move it in. My original implementation had to check the whole array which is why it is where it is. 

The other method I considered was building a tree, where each node had a type associated with it. For each node that's added, you can check the types match for all ancestors.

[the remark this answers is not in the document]
tfogo: Also it looks like this validation is slightly broken. At the top level of a document "0" and "a" should be compatible but this throws an error. So I need to fix that.

tfogo:
I made some improvements to the field validation. It's now a bit more verbose but the time complexity is better. With --useArrayIndexFields, we only compare fields to two other fields during validation. So it's now O(n*m) where m is the average levels of nesting. Without --useArrayIndexFields, we now only compare to one other field, so it's now O(n).

I've also managed to push some more checks into the validation so we can find problems before we start to insert documents. Specifically I added checks for indexing not starting at 0 and for indexes being out of order.

I investigated the behaviour of `strconv.Atoi()` and added some more restrictions on `isNatNum()` as a result. Now key parts in the form of "00001" or "+15558675309" will no longer be interpreted as indexes. I think it's safe to assume key parts in this form aren't intended to be interpreted as indexes since they could be product IDs, phone numbers, etc.

mongoimport/common.go
```diff
@@ -533,13 +544,37 @@ func tokensToBSON(colSpecs []ColumnSpec, tokens []string, numProcessed uint64, i
...
+//
+//     (4). One field implies there is a value, another implies there is an array (e.g. a,a.0). This
+//          check is covered by the check for number 2.
+//     (5). One field implies that there is a document, another implies there is an array.
+//          (e.g. a.b,a.0 or a.b.c,a.0.c)
+//     (6). The indexes for an array don't start from 0 (e.g. a.1,a.2)
+//     (7). Array indexes are out of order (e.g. a.1,a.0 or a.0.b,a.1,a.0.c)
+//     (8). An array is missing an index (e.g. a.0,a.2)
+func validateFields(inputFields []string, useArrayIndexFields bool) error {
+	fields := make([]FieldInfo, len(inputFields), len(inputFields))
```
xdg: nit: you generally don't need the third argument unless you're doing something really specific with memory usage.

mongoimport/common.go
```diff
@@ -552,47 +587,168 @@ func validateFields(fields []string, useArrayIndexFields bool) error {
...
-						// Out of bounds of otherFieldParts
-						break
-					}
-					if otherFieldParts[i] != part {
-						if !(isNatNum(otherFieldParts[i]) == isNatNum(part)) {
-							return fmt.Errorf("fields '%v' and '%v' are incompatible", field, otherField)
-						}
-						break
-					}
+			var nextFieldInfo *FieldInfo
```
xdg: I don't think you need pointers for these: FieldInfo is a small, cheap structure to copy, as the slice is just a pointer-type anyway.  I think the code will be clearer, as it won't make the reader wonder why you're passing by reference and if something is actually depending on mutability.

mongoimport/common.go
```diff
@@ -552,47 +587,168 @@ func validateFields(fields []string, useArrayIndexFields bool) error {
...
+			var previousFieldInfo *FieldInfo
+			if index+1 < len(fields) {
+				nextFieldInfo = &fields[index+1]
+			}
+			if index != 0 {
+				previousFieldInfo = &fields[index-1]
+			}
+			// This checks cases (5), (6), (7), and (8). This is achieved by comparing each field against the
+			// next field and the previous field in the sorted list.
```
xdg: Is it really necessary to compare in both directions instead of just looking at all successive pairs?

tfogo: No, it isn't necessary. I got some tunnel vision on checking the previous value when trying to solve for cases 5 and 6. I've made some changes so now we only compare the next value.

xdg:
I've asked @pmeredit to also review as the prefix collision issue is similar to other things he's done and I'd like an extra pair of eyes.

mongoimport/common.go
```diff
@@ -399,28 +587,130 @@ func validateFields(fields []string) error {
...
+		}
+
+		if useArrayIndexFields {
+			var nextFieldInfo FieldInfo
+			if index+1 < len(fields) {
+				nextFieldInfo = fields[index+1]
+			}
+			// This checks cases (5), (6), (7), and (8). This is achieved by comparing each field against the
+			// next field in the sorted list.
+			err := validateFieldsUsingArrayIndexes(fieldInfo, nextFieldInfo, index, len(fields))
```
xdg: Maybe rename `validateFieldsForArrayIndexes` as a given field may or may not have them?

mongoimport/common.go
```diff
@@ -399,28 +587,130 @@ func validateFields(fields []string) error {
...
+			err = checkAgainstNextField(fieldInfo, nextFieldInfo, i)
+			if err != nil {
+				return err
 			}
-			// NOTE: this means we will not support imports that have fields like
-			// a, a - since this is invalid in MongoDB
-			if field == latterField {
-				return fmt.Errorf("fields cannot be identical: '%v' and '%v'", field, latterField)
+		}
+		if num, isNum := isNatNum(fieldParts[i]); isNum && fieldIndex == 0 && num != 0 {
```
xdg: I don't understand this clause.  The `fieldIndex` test indicates that it only applies to the very first field in the list of sorted fields?!

mongoimport/common.go
```diff
@@ -399,28 +587,130 @@ func validateFields(fields []string) error {
...
-		// NOTE: since fields is sorted, this check ensures that no field
-		// is incompatible with another one that occurs further down the list.
-		// meant to prevent cases where we have fields like "a" and "a.c"
-		for _, latterField := range fieldsCopy[index+1:] {
-			// NOTE: this means we will not support imports that have fields that
-			// include e.g. a, a.b
-			if strings.HasPrefix(latterField, field+".") {
-				return fmt.Errorf("fields '%v' and '%v' are incompatible", field, latterField)
+
+		if index+1 < len(fields) {
```
xdg: nit: I find it easier to read to have "(index) (test) (condition)" , in this case `index < len(fields) -1`

mongoimport/common.go
```diff
@@ -399,28 +587,130 @@ func validateFields(fields []string) error {
...
+		}
+		if num, isNum := isNatNum(fieldParts[i]); isNum && fieldIndex == 0 && num != 0 {
+			// Can't have non-zero index in very first value, this violates case (6).
+			return fmt.Errorf("Array indexes must start from 0. No index found before index '%v' in field '%v'", part, field)
+		}
+	}
+
+	// Second, if the next field has more parts than the current field, loop through
+	// the next field parts and make sure that if they are a number, they are 0.
+	if fieldIndex+1 < numFields {
```
xdg: ditto `fieldIndex < numFields-1` or `!isLastField`

xdg:
I was thinking more about the sort order issue and a comment Patrick made yesterday about doing this unordered with maps.  I think that might be a better approach -- not as memory efficient but I think the code will be easier to verify and I don't think memory is a constraint for the number of fields we'll see in typical usage.

I think Patrick's idea might be to mimic the document structure using `map[string]interface{}` and `[]interface{}` (Patrick, correct me if I'm wrong).  If there's a terminal node, make it `true` or something.  Then for each field, you can walk the shape, constructing it as you go if needed, until you can add the last element of the field as a new terminal node to the structure or until you encounter a conflict.  For array fields, you can look to see how the index compares to an array already there -- if the array doesn't exist, we must have index `0`, or if the array does exist, the index ought to be the next element.  That would require users to put fields in the right order, but alternatively we could relax that and expand the array as necessary, leaving sentinel values in the holes that might be replaced later as we process subsequent fields -- and then at the end we would walk the structure and verify that all arrays no longer have hole-sentinels. 

That avoids both the sort and the "nextField" logic, and does the job in a single pass over fields and a single recursive walk of the structure.

You might want to spike it out and see how it looks.  I bet it would come together quickly.

xdg:
Thinking more about it, that approach is actually almost exactly what you're doing to build up the document to insert -- the difference would be writing sentinel values into array holes.  Imagine filling an array hole with some new ErrMissingValue constant as you construct the document to insert, then just walk it at the end.  I'm not sure we want to actually do that because checking for duplicate values (a,a) would be expensive with bson.D and switching to bson.M would mean documents would have fields inserted in random orders instead of having the same shape -- but conceptually it's the same problem you've already solved.

tfogo:
It might be worth asking, do we even need to perform these validation checks before moving onto the insertion logic. The errors will be thrown when constructing the documents. In the best case scenario, an error will be thrown when building the first document, resulting in a failure and no changes in the DB. If `--ignoreBlanks` is set, then it's possible that many documents could be inserted before the first problem arises. This would be an awkward situation for the user to fix. But maybe it's an acceptable situation.

Either way, I'm happy to change the strategy to filling array holes. This would have the impact of allowing indexes to be out of order in the source: e.g. a.1,a.0

mongoimport/common.go
```diff
@@ -252,30 +260,183 @@ func removeBlankFields(document bson.D) (newDocument bson.D) {
...
+// of setNestedValue for more information.
+func setNestedArrayValue(fieldParts []string, value interface{}, array *bson.A) error {
+
+	// The first part of the field should be an index of an array
+	idx, err := strconv.Atoi(fieldParts[0])
+	if err != nil {
+		return fmt.Errorf("setNestedArrayValue expected an integer field, but instead received %s", fieldParts[0])
+	}
+
+	v := reflect.ValueOf(*array)
```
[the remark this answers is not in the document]
tfogo: I think maybe I was using reflection for something else and I never changed it back

mongoimport/common.go
```diff
@@ -380,13 +544,37 @@ func tokensToBSON(colSpecs []ColumnSpec, tokens []string, numProcessed uint64, i
...
+// are invalid, returns nil otherwise. Fields are invalid in the following cases:
+//
+//     (1). A field contains an invalid series of characters
+//     (2). Two fields are the same (e.g. a,a)
+//     (3). One field implies there is a value, another implies there is a document (e.g. a,a.b)
+//
+// In the case that --useArrayIndexFields is set, fields are also invalid in the following cases:
+//
+//     (4). One field implies there is a value, another implies there is an array (e.g. a,a.0). This
+//          check is covered by the check for number 2.
```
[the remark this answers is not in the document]
tfogo: Yes, you're right it should say number 3

mongoimport/common.go
tfogo: Yes, because `checkAgainstNextField()` actually checks case (6) (indexes must start with a zero) for the next field, not the current field. If we want to check for the current field, we need to look back at the previous field. This means the very first field doesn't get checked. That check occurs here.

mongoimport/common.go
```diff
@@ -380,13 +544,37 @@ func tokensToBSON(colSpecs []ColumnSpec, tokens []string, numProcessed uint64, i
...
+	fields := make([]FieldInfo, len(inputFields))
+	for i, f := range inputFields {
+		fields[i] = FieldInfo{i, f, strings.Split(f, ".")}
+	}
+
+	// By using the FieldInfo type, once fields are sorted alphabetically, we maintain the
+	// information about the original position of fields. This is used to check that array
+	// indexes aren't out of order. Sorting the list of fields means for each field we only
+	// have to compare it against two other fields at most to ensure validity of the whole set.
+	sort.Slice(fields, func(i, j int) bool { return fields[i].field < fields[j].field })
```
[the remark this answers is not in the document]
tfogo: Damn! That's really bad. Thanks for catching it.

xdg:
I think pre-validation is still worthwhile, because the duplicate key check will be "expensive" in bson.D -- though OTOH, we only need to do that check (and the array-holes checks) for the *first* document and then skip it after that as we've effectively validated the fields at that point -- so maybe we can do it as we insert rather than as a separate pass.  @pmeredit, what do you think?

I don't follow how the `ignoreBlanks` option matters -- whether there's a value or not, I think we can assume it to be a terminal node for the purpose of validation during first-document construction.  In the weird case that an entire column is blank and that would avoid a conflict if the field name conflicted, I think we can safely error because the existing field name validation would error.

tfogo:
>  though OTOH, we only need to do that check (and the array-holes checks) for the first document and then skip it after that as we've effectively validated the fields at that point -- so maybe we can do it as we insert rather than as a separate pass.

The checks already happen during constructing a document before inserting it. So `tokensToBSON` throws an error if there are any problems with incompatible fields... these cases are already caught by `setNestedValue`/`setNestedArrayValue`.

The issue there is that if `--ignoreBlanks` is true, some documents might get through `tokensToBSON` since we don't go through the logic of adding a field to a document if the value is blank.

e.g. in this case, the first two records will be inserted, and an error will only be thrown on the third record:

```
a.b,a.0
1,
2,
3,4
```

One possibility is to use `tokensToBSON` for a validation run before inserting documents. We create a list of tokens that is all 1s (for example) and pass that to `tokensToBSON` with the column spec. This is probably not as efficient as a purpose-written validation function, but I do like that we get to reuse the code.

xdg:
I think I lean towards a separate validation routine on the grounds that it will leave `tokensToBSON` uncluttered and thus easier to maintain.  Two focused routines that are easy to read and understand are more maintainable than one routine with a lot of internal branching.

mongoimport/common.go
```diff
@@ -380,13 +544,37 @@ func tokensToBSON(colSpecs []ColumnSpec, tokens []string, numProcessed uint64, i
...
+//
+//     (4). One field implies there is a value, another implies there is an array (e.g. a,a.0). This
+//          check is covered by the check for number 2.
+//     (5). One field implies that there is a document, another implies there is an array.
+//          (e.g. a.b,a.0 or a.b.c,a.0.c)
+//     (6). The indexes for an array don't start from 0 (e.g. a.1,a.2)
+//     (7). Array indexes are out of order (e.g. a.1,a.0 or a.0.b,a.1,a.0.c)
+//     (8). An array is missing an index (e.g. a.0,a.2)
+func validateFields(inputFields []string, useArrayIndexFields bool) error {
+	fields := make([]FieldInfo, len(inputFields))
```
pmeredit: As we discussed in the meeting, we're going to try an approach that builds a trie with embedded arrays (where necessary). It will be able to handle this linearly as long as we require that array indexes always be specified in order by the user.

tfogo:
I reimplemented validation by constructing a tree of fields. Also added a few more test cases. Let me know what you think.

By the way, we could get rid of the restriction that array indexes must be in order in the source file. We would have to sort all the tokens and colSpecs for every document when useArrayIndexFields is set. This would add some overhead for processing every document so do we think the increased flexibility is worth the overhead?

pmeredit:
> 
> 
> I was thinking more about the sort order issue and a comment Patrick made yesterday about doing this unordered with maps. I think that might be a better approach -- not as memory efficient but I think the code will be easier to verify and I don't think memory is a constraint for the number of fields we'll see in typical usage.
> 
> I think Patrick's idea might be to mimic the document structure using `map[string]interface{}` and `[]interface{}` (Patrick, correct me if I'm wrong). If there's a terminal node, make it `true` or something. Then for each field, you can walk the shape, constructing it as you go if needed, until you can add the last element of the field as a new terminal node to the structure or until you encounter a conflict. For array fields, you can look to see how the index compares to an array already there -- if the array doesn't exist, we must have index `0`, or if the array does exist, the index ought to be the next element. That would require users to put fields in the right order, but alternatively we could relax that and expand the array as necessary, leaving sentinel values in the holes that might be replaced later as we process subsequent fields -- and then at the end we would walk the structure and verify that all arrays no longer have hole-sentinels.
> 
> That avoids both the sort and the "nextField" logic, and does the job in a single pass over fields and a single recursive walk of the structure.
> 
> You might want to spike it out and see how it looks. I bet it would come together quickly.

Yes, exactly!

pmeredit:
> 
> 
> Thinking more about it, that approach is actually almost exactly what you're doing to build up the document to insert -- the difference would be writing sentinel values into array holes. Imagine filling an array hole with some new ErrMissingValue constant as you construct the document to insert, then just walk it at the end. I'm not sure we want to actually do that because checking for duplicate values (a,a) would be expensive with bson.D and switching to bson.M would mean documents would have fields inserted in random orders instead of having the same shape -- but conceptually it's the same problem you've already solved.

Yes, I was thinking the same thing actually, we could actually construct the first document while doing validation, but I think that's a little unnecessary as a micro-optimization :)

pmeredit:
> 
> 
> I reimplemented validation by constructing a tree of fields. Also added a few more test cases. Let me know what you think.
> 
> By the way, we could get rid of the restriction that array indexes must be in order in the source file. We would have to sort all the tokens and colSpecs for every document when useArrayIndexFields is set. This would add some overhead for processing every document so do we think the increased flexibility is worth the overhead?

I don't think the increased overhead or complexity of the code is worth it. I think in general, people should at least put their array indexes in order. It would be too easy for a user to accidentally miss an index if they specified them out of order, anyway.

mongoimport/common.go
```diff
@@ -399,28 +570,201 @@ func validateFields(fields []string) error {
...
+	}
+	return nil
+}
+
+// addFieldToTree is a recursive function that builds up a tree of fields. It is used to check
+// that fields are compatible with each other. When useArrayIndexFields is set, it is mutually recursive
+// with addFieldToArray(). It closely mimics the behaviour of setNestedDocumentValue(). See validateFields()
+// for more information on the validity checks that are made when constructing a tree of fields.
+func addFieldToTree(fieldParts []string, partIndex int, tree *map[string]interface{}, useArrayIndexFields bool) (err error) {
+	value, exists := (*tree)[fieldParts[partIndex]]
```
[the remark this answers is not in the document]
tfogo: Ah yeah, same with the slices. I'll update that.

tfogo:
I updated addFieldToTree and addFieldToArray so we pass maps and slices by value. I think it makes it a bit more readable. bson.A's underlying type is a slice. So we could do the same thing with `setNestedArrayValue`. But I thought it might be best to keep it as is to maintain the symmetry with `setNestedDocumentValue`.

mongoimport/common.go
```diff
@@ -399,28 +570,197 @@ func validateFields(fields []string) error {
...
+				"Array indexes in fields must start from 0 and increase sequentially",
+				strings.Join(fieldParts, "."),
+				strings.Join(fieldParts[:partIndex], "."),
+				(len(array) - 1))
+		}
+
+		array = append(array, true)
+		return array, nil
+	}
+
```
xdg: At this point, for a non-terminal node, we must check `idx <= len(array)` -- either it exists or it must be the next sequential index.  This checks for the case 'a.0.b, a.2.b'.

mongoimport/common.go
```diff
@@ -399,28 +570,197 @@ func validateFields(fields []string) error {
...
 		}
+		tree[fieldParts[0]] = subTree
 	}
-	return nil
+	return tree, nil
+
+}
+
+// addFieldToArray is used with addFieldToTree() to build a valid tree of fields.
+func addFieldToArray(fieldParts []string, partIndex int, array []interface{}) ([]interface{}, error) {
```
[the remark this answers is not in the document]
pmeredit: I'm partial to `headIndex`

mongoimport/common.go
```diff
@@ -399,28 +570,197 @@ func validateFields(fields []string) error {
...
+
+// addFieldToTree is a recursive function that builds up a tree of fields. It is used to check
+// that fields are compatible with each other. When useArrayIndexFields is set, it is mutually recursive
+// with addFieldToArray(). It closely mimics the behaviour of setNestedDocumentValue(). See validateFields()
+// for more information on the validity checks that are made when constructing a tree of fields.
+func addFieldToTree(fieldParts []string, partIndex int, tree map[string]interface{}, useArrayIndexFields bool) (map[string]interface{}, error) {
+	value, exists := tree[fieldParts[partIndex]]
+
+	if exists {
+		if len(fieldParts[partIndex:]) == 1 {
```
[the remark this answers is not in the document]
tfogo: Yep, you're correct. Great catch!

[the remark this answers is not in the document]
tfogo: Except I think the code should read like: "given that head is terminal: if value is terminal it's a dup, otherwise it's incompatible". Otherwise you miss the case that head is terminal and it is incompatible with a subtree/array that already exists in the tree.

mongoimport/common.go
```diff
@@ -399,28 +570,204 @@ func validateFields(fields []string) error {
...
+	fieldPrefix += "." + head
+
+	// The first part of the field should be an index of an array
+	headIndex, ok := isNatNum(head)
+	if !ok {
+		// We shouldn't ever get here
+		panic(fmt.Sprintf("addFieldToArray expected an integer field, but instead received %s", fieldParts[0]))
+	}
+
+	if len(fieldParts) == 1 {
```
xdg: `len(tail) == 0`

mongoimport/common.go
```diff
@@ -399,28 +570,204 @@ func validateFields(fields []string) error {
...
+		// We're at the terminus of a field so we have to check if we can append to the array
+		if headIndex == len(array) {
+			array = append(array, true)
+			return array, nil
+		}
+		return nil, indexError(fullField)
+	}
+
+	if headIndex > len(array) {
```
xdg: Suggest adding a brief comment  above explaining this case in English. (e.g. "headIndex must be an existing element or be adding an element at the end"?)

mongoimport/common.go
```diff
@@ -399,28 +570,204 @@ func validateFields(fields []string) error {
...
+		return nil, indexError(fullField)
+	}
+
+	if _, ok := isNatNum(tail[0]); ok {
```
xdg: Suggest adding a similar explanatory comment as discussed at line 624.

tfogo:
🎉Thanks for the help all!

PR #176, opened 2019-10-14
Fix for the build error

PhilStollery:
Sorry I didn't see a bug fix branch - so fixed in main. 

Error: 

Building bsondump...
# github.com/mongodb/mongo-tools/vendor/github.com/mongodb/mongo-tools-common/db
vendor/github.com/mongodb/mongo-tools-common/db/db.go:204:20: crt.Subject.String undefined (type pkix.Name has no field or method String)
Error building bsondump

I'm getting a build error - this fixes it because crt.Subject is an enum pkix.Name, so just returning CommonName

### DIFF
--- a/vendor/github.com/mongodb/mongo-tools-common/db/db.go
+++ b/vendor/github.com/mongodb/mongo-tools-common/db/db.go
@@ -201,7 +201,7 @@ func addClientCertFromFile(cfg *tls.Config, clientFile, keyPasswd string) (strin
 		return "", err
 	}
 
-	return crt.Subject.CommonName, nil
+	return crt.ToRDNSequence().String(), nil
 }
 
 // addCACertFromFile adds a root CA certificate to the configuration given a path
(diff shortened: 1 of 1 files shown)


PhilStollery:
> Hi, Phil. The `String` method was added in Go 1.10 (see [release notes](https://golang.org/doc/go1.10#crypto/x509/pkix)). Is there a reason you need to build with an unsupported version of Go?

I'm working on Azure using their cloud shell, need to build the mongo tools. But in the cloud shell they only go up to go1.9. 

There is no way to upgrade go in the shell, we don't have the permissions.

tfogo:
Hi Phil, I'm going through our old PRs to try to resolve them. 

To solve this we would have to maintain our own version of [`RDNSequence.String()`](https://golang.org/src/crypto/x509/pkix/pkix.go?s=6036:6065#L40). We don't want to have to maintain a separate copy of this for an older version of Go. We no longer test the tools with Go 1.9 and the Go team stopped supporting 1.9 in 2018.

We have updated the README as part of #270 so we no longer say we support 1.9.

Sorry I don't have better news here. Thank you for opening the PR, but I'm going to close it now.

PR #177, opened 2019-10-30
TOOLS-2379 Add fallback to applyOps when createIndexes fails

tfogo:
Evergreen: https://evergreen.mongodb.com/version/5db9c62f2fbabe7944a10c77


mongorestore/testdata/index_collation/test/foo.metadata.json
```diff
@@ -0,0 +1 @@
+{"indexes":[{"v":{"$numberInt":"2"},"ns":"test.foo","name":"_id_","collation":{"alternate":"non-ignorable","normalization":false,"caseFirst":"off","strength":{"$numberInt":"2"},"caseLevel":false,"numericOrdering":false,"locale":"en_US","backwards":false,"maxVariable":"punct","version":"57.1"},"key":{"_id":{"$numberInt":"1"}}},{"ns":"test.foo","weights":{"name":{"$numberInt":"1"}},"language_override":"language","key":{"_fts":"text","_ftsx":{"$numberInt":"1"}},"v":{"$numberInt":"2"},"name":"example_text","textIndexVersion":{"$numberInt":"3"},"default_language":"english"}],"options":{"collation":{"maxVariable":"punct","version":"57.1","backwards":false,"locale":"en_US","strength":{"$numberInt":"2"},"caseLevel":false,"numericOrdering":false,"alternate":"non-ignorable","caseFirst":"off","normalization":false}}}
```
xdg: What is this testing for?

tfogo: `TestIndexGetsSimpleCollation` in `metadata_test.go`. This is unrelated to TOOLS-2379. Several files in testdata were untracked. It looked like this was an accident due to some overzealous rules in the .gitignore so I added them all. The .gitignore includes `*.json` and `*.bson`. That seems a bit broad, particularly when there are a lot of test cases that use json/bson files. Would it be okay to remove those from the .gitignore?

xdg: I'm pretty sure that data is generated from [this code](https://github.com/mongodb/mongo-tools/blob/0660bca4bee742d34d34c8cdbf91f4bb6ee552ba/mongorestore/metadata_test.go#L195-L203).  I don't like that there's generated data getting mixed up in the source code directories, but that's the way the tests were designed.  I think the aggressive .gitignore rules are exactly to prevent them from getting committed.  Maybe a better naming scheme would help so it's clearer what's generated.  I'll open a ticket about it.  But in this case, I don't think the file should be committed.

tfogo: Ah that makes sense. I'll take it out of this PR then and we can deal with TOOLS-2409 in another PR.

tfogo:
Looks like qa-tests-unstable is failing across the board. Is that expected or has something gone wrong?

PR #178, opened 2019-11-04
TOOLS-2412 Add --convertLegacyIndexes flag

tfogo:


mongorestore/options.go
```diff
@@ -65,21 +65,22 @@ func (*InputOptions) Name() string {
...
-	BypassDocumentValidationOption = "--bypassDocumentValidation"
-	PreserveUUIDOption             = "--preserveUUID"
-	TempUsersCollOption            = "--tempUsersColl"
-	TempRolesCollOption            = "--tempRolesColl"
-	BulkBufferSizeOption           = "--batchSize"
+	DropOption                      = "--drop"
+	DryRunOption                    = "--dryRun"
+	WriteConcernOption              = "--writeConcern"
+	NoIndexRestoreOption            = "--noIndexRestore"
+	IgnoreInvalidIndexOptionsOption = "--ignoreInvalidIndexOptions"
```
xdg: I think if this is stripping invalid options, we should rename the option to "--removeInvalidIndexOptions".

mongorestore/options.go
```diff
@@ -88,19 +89,20 @@ type OutputOptions struct {
...
-	BypassDocumentValidation bool   `long:"bypassDocumentValidation" description:"bypass document validation"`
-	PreserveUUID             bool   `long:"preserveUUID" description:"preserve original collection UUIDs (off by default, requires drop)"`
-	TempUsersColl            string `long:"tempUsersColl" default:"tempusers" hidden:"true"`
-	TempRolesColl            string `long:"tempRolesColl" default:"temproles" hidden:"true"`
-	BulkBufferSize           int    `long:"batchSize" default:"1000" hidden:"true"`
+	WriteConcern              string `long:"writeConcern" value-name:"<write-concern>" default-mask:"-" description:"write concern options e.g. --writeConcern majority, --writeConcern '{w: 3, wtimeout: 500, fsync: true, j: true}'"`
+	NoIndexRestore            bool   `long:"noIndexRestore" description:"don't restore indexes"`
+	IgnoreInvalidIndexOptions bool   `long:"ignoreInvalidIndexOptions" description:"Ignores invalid options in index specifications" hidden:"true"`
```
xdg: I think given this could be an issue for anyone restoring to shared, we shouldn't make it hidden.

mongorestore/restore.go
```diff
@@ -25,6 +25,33 @@ import (
 
 const insertBufferFactor = 16
 
+var indexOptions = map[string]bool{
+	"bits": true,
+	"min":  true,
+	"max":  true,
+	"coarsestIndexedLevel": true,
```
xdg: This (and finest) aren't in the documented index options.  Can you find out if they would actually appear in an index definition document?

mongorestore/restore.go
```diff
@@ -25,6 +25,33 @@ import (
 
 const insertBufferFactor = 16
 
+var indexOptions = map[string]bool{
```
xdg: Please add a comment about which source file this is taken from.  Also, please rename to something like `validIndexOptions` for clarity.

mongorestore/options.go
xdg: Though if we're also going to fix up the index key document, maybe we need a more inclusive name.  "convertLegacyIndexes"?

mongorestore/restore.go
tfogo: It's undocumented but they can appear. I found an old JIRA ticket which suggests Lyft uses them. I'll open a DOCS ticket about it.

tfogo:
Hey @xdg I'm still working on figuring out how indexes used to deal with different edge cases in 3.2. But I thought you might want to just review this as it is currently. Might add some different cases to the switch statement depending on what I find in my investigations.

mongorestore/restore.go
```diff
@@ -25,24 +25,25 @@ import (
...
-	"dropDups":             true,
-	"expireAfterSeconds":   true,
-	"bucketSize":           true,
-	"name":                 true,
-	"v":                    true,
-	"key":                  true,
-	"language_override":    true,
-	"ns":                   true,
+// validIndexOptions are taken from https://github.com/mongodb/mongo/blob/master/src/mongo/db/index/index_descriptor.h
+var validIndexOptions = map[string]bool{
```
xdg: nit: I suggest sorting the list of options below so the order isn't so random.  (Keeps readers from wondering if there is any meaning to the order.)

mongorestore/restore.go
```diff
@@ -25,24 +25,25 @@ import (
...
+	"bits":                    true,
+	"min":                     true,
+	"max":                     true,
+	"coarsestIndexedLevel":    true,
+	"finestIndexedLevel":      true,
+	"2dsphereIndexVersion":    true,
+	"background":              true,
+	"collation":               true,
+	"default_language":        true,
+	"dropDups":                true,
```
xdg: IIUC, `dropDups` isn't currently valid.  I wonder if you need to make this map version-specific as it may change over time. Or populate it with some version-specific data at startup.

mongorestore/restore.go
```diff
@@ -331,11 +332,29 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
+		for j, elem := range index.Key {
+			switch elem.Value {
+			case true:
+				md := bsonutil.MarshalD(indexes[i].Key)
+				indexJSON, err := md.MarshalJSON()
+				if err != nil {
+					indexJSON = []byte("<error formatting index>")
+				}
+				indexes[i].Key[j].Value = 1
+				log.Logvf(log.Always, "convertLegacyIndexes: converting the legacy vaule of '%s' from '%s' to '%s' in index '%s'", indexes[i].Key[j].Key, "true", "1", string(indexJSON))
```
xdg: Rather than doing this logging element by element, I suggest just flagging if the Key changed and then logging the whole Key document at the end if so.

mongorestore/restore.go
```diff
@@ -331,11 +332,29 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
 	return result
 }
 
-func stripInvalidOptions(indexes []IndexDocument) []IndexDocument {
+func convertLegacyIndexes(indexes []IndexDocument) []IndexDocument {
 	for i, index := range indexes {
+		for j, elem := range index.Key {
```
xdg: Please factor this out into a `convertLegacyIndexKey` (and later `convertLegacyIndexOptions`).  Otherwise, code complexity is high, hurting readability.  Consider how easy this is to follow (also added an error return -- maybe worth considering):

```
func convertLegacyIndexes(indexes []IndexDocument) ([]IndexDocument, error) {
    var err error
    for i, index := range indexes {
        indexes[i], err = convertLegacyIndexKeys(index)
        if err != nil {
             return nil, err
        }
        indexes[i], err = convertLegacyIndexOptions(index)
        if err != nil {
             return nil, err
        }
    }
    return indexes, nil
}

File: mongorestore/restore.go [hunk 9]
```diff
@@ -331,11 +332,29 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
+
 		for key := range index.Options {
-			if _, ok := indexOptions[key]; !ok {
+			if _, ok := validIndexOptions[key]; !ok {
+				optionsJSON, err := bson.MarshalExtJSON(indexes[i].Options, true, false)
+				if err != nil {
+					optionsJSON = []byte("<error formatting index>")
+				}
 				delete(indexes[i].Options, key)
+				log.Logvf(log.Always, "convertLegacyIndexes: removing invalid option '%v' from index options '%v'", key, string(optionsJSON))
```
xdg: Same.  I suggest saving all the logging for after all options are updated.  Rather than saying "removed this option", you can just say "updated options: Old: <old>; New: <new>".

tfogo:
@xdg, I implemented your suggestions and also added logic to deal with index keys that include "", 0, and types other than numbers and strings.

I had to make a function to check if a Decimal128 type is 0. That probably belongs in the driver or in bsonutil. Could you please check it's correct though? I believe the trailing 113 bits should all be zero if the value is 0.

mongorestore/restore.go
```diff
@@ -332,26 +332,87 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
+func convertLegacyIndexes(indexes []IndexDocument) {
+	for _, index := range indexes {
+		convertLegacyIndexKeys(index)
+		convertLegacyIndexOptions(index)
+	}
+}
+
+func convertLegacyIndexKeys(index IndexDocument) {
+	var converted bool
+	originalJSONString := createExtJSONString(index.Key)
```
xdg: you can move this into the "if converted" stanza, as well, as it's only needed conditionally.

mongorestore/restore.go
```diff
@@ -301,6 +332,89 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
+	}
+	if converted {
+		newJSONString := createExtJSONString(index.Options)
+		log.Logvf(log.Always, "convertLegacyIndexes: converted index options '%s' to '%s'",
+			originalJSONString, newJSONString)
+	}
+}
+
+func createExtJSONString(doc interface{}) string {
+	// by default return "<unable to format document>"" since we don't
```
tfogo: I made a call here not to throw an error when formatting informational messages. Let me know if you think we should be more strict.

mongorestore/restore.go
```diff
@@ -332,26 +332,87 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
+	}
+}
 
-		for key := range index.Options {
-			if _, ok := validIndexOptions[key]; !ok {
-				delete(indexes[i].Options, key)
-				log.Logvf(log.Always, "convertLegacyIndexes: removing invalid option '%s' from index '%s' on collection '%s",
-					key, indexes[i].Options["name"], indexes[i].Options["ns"])
-			}
+func isZero(num primitive.Decimal128) bool {
```
xdg: I would avoid all this and just compare to `primitive.ParseDecimal128("0")`.  Odds are no-one has ever created an index with a decimal 128 anyway.  (I'm intentionally ignoring that Decimal 128 "0.00" is different than "0" or "-0", etc.)

mongorestore/restore.go
```diff
@@ -332,26 +332,87 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
+		newJSONString := createExtJSONString(index.Options)
+		log.Logvf(log.Always, "convertLegacyIndexes: converted index options '%s' to '%s'",
+			originalJSONString, newJSONString)
+	}
+}
+
+func createExtJSONString(doc interface{}) string {
+	// by default return "<unable to format document>"" since we don't
+	// want to throw an error when formatting informational messages.
+	// An error would be inconsequential.
```
xdg: 👍

mongorestore/restore.go
tfogo: But index.Key is a slice and I'm editing the underlying array elements. I think this is less intrusive than making a full copy of the array.

xdg: right. nevermind. :-)

PR #183, opened 2020-01-15
TOOLS-2453: Fix invalid key escaping in dump metadata

rychipman:
This just adds a test and revendors the go driver, since it is an extjson marshaling bug.


mongodump/mongodump_test.go
```diff
@@ -243,6 +243,14 @@ func setUpMongoDumpTestData() error {
...
 				return err
 			}
+
+			idx := mongo.IndexModel{
+				Keys: bson.M{`"`: 1},
+			}
+			_, err = coll.Indexes().CreateOne(context.Background(), idx)
+			if err != nil {
+				return err
+			}
```
rychipman: Please let me know if there's a better way of adding this to the dump testing data!

rychipman:
failing evg patch from before fix: https://evergreen.mongodb.com/version/5e1f617c850e6149bc532fc4
passing evg patch from after fix: https://evergreen.mongodb.com/version/5e1f619b1e2d1710df197912

PR #184, opened 2020-01-28
TOOLS-2443: Create zip and tarball archives in dist task

rychipman:
In the spirit of this morning's retro discussion, I tried to push as much of this as possible into go code instead of tacking more functionality onto the evergreen config or putting it into shell scripts.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -701,6 +701,49 @@ functions:
       content_type: application/octet-stream
       display_name: ${tool}
 
+  "create release artifacts":
+    # build release artifacts
+    - command: shell.exec
+      params:
+        working_dir: src/github.com/mongodb/mongo-tools
+        script: |
+          . ./set_goenv.sh
+          GOROOT="" set_goenv || exit
+          export EVG_PLATFORM=${_platform}
+          go run release/release.go build-archive
+
+    # upload individual release artifacts to task page
+    - command: s3.put
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_files_include_filter:
+          - src/github.com/mongodb/mongo-tools/mongodb-cli-tools-*
+        remote_file: mongo-tools/pkgs/${build_id}/
+        content_type: application/octet-stream
+        bucket: mciuploads
+        permissions: public-read
+        display_name: "Release Artifact - "
+
+    # pack all release artifacts into a tarball and upload them to one
+    # place so that they can be more easily downloaded by the sign task
+    - command: archive.targz_pack
+      params:
+        target: src/github.com/mongodb/mongo-tools/upload.tgz
+        source_dir: src/github.com/mongodb/mongo-tools
+        include:
+          - "./mongodb-cli-tools-*"
+    - command: s3.put
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file: src/github.com/mongodb/mongo-tools/upload.tgz
+        remote_file: mongo-tools/pkgs/${build_id}/all-release-artifacts.tgz
+        content_type: application/x-gzip
+        bucket: mciuploads
+        permissions: public-read
+        display_name: All Release Artifacts (.tgz)
+
   "fetch source" :
     - command: shell.exec
       params:
@@ -975,6 +1018,7 @@ tasks:
     - func: "upload tool"
       vars:
         tool: mongotop
+    - func: "create release artifacts"
 
 - name: integration-4.0
   commands:
@@ -1720,6 +1764,7 @@ buildvariants:
   - amazon1-2018-test
   expansions:
     build_tags: ""
+    _platform: amazon1
   tasks:
   - name: dist
   - name: replay-dist
@@ -1730,6 +1775,7 @@ buildvariants:
   - amazon1-2018-test
   expansions:
     build_tags: "sasl gssapi ssl"
+    _platform: amazon1
   tasks:
   - name: dist
   - name: replay-dist
@@ -1740,6 +1786,7 @@ buildvariants:
   - amazon2-test
   expansions:
     build_tags: ""
+    _platform: amazon2
   tasks:
   - name: dist
   - name: replay-dist
@@ -1750,6 +1797,7 @@ buildvariants:
   - amazon2-test
   expansions:
     build_tags: "sasl gssapi ssl"
+    _platform: amazon2
   tasks:
   - name: dist
   - name: replay-dist
@@ -1764,6 +1812,7 @@ buildvariants:
   - debian81-test
   expansions:
     build_tags: ""
+    _platform: debian81
   tasks:
   - name: dist
   - name: replay-dist
@@ -1774,6 +1823,7 @@ buildvariants:
   - debian81-test
   expansions:
     build_tags: "sasl gssapi ssl"
+    _platform: debian81
   tasks:
   - name: dist
   - name: replay-dist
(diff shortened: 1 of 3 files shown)


rychipman:
evergreen patch: https://evergreen.mongodb.com/version/5e30a3eec9ec44566291d37e

release/platform/platform.go
```diff
@@ -0,0 +1,135 @@
...
+
+const evgPlatformVar = "EVG_PLATFORM"
+
+// Get returns the Platform for this host, based on the value of
+// EVG_PLATFORM. It returns an error if EVG_PLATFORM is unset or set
+// to an unknown value.
+func Get() (Platform, error) {
+	evgPlatform := os.Getenv(evgPlatformVar)
+	if evgPlatform == "" {
+		return Platform{}, errors.New("EVG_PLATFORM not set")
```
xdg: minor: if you're going to have a constant on line 18, you should probably use it here, too, for consistency in case it ever changes.

release/platform/platform.go
```diff
@@ -0,0 +1,135 @@
...
+	"debian81": {
+		Name: "debian8",
+		Arch: "x86_64",
+	},
+	"debian92": {
+		Name: "debian9",
+		Arch: "x86_64",
+	},
+	"macos1014": {
+		Name: "macos1014",
```
xdg: I believe the server uses "macos" for the name that shows up in the tarball -- i.e. not version specific.  I think that makes sense for tools as well.

release/platform/platform.go
```diff
@@ -0,0 +1,135 @@
...
+	"ubuntu1604": {
+		Name: "ubuntu1604",
+		Arch: "x86_64",
+	},
+	"ubuntu1804": {
+		Name: "ubuntu1804",
+		Arch: "x86_64",
+	},
+	"windowsVS2017": {
+		Name: "windowsVS2017",
```
xdg: Likewise, maybe this should be like the server: "win32".  (Though they use an extra suffix, too -- but I'm not sure end-users care)

release/platform/platform.go
```diff
@@ -0,0 +1,135 @@
...
+		Arch: "x86_64",
+	},
+	"ubuntu1604-arm": {
+		Name: "ubuntu1604",
+		Arch: "arm",
```
xdg: Server uses 'arm64'.

release/release.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/mongodb/mongo-tools/release/platform"
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fail("please provide exactly one subcommand name")
+	}
+	cmd := os.Args[1]
+
+	switch cmd {
```
xdg: I suggest adding a 'default' case that fails

release/release.go
```diff
@@ -0,0 +1,199 @@
...
+
+	err = tw.WriteHeader(header)
+	check(err, "write header to archive")
+
+	_, err = io.Copy(tw, file)
+	check(err, "write file to archive")
+}
+
+func buildTarball() {
+	win, err := platform.IsWindows()
```
xdg: minor: I probably would have pulled this logic up into command dispatch for the case in line 24 so it's obvious there that only one  thing being built.  (Or had a single `buildArchive` function that just dispatches to `buildTarball` or `buildZip`.)

release/release.go
```diff
@@ -0,0 +1,199 @@
...
+	check(err, "open file")
+	defer file.Close()
+
+	stat, err := file.Stat()
+	check(err, "stat file")
+
+	header := &tar.Header{
+		Name: dst,
+		Size: stat.Size(),
+		Mode: int64(stat.Mode()),
```
xdg: Do we want to force this to something like 0755 rather than rely on what's on the filesystem?

release/release.go
```diff
@@ -0,0 +1,199 @@
...
+	for _, name := range staticFiles {
+		log("adding %s to tarball", name)
+		src := name
+		dst := filepath.Join(releaseName, name)
+		addToTarball(tw, dst, src)
+	}
+
+	for _, binName := range binaries {
+		log("adding %s binary to tarball", binName)
+		src := filepath.Join(".", "bin", binName)
```
xdg: minor: here you use "." and above in line 137, you don't.   Also, I might have abstracted "write a list of files with an optional prefix" as the two for loops are otherwise identical.  E.g.:
```
addFilesToTarball("", staticFiles)
addFilesToTarball("bin",binaries)
```
But I'm not sure it's worth rewriting.

common.yml
```diff
@@ -701,6 +701,43 @@ functions:
...
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file: src/github.com/mongodb/mongo-tools/upload.tgz
+        remote_file: mongo-tools/pkgs/${build_id}/all-release-artifacts.tgz
+        content_type: application/x-gzip
+        bucket: mciuploads
+        permissions: public-read
+        display_name: All Release Artifacts (.tgz)
+    - command: s3.put
```
xdg: Suggest moving this up above line 713 so it more directly follows the build, then have the "bundle everything, upload everything" commands follow, possibly with a comment (e.g. "for the notary task").

release/release.go
rychipman: removed the `.` for consistency; I thought about factoring it further, but decided to keep the duplication since the loops were so small

common.yml
rychipman: Done!

release/release.go
rychipman: done!

release/release.go
rychipman: done!

release/release.go
rychipman: done!

release/platform/platform.go
rychipman: done!

release/release.go
```diff
@@ -0,0 +1,199 @@
...
+	"mongotop",
+}
+
+var staticFiles = []string{
+	"LICENSE.md",
+	"README.md",
+	"THIRD-PARTY-NOTICES",
+}
+
+func log(format string, args ...interface{}) {
```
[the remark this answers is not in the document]
rychipman: good suggestion! done

release/release.go
rychipman: removed the `.` for consistency. thought about that rewrite, but decided to leave the duplication instead of abstracting since the loops were so small

release/release.go
rychipman: done

release/platform/platform.go
rychipman: updated

release/platform/platform.go
rychipman: updated

release/platform/platform.go
rychipman: updated; decided to leave out the extra suffix

rychipman:
new evg patch: https://evergreen.mongodb.com/version/5e31ed4c57e85a3f3c004486

release/platform/platform.go
pmeredit: Yes. The general approach of OS X is supposed to be always being up-to-date, even though they are pretty bad about breaking things, lately.

release/platform/platform.go
pmeredit: I don't think many things use 32 bit arm anymore. Definitely nothing we want to run the tools on. It would likely be some really cheap embedded microcontroller (though a lot of those are actually 8 bit, still).

PR #185, opened 2020-01-28
TOOLS-1786: mongodump does not create metadata.json file for views dumped as collections

varsha804:


varsha804:
> Nice!
> 
> Could you add a Go test for --viewsAsCollection? There is already a jstest for --viewsAsCollection in test/qa-tests/jstests/dump/dump_views.js. But while we have the opportunity I think it would be good to add one to the Go suite.

Sure, will do! Why do we need both, though?

mattChiaravalloti:
Removing myself as a reviewer since I'm not familiar enough with tools _yet_ to make meaningful reviews.

edobranov:
^ Might have to do the same here 😶 The fix looks good to me though :)

rychipman:
> Sure, will do! Why do we need both, though?
At the very least, we will want a test that demonstrates that the bug has been fixed (the old js tests don't do that, otherwise they would have been failing). We _could_ do this by modifying the old js tests, but in general we want to be working towards having all our tests in go, so adding a new go regression test would be preferable!

(@tfogo feel free to add on if I missed anything or am off base here, but that was my interpretation)

tfogo:
The other option here is rewriting the metadata.json for views so that they actually represent collections in the dump.

mongodump/mongodump_test.go
```diff
@@ -1024,3 +1047,58 @@ func TestMongoDumpOrderedQuery(t *testing.T) {
...
+	log.SetWriter(ioutil.Discard)
+
+	Convey("With a MongoDump instance", t, func() {
+		err := setUpMongoDumpTestData()
+		So(err, ShouldBeNil)
+
+		err = setUpDBView()
+		So(err, ShouldBeNil)
+
+		Convey("testing that the dumped directory contains information about indexes", func() {
```
tfogo: Probably want to change the description here to say "metadata files" instead

mongodump/mongodump_test.go
```diff
@@ -1024,3 +1047,110 @@ func TestMongoDumpOrderedQuery(t *testing.T) {
...
+		})
+
+		Reset(func() {
+			So(tearDownMongoDumpTestData(), ShouldBeNil)
+		})
+
+	})
+}
+
+func TestMongoDumpViews(t *testing.T) {
```
tfogo: What's this test meant to be for?

varsha804: Ryan pointed out that if we dump a view without using the flag `--viewAsCollections`, we will attempt to run `listIndexes`, so I wrote this test to make sure views are being dumped correctly in the case that they are not read-only.

Do you think it's unnecessary/redundant?

varsha804:
> Currently when we dump a db without `--viewsAsCollections`, it dumps the metadata files for views, even though the views themselves aren't dumped. I don't think the views should even be making it into the intent manager. So we should edit `CreateIntentsForDatabase()` to skip views if `--viewsAsCollections` isn't set.

Not sure if I've misunderstood your comment, but do you mean a mongodump command without the `--viewAsCollections` flag shouldn't be dumped at all? If the views don't make it to the intent manager, then the .metadata.json file wouldn't be created, right?

mongodump/mongodump_test.go
tfogo: I see, I think there's a small bit of confusion here about the terminology of collections vs views. This test is dumping collections, not views. If `--viewsAsCollections` isn't set, then only collections are dumped. So the name of the test confused me a bit.

Just to clarify the terminology, there are two types of collection-like objects: regular collections which store data on disk, and views. Views are always read-only. They are a stored aggregation pipeline and they don't save any data to disk. 

(There are also _materialized views_, which do save the output of an aggregation to disk. Those can be read/write. But materialized views are just regular collections. The name describes the technique for populating the collection, not anything about the collection itself.)

I don't think this test is necessary since it's already covered by `TestMongoDumpMetaData()`.

tfogo:
If `--viewAsCollections` is not set, then the regular collections should be dumped, but no views should be dumped. So if you had collection `foo` and view `bar`, running mongodump without `--viewAsCollections` should dump foo.bson and foo.metadata.json only. 

Yes, if the views aren't added to the intent manager, the metadata files won't be created for them.

PR #186, opened 2020-01-30
TOOLS-2449: Create sign task

rychipman:
This PR creates the sign task, though it doesn't yet do any signing (since the artifacts we will sign, MSIs and linux packages, have not been created yet).

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -13,6 +13,8 @@ mongo_tools_variables:
   mongo_tools_task_lists:
     mac_task_list: &macos_tasks
       - name: dist
+      - name: sign
+        run_on: amazon1-2018-test
       - name: integration-4.0
       - name: integration-4.0-auth
       - name: integration-4.2
@@ -44,6 +46,8 @@ mongo_tools_variables:
       - name: replay-dist
     rhel_x86_64_task_list: &rhel_x86_64_tasks
       - name: dist
+      - name: sign
+        run_on: amazon1-2018-test
       - name: integration-4.0
       - name: integration-4.0-auth
       - name: integration-4.2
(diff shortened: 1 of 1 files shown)


rychipman:
evergreen patch: https://evergreen.mongodb.com/version/5e332aa4850e612f881065b4

common.yml
```diff
@@ -13,6 +13,8 @@ mongo_tools_variables:
   mongo_tools_task_lists:
     mac_task_list: &macos_tasks
       - name: dist
+      - name: sign
+        run_on: amazon1-2018-test
```
rychipman: Not every evg distro has the notary client on it, which is why we can't just bundle signing into the dist task. This is one that happens to have it available, so we will run every sign task on it, regardless of what the default distro for the buildvariant is.

common.yml
```diff
@@ -1020,6 +1057,13 @@ tasks:
         tool: mongotop
     - func: "create release artifacts"
 
+- name: sign
+  depends_on:
+    - name: dist
+  commands:
+    - func: "fetch dist release artifacts"
```
rychipman: once https://jira.mongodb.org/browse/TOOLS-2438 is done, https://jira.mongodb.org/browse/TOOLS-2440 will add another step here that actually signs the artifacts that need to be signed.

rychipman:
I added `/task/` to the path to make it a little more clear that the remote path is named after the task!
New evg patch: https://evergreen.mongodb.com/version/5e3467b1a4cf4702e889abc3

common.yml
pmeredit: I say, let's add this comment to the file. It'll help people looking at it in the future :)

PR #190, opened 2020-02-05
TOOLS-2462: Password prompt does not work on windows

varsha804:


varsha804:
> Is it possible for us to add a regression test for this bug? For example, if the test just called `GetPass()`, would that fail on windows prior to this fix?

I don't think this would work because `GetPass()` uses `terminal` functions, and passes the file descriptor to `terminal.MakeRaw()`. If we run this from a testing environment, the file descriptor would be different, right?

I've created a test which just runs `GetPass()`, but it fails (with the error `The handle is invalid`) despite which file descriptor is used in `GetPass()`.

rychipman:
Makes sense! In that case, manual testing is just fine!

PR #191, opened 2020-02-05
TOOLS-1493: Update mongo-tools-common and go-flags dependencies

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-1493

This PR is a continuation of [another PR](https://github.com/mongodb/mongo-tools-common/pull/24) on mongo-tools-common.

The only lines manually changed are in `Gopkg.toml` and `mongoexport/options.go`. The rest is a result of `dep ensure -update github.com/mongodb/mongo-tools-common`and `dep ensure -update github.com/jessevdk/go-flags`


mongoexport/options.go
```diff
@@ -33,7 +33,7 @@ type OutputFormatOptions struct {
 	Type string `long:"type" value-name:"<type>" default:"json" default-mask:"-" description:"the output format, either json or csv"`
 
 	// Deprecated: allow legacy --csv option in place of --type=csv
-	CSVOutputType bool `long:"csv" default:"false" hidden:"true"`
```
tfogo: What's up with the `default:"false"` tags being removed?

edobranov: This was a design change in the go-flags library that we had to account for. It throws an error if you attempt to use a default tag on a boolean flag.

This is where the change was introduced: https://github.com/jessevdk/go-flags/commit/7f2ab82552ae1b12e070838918ac056507c0f114#diff-103d48dbdf9e16c6eb09b90668a9e984R279

PR #192, opened 2020-02-07
Update rhel72-zseries "name" field.

andf-mongodb:


rychipman:
Thank you @andf-mongodb!

PR #195, opened 2020-02-11
TOOLS-2275: autoIndexId:false is not supported in 4.0

varsha804:


mongorestore/metadata.go
```diff
@@ -288,6 +288,12 @@ func (restore *MongoRestore) createCollectionWithApplyOps(session *mongo.Client,
 }
 
 func createCollectionCommand(intent *intents.Intent, options bson.D) bson.D {
+	for i, elem := range options {
```
pmeredit: This code doesn't seem to be checking the version of the server at all. Did we decide to disallow this on all versions even though it's not supported on 4.2 (or is it 4.0?)

mongorestore/metadata.go
```diff
@@ -288,6 +288,12 @@ func (restore *MongoRestore) createCollectionWithApplyOps(session *mongo.Client,
...
+		if elem.Key == "autoIndexId" {
+			options[i].Value = true
+			log.Logvf(log.Always, "autoIndexId: false is not allowed in server version 4.2.0. Changing to autoIndexId: true.")
```
huan-Mongo: Is this log accurate? The value might be true already.

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,37 @@ func TestFixHashedIndexes(t *testing.T) {
...
+	session, err := testutil.GetBareSession()
+	if err != nil {
+		t.Fatalf("No server available")
+	}
+
+	type indexRes struct {
+		Key bson.D
+	}
+
+	Convey("Test MongoRestore with hashed indexes and --fixHashedIndexes", t, func() {
```
huan-Mongo: Was this test name copied from above?

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,37 @@ func TestFixHashedIndexes(t *testing.T) {
...
+		db := session.Database("testdata")
+
+		defer func() {
+			db.Collection("autoIndexId").Drop(nil)
+		}()
+
+		Convey("{autoIndexId: false} should be ignored and mongorestore should be successful", func() {
+			restore.TargetDirectory = "testdata/autoindexid/test/test_auto_idx.bson"
+			result := restore.Restore()
+			So(result.Err, ShouldBeNil)
```
huan-Mongo: Maybe a check on the index on _id was created?

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,37 @@ func TestFixHashedIndexes(t *testing.T) {
 		})
 	})
 }
+
+func TestDeprecatedAutoIndexId(t *testing.T) {
```
huan-Mongo: We need another test showing this fix won't affect versions below 4.0

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,91 @@ func TestFixHashedIndexes(t *testing.T) {
...
+			panic("no matching collection found")
+		}
+		var collInfo struct {
+			Options bson.M
+		}
+		if err = cursor.Decode(&collInfo); err != nil {
+			panic(err)
+		}
+
+		Convey("{autoIndexId: false} should be flipped to true iff server version >= 4.0", func() {
```
huan-Mongo:
```suggestion
		Convey("{autoIndexId: false} should be flipped to true if server version >= 4.0", func() {
```

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,91 @@ func TestFixHashedIndexes(t *testing.T) {
...
+		}
+		var collInfo struct {
+			Options bson.M
+		}
+		if err = cursor.Decode(&collInfo); err != nil {
+			panic(err)
+		}
+
+		Convey("{autoIndexId: false} should never be flipped to true", func() {
+			if restore.serverVersion.LTE(db.Version{4, 2, 0}) {
```
huan-Mongo: do we need the version check here?

varsha804: I figured this would be a sanity check to make sure that the behavior we added for 4.0 and 4.2 is not applied in the case that the database is `local`.

huan-Mongo: but even so, this checks all the version <=4.2 instead of 4.0 and 4.2?

mongorestore/metadata.go
```diff
@@ -229,6 +229,17 @@ func (restore *MongoRestore) LegacyInsertIndex(intent *intents.Intent, index Ind
 	return nil
 }
 
+// UpdateAutoIndexId updates {autoIndexId: false} to {autoIndexId: true} if the server version is
+// >= 4.0 and the database is not `local`.
+func (restore *MongoRestore) UpdateAutoIndexId(options bson.D) {
```
rychipman: Based on the docstring for this function, it looks like we should actually have the version check inside here instead of at the call sites.

Alternatively, if you'd prefer, you could keep the version check outside of this function; in that case, however, I don't think there would be any need for this to be a method on `MongoRestore` (it could just be a plain helper function)

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,91 @@ func TestFixHashedIndexes(t *testing.T) {
 		})
 	})
 }
+
+func TestRandom(t *testing.T) {
```
rychipman: This looks like this may have just been a placeholder name

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,91 @@ func TestFixHashedIndexes(t *testing.T) {
...
+	testtype.SkipUnlessTestType(t, testtype.IntegrationTestType)
+	ctx := context.TODO()
```
rychipman: I think it's okay to just use `context.Background()` here!

mongorestore/local/test_auto_idx.metadata.json
```diff
@@ -0,0 +1 @@
+{"options":{"autoIndexId":false},"indexes":[],"uuid":"0715c8e614a747d8baf5853ab381a303"}
```
rychipman: I think you might want this and its corresponding bson file to be in `mongorestore/testdata/local/` instead of `mongorestore/local/`

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,91 @@ func TestFixHashedIndexes(t *testing.T) {
...
+
+	session, err := testutil.GetBareSession()
+	if err != nil {
+		t.Fatalf("No server available")
+	}
+
+	Convey("Test MongoRestore with {autoIndexId: false} in a non-local database's collection", t, func() {
```
rychipman: Nice tests!! Can we also add a case for a non-local db with the `PreserveUUID` set, so we get coverage for `createCollectionWithApplyOps` too?

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,91 @@ func TestFixHashedIndexes(t *testing.T) {
...
+		dbName := session.Database("testdata")
+
+		defer dbName.Collection("test_auto_idx").Drop(ctx) // drop the collection to clean up resources
```
rychipman: I don't remember how exactly it all works under the hood, but I think goconvey does some funny things with the closures you pass to it. I'm not sure whether a defer is safe or not in a convey block.

IIRC, there's some sort of special way to declare cleanup code that guarantees it will play nicely with the rest of convey.

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,91 @@ func TestFixHashedIndexes(t *testing.T) {
...
+			panic("no matching collection found")
+		}
+		var collInfo struct {
+			Options bson.M
+		}
+		if err = cursor.Decode(&collInfo); err != nil {
+			panic(err)
+		}
+
+		Convey("{autoIndexId: false} should be flipped to true if server version >= 4.0", func() {
```
rychipman: did you manually test this against a pre-4.0 mongodb instance? At the moment, I don't think we actually run our go integration tests against anything earlier than MongoDB 4.0

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,91 @@ func TestFixHashedIndexes(t *testing.T) {
...
+		So(err, ShouldBeNil)
+
+		restore.TargetDirectory = "testdata/test_auto_idx.bson"
+		result := restore.Restore()
+		So(result.Err, ShouldBeNil)
+
+		// Find the collection
+		filter := bson.D{{"name", "test_auto_idx"}}
+		cursor, err := session.Database("testdata").ListCollections(ctx, filter)
+		if err != nil {
```
rychipman: here and elsewhere in this file, I think we want to use `So(err, ShouldBeNil)` instead of checking the err and panicking when non-nil

mongorestore/mongorestore_test.go
varsha804: Oops, you're right!

mongorestore/metadata.go
varsha804: True! Doing so also eliminates some duplicate code.

mongorestore/mongorestore_test.go
varsha804: Discussed offline: defers seem to work inside of conveys used by other mongorestore tests, so it should be fine.

mongorestore/mongorestore_test.go
varsha804: Discussed offline: I've modified the evergreen file (`common.yml`) so that a 3.6 integration test is run.

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,145 @@ func TestFixHashedIndexes(t *testing.T) {
...
+		So(documentExists, ShouldBeTrue)
+
+		var collInfo struct {
+			Options bson.M
+		}
+		err = cursor.Decode(&collInfo)
+		So(err, ShouldBeNil)
+
+		Convey("{autoIndexId: false} should not be flipped to true", func() {
+			if restore.serverVersion.GTE(db.Version{4, 0, 0}) {
```
tfogo: If we're only testing this for >= 4.0.0, I'd just skip the entire test at the start if it's >= 4.0.0.

mongorestore/mongorestore_test.go
```diff
@@ -583,3 +583,145 @@ func TestFixHashedIndexes(t *testing.T) {
...
+
+			Convey("{autoIndexId: false} should be flipped to true if server version >= 4.0", func() {
+				if restore.serverVersion.GTE(db.Version{4, 0, 0}) {
+					So(collInfo.Options["autoIndexId"], ShouldBeTrue)
+				} else {
+					So(collInfo.Options["autoIndexId"], ShouldBeFalse)
+				}
+			})
+		})
+		Convey("Set --preserveUUID", func() {
```
tfogo: What difference is there if --preserveUUID is set?

tfogo: Oh! Using ApplyOps! I see now.

mongorestore/mongorestore_test.go
rychipman: I actually don't know if we need to check the version here at all. If we're restoring to the `local` db, shouldn't `autoIndexId: false` always be preserved by restore, regardless of version?

varsha804: Ah, that's right Ryan. I'll remove the version check!

varsha804:
Just realized there should be a version check for `--preserveUUID` since it's new as of 4.0.

PR #197, opened 2020-02-12
TOOLS-2418: Remove mongoreplay

rychipman:
This PR removes mongoreplay from the mongo-tools repository. This includes all the code for mongoreplay, some code in `legacy/` that was used only by mongoreplay, and a number of tasks and functions in the evergreen project file that were used only for mongoreplay.


huan-Mongo:
The failed evergreen tasks seem nothing related this change.

PR #198, opened 2020-02-12
TOOLS-2473: Consolidate community, enterprise, and ssl buildvariants

rychipman:
There will be a single release of the tools for each platform we support (not separate enterprise and community releases, as there were when we released with the server). As such, we no longer need separate enterprise and community buildvariants for each supported platform.

Additionally, with the removal of mongoreplay from the tools, we do not need separate variants testing with and without SSL.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -27,12 +27,6 @@ mongo_tools_variables:
       - name: qa-tests-unstable
       - name: qa-dump-restore-with-gzip-current
       - name: qa-dump-restore-with-archiving-current
-      - name: unit
-    macos_ssl_task_list: &macos_ssl_tasks
-      - name: dist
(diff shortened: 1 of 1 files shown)


tfogo:
I was looking at the evg patches for this. Looks like the MacOS 10.14 integration tests started failing when you pulled out SSL.

rychipman:
I've added the missing SSL expansions that were causing the macOS failures. Still looking into some strange jstest failures that don't seem like they should be related. Might just be flakiness, so I'm restarting some tasks and waiting to see what happens

rychipman:
After restarting the failing tasks on windows, I don't see evidence of anything other than some jstests flakiness; I think this is ready for a look ffrom everyone now!

PR #200, opened 2020-02-18
TOOLS-2455: mongorestore hangs on invalid archive

varsha804:


mongorestore/restore.go
```diff
@@ -147,6 +147,7 @@ func (restore *MongoRestore) RestoreIntents() Result {
 			totalResult.combineWith(result)
 			if totalErr == nil && totalResult.Err != nil {
 				totalErr = totalResult.Err
+				return totalResult.withErr(totalErr)
```
dragonlee8: Without this line, the function will still exit in the return statement below, right? What's the change here?

huan-Mongo: Without this line, this function will still exit from the return statement below, right? What's the change here?

mongorestore/restore.go
```diff
@@ -147,6 +147,7 @@ func (restore *MongoRestore) RestoreIntents() Result {
 			totalResult.combineWith(result)
 			if totalErr == nil && totalResult.Err != nil {
```
huan-Mongo: After the change, this piece of logic is kinda convoluted. From my understanding, it seems you gonna return from the loop if there is any error and return with success if the function hits the bottom? 
So it can be simplified to
```
		for i := 0; i < restore.OutputOptions.NumParallelCollections; i++ {
			if totalErr == nil && totalResult.Err != nil {
			result := <-resultChan
                         totalResult.combineWith(result)
			if totalResult.Err != nil {
                               return totalResult
                         }
             }
             return totalResult
```
Please let me know if I understand correctly

varsha804: Yes, I believe we want to terminate the process after the first error. I'll update the file to reflect this simplification!

huan-Mongo:
LGTM!

PR #202, opened 2020-02-18
TOOLS-2442: Naming conventions and refactors

rychipman:
This PR contains some platform renaming that is better to do in a separate PR from the rest of 2442 for ease of testing. Also wrapped in some refactors/rationalizations of the helper functions and types we're using in the release/ dir.

Only requiring @pmeredit from the get-go, since it's mostly just internal refactors of code he and I have added in the release/ dir.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -188,10 +188,6 @@ mongo_tools_variables:
     ssl: &mongod_ssl_startup_args
       mongod_args: "--sslMode requireSSL --sslCAFile testdata/certs/ca.pem --sslPEMKeyFile testdata/certs/server.pem"
       mongod_port: 33333
-    # Set storage engine as mmapv1 for 32 bit variants because WiredTiger requires 64 bit support.
-    win32: &mongod_win32_startup_args
-      mongod_args: "--storageEngine=mmapv1"
-      mongod_port: 33333
 
   mongo_arguments:
     default: &mongo_default_startup_args
@@ -574,10 +570,7 @@ functions:
       params:
         working_dir: src/github.com/mongodb/mongo-tools
         script: |
-          . ./set_goenv.sh
-          GOROOT="" set_goenv || exit
-          export EVG_PLATFORM=${_platform}
-          set -o errexit
+          ${_set_shell_env}
           go run release/release.go build-archive
           go run release/release.go build-packages
 
@@ -587,7 +580,7 @@ functions:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
         local_files_include_filter:
-          - src/github.com/mongodb/mongo-tools/mongodb-*
+          - src/github.com/mongodb/mongo-tools/release.*
         remote_file: mongo-tools/pkgs/${build_id}/
         content_type: application/octet-stream
         bucket: mciuploads
@@ -601,7 +594,7 @@ functions:
         target: src/github.com/mongodb/mongo-tools/upload.tgz
         source_dir: src/github.com/mongodb/mongo-tools
         include:
-          - "./mongodb-cli-tools-*"
+          - "./release.*"
     - command: s3.put
       params:
         aws_key: ${aws_key}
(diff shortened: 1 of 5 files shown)


common.yml
```diff
@@ -587,7 +584,7 @@ functions:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
         local_files_include_filter:
-          - src/github.com/mongodb/mongo-tools/mongodb-cli-tools-*
+          - src/github.com/mongodb/mongo-tools/release.*
```
rychipman: Since the same artifact may get uploaded as multiple different names (for example, as a "stable" and a "latest stable" release), it is easier and less confusing to just use release.tgz (and release.msi, etc) here.

common.yml
```diff
@@ -1416,13 +1430,12 @@ buildvariants:
 #     Amazon x86_64 Buildvariants     #
 #######################################
 
-- name: amazonlinux64
+- name: amazon
```
rychipman: did some work to match the server's platform naming here. also removed the largely-redundant `_platform` expansion

release/env/env.go
```diff
@@ -0,0 +1,34 @@
+package env
```
rychipman: consolidated a bunch of funcs that access env variables into one package

release/platform/platform.go
```diff
@@ -12,180 +22,185 @@ import (
 type Platform struct {
 	Name string
 	Arch string
+	OS   string
+	Pkg  string
+}
```
rychipman: as much as possible, trying to move data about platforms into the structs themselves instead of having it in helper funcs that switch on the name/arch fields

release/release.go
```diff
@@ -75,13 +78,18 @@ func run(name string, args ...string) (string, error) {
 }
 
 func getVersion() string {
-	desc, err := run("git", "describe")
+	desc, err := run("git", "describe", "--dirty")
```
rychipman: differentiate between a build run with a clean git status and one run with uncommitted changes

release/release.go
```diff
@@ -159,20 +167,20 @@ func buildDeb() {
...
+	check(os.MkdirAll(binDir, os.ModePerm), "mkdirAll "+binDir)
 	docDir := filepath.Join(releaseName, "usr", "share", "doc", mdt)
-	check(os.MkdirAll(docDir, os.ModePerm), "mkdirAll " + docDir)
+	check(os.MkdirAll(docDir, os.ModePerm), "mkdirAll "+docDir)
 
 	md5sums := make(map[string]string)
 	// We use the order just to make sure the md5sums are always in the same order.
 	// This probably doesn't matter, but it looks nicer for anyone inspecting the md5sums file.
-	md5sumsOrder := make([]string, 0, len(binaries) + len(staticFiles))
+	md5sumsOrder := make([]string, 0, len(binaries)+len(staticFiles))
```
rychipman: `gofmt` did this

set_goenv.sh
```diff
@@ -83,7 +83,7 @@ set_goenv() {
 }
 
 print_ldflags() {
-    VersionStr="$(git describe)"
+    VersionStr="$(go run release/release.go get-version)"
```
rychipman: make sure `--version` is getting its version info from the same place as our packaging code

common.yml
```diff
@@ -1472,7 +1482,7 @@ buildvariants:
 #           macOS Buildvariant        #
 #######################################
 
-- name: macOS-1014
+- name: macos
```
pmeredit: much better, macos should always be latest.

release/platform/platform.go
```diff
@@ -12,180 +22,185 @@ import (
...
-	}
-
-	switch p.Name {
-	case "win32", "macos":
-		return false, nil
-	default:
-		return true, nil
+func (p Platform) DebianArch() string {
+	if p.Pkg != PkgDeb {
+		panic("called DebianArch on non-debian platform")
```
pmeredit: Nice.

release/release.go
pmeredit: We should probably set gofmt to run automatically at some point. But I've always felt that way :D

release/platform/platform.go
pmeredit: I'm wondering if we should stick DebArch in here too, but I think that's probably overkill, since not every platform needs that.

PR #203, opened 2020-02-21
TOOLS-2489: Fix and separate static analysis tasks 

rychipman:
this PR fixes a failing format-go task and moves format-go, lint-go, vet, and lint-js into their own buildvariant for easier viewing from the github PR patchbuild widget

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -40,9 +40,6 @@ mongo_tools_variables:
       # Disabled due to TOOLS-2119
       # - name: kerberos
       - name: legacy30
-      - name: lint-go
-      - name: format-go
-      - name: lint-js
       - name: qa-tests-3.2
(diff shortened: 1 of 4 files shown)


common.yml
```diff
@@ -1412,6 +1404,23 @@ tasks:
 
 buildvariants:
 
+#######################################
+#     Static Analysis Buildvariant    #
+#######################################
+
+- name: static
+  display_name: '! Static Analysis'
```
mattChiaravalloti: why is this "! Static Analysis" instead of just "Static Analysis"?

rychipman: I actually borrowed this trick from the server evergreen project: they just use it to keep certain variants at the top of the list in patches/waterfall

PR #207, opened 2020-02-24
TOOLS-2275: Revert common.yml reformatting

rychipman:
There was some extra reformatting bundled in with #195 that is causing major rebase conflicts with in-progress work. Putting up a quick PR to revert the reformatting, since that will be significantly less effort than working through the rebase!


varsha804:
LGTM!

PR #208, opened 2020-02-24
TOOLS-2460 Attempt to parse positional arguments as URIs

tfogo:
This PR allows the tools to take a connection string as an extra positional argument. The parsing and some error handling takes place in mongo-tools-common. Tool-specific error handling and tests are in the mongo-tools repo.

`options.ToolOptions.New()` now takes a boolean argument to specify if we should attempt to parse positional args as connection strings. This allows mongomirror to opt out of this change. This is a small breaking change and will require bumping the major version of mongo-tools-common.

This PR also unifies the approach to parsing args in the tools. Now each tool has an Options type and a ParseOptions function.

### Changes from the design document

There are changes to the error messages described in sections 3(d)-(g). The following message is now used for all of those eventualities:

> error parsing positional arguments: provide <tool-specific-args> and only one MongoDB connection string. Connection strings must begin with mongodb:// or mongodb+srv:// schemes

This error message conveys the necessary information for solving each of those error cases. Using one error message reduces the complexity of the error checking required.

An extra error has been added if multiple URIs are specified in the positional arguments:

> too many URIs found in positional arguments: only one URI can be set as a positional argument

The design doc focused on the pre-existing positional argument of mongorestore, it fails to mention that mongoimport, mongofiles, mongostat, and mongotop also have pre-existing positional arguments.

---

There are a few evergreen failures that I need to fix up, but they're small so I wanted to get this into review.

This PR includes the TOOLS-2459 work in mongo-tools-common and GODRIVER-1490.


mongoexport/options_test.go
```diff
@@ -106,3 +109,60 @@ func TestParseOptions(t *testing.T) {
 		}
 	})
 }
+
```
pmeredit: I'm not sure we need to repeat the same tests in every one of the tools. Is there some good way we could centralize these tests into m-c-t?

tfogo: That's a good question. Each tool has a `ParseOptions()` function. These functions do tool-specific parsing and error handling. If we wanted to properly test all the errors in m-t-c, we would have to push some of that tool-specific logic into m-t-c. I actually started implementing this PR in that way, but eventually decided on the current approach. My reasoning was:

1. We should limit the amount of tool-specific code in m-t-c. It's meant to be a library of common functionality. If m-t-c performs different logic for each tool, it will become hard to reason about the functionality of tools when just looking at the tools repo.
2. It's more prone to regression if the code path is still `main()`->`ParseOptions()`->`options.ParseArgs()`. If we're only testing the m-t-c portion (`options.ParseArgs`), it won't catch regressions that could arise in `ParseOptions()`.

I agree it's not ideal, but I don't think there's a quick and easy way to centralize these tests.

tfogo:
Here's an easier way to see the changes to m-t-c: https://github.com/mongodb/mongo-tools-common/pull/28

xdg:
FYI: Taking myself off review.

mongoimport/options_test.go
```diff
@@ -100,3 +101,135 @@ func TestLegacyOptionParsing(t *testing.T) {
...
+			{
+				InputArgs: []string{"foo", "bar"},
+				ExpectErr: "error parsing positional arguments: " +
+					"provide only one polling interval in seconds and only one MongoDB connection string. " +
+					"Connection strings must begin with mongodb:// or mongodb+srv:// schemes",
+			},
+			{
+				InputArgs: []string{"foo", "bar", "mongodb://foo"},
+				ExpectErr: "error parsing positional arguments: " +
+					"provide only one polling interval in seconds and only one MongoDB connection string. " +
```
rychipman: Looks like this might be a copy-and-pase err from another tool's tests?

mongostat/options_test.go
```diff
@@ -0,0 +1,123 @@
...
+					"Connection strings must begin with mongodb:// or mongodb+srv:// schemes",
+			},
+			{
+				InputArgs: []string{"2", "3", "mongodb://foo"},
+				ExpectErr: "error parsing positional arguments: " +
+					"provide only one polling interval in seconds and only one MongoDB connection string. " +
+					"Connection strings must begin with mongodb:// or mongodb+srv:// schemes",
+			},
+			{
+				InputArgs: []string{"mongodb://foo", "--uri=mongodb://bar"},
```
rychipman: Feel free to ignore if this is already tested elsewhere, but perhaps we may also want to test flags whose args are separate (as opposed to part of the same arg and separated with `=`)?

mongodump/main/mongodump.go
```diff
@@ -31,23 +30,9 @@ var (
...
-	opts := options.New("mongodump", VersionStr, GitCommit, mongodump.Usage, options.EnabledOptions{Auth: true, Connection: true, Namespace: true, URI: true})
-
-	inputOpts := &mongodump.InputOptions{}
-	opts.AddOptions(inputOpts)
-	outputOpts := &mongodump.OutputOptions{}
-	opts.AddOptions(outputOpts)
-	opts.URI.AddKnownURIParameters(options.KnownURIOptionsReadPreference)
-
-	args, err := opts.ParseArgs(os.Args[1:])
+	opts, err := mongodump.ParseOptions(os.Args[1:], VersionStr, GitCommit)
```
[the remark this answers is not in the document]
tfogo: Just for the sake of consistency. I think the only other change was the new error message.

mongoimport/options.go
```diff
@@ -139,6 +148,24 @@ func ParseOptions(rawArgs []string, versionStr, gitCommit string) (Options, erro
 	}
 	opts.WriteConcern = wc
 
+	// ensure no more than one positional argument is supplied
+	if len(args) > 1 {
```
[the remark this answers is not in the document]
tfogo: Yes you're correct, I should remove this.

mongoimport/options_test.go
tfogo: Good catch!

mongostat/options_test.go
tfogo: I think it's appropriately tested in mongo-tools-common

tfogo:
I'll resolve the Gopkg.lock issues before merging.

PR #209, opened 2020-02-26
TOOLS-2440: Sign MSI installer

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2440

I copied over the [sqlproxy `"sign msi installer"` function](https://github.com/10gen/sqlproxy/blob/1add0bfe7e43375274a936ac427f3ffb3d083d14/.evg.yml#L411-L430) and just changed the working dir, key name and auth token. There might be something else, but I believe this was it.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -617,6 +617,27 @@ functions:
         extract_to: src/github.com/mongodb/mongo-tools/
         bucket: mciuploads
 
+  "sign msi installer":
+    - command: shell.exec
+      type: system
+      params:
+        silent: true
+        script: |
+          echo "${signing_auth_token}" > ${PROJECT_DIR}/signing_auth_token
+    - command: shell.exec
+      # If the notary service is failing, it should be a system failure to alert the evg team.
+      type: system
+      params:
+        working_dir: src/github.com/mongodb/mongo-tools/
+        script: |
+          /usr/local/bin/notary-client.py \
+              --key-name "server-Tools" \
+              --auth-token-file ${PROJECT_DIR}/signing_auth_token \
+              --comment "Evergreen Automatic Signing (mongo-tools) - ${version_id} - ${build_variant}" \
+              --notary-url http://notary-service.build.10gen.cc:5000 \
+              --skip-missing \
+              release.msi
+
   "upload signed release artifacts":
     - command: s3.put
       params:
@@ -869,6 +890,7 @@ tasks:
     - name: dist
   commands:
     - func: "fetch dist release artifacts"
+    - func: "sign msi installer"
     - func: "upload signed release artifacts"
 
 - name: integration-3.6
--- a/common.yml
+++ b/common.yml
@@ -632,7 +632,7 @@ functions:
         working_dir: src/github.com/mongodb/mongo-tools/
         script: |
           /usr/local/bin/notary-client.py \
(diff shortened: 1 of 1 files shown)


pmeredit:
release.msi also doesn't exist on all the failing platforms. We probably need to make this conditional on Windows (I'm not sure if the client fails when the file doesn't exist, and the log isn't helpful).

edobranov:
> release.msi also doesn't exist on all the failing platforms. We probably need to make this conditional on Windows (I'm not sure if the client fails when the file doesn't exist, and the log isn't helpful).

Shouldn't the `--skip-missing` flag for the client take care of that? The docs mention that turning that on will `Skip missing files, instead of exiting with an error`, so I'd imagine we can get away with that.

edobranov:
Looks like the issue was mentioning variables that don't exist on the Evergreen project. Should be working now that I removed them.

I'm assuming that having the `signing_auth_token` in the current dir alongside `release.msi` doesn't have any security concerns since only `release.*` artifacts are uploaded, if you guys want to double-check.

rychipman:
Code changes look good, just two other requests:
- Can you add the windows variant to the latest patchbuild to verify that the sign task works there?
- Can you do a quick manual test to ensure that the signed MSI behaves as expected when installed?

pmeredit:
I added WIndows, we should see soon :)

edobranov:
Thanks @pmeredit! It looks to be working. I'll spawn a host off of that variant and double-check the installation.

pmeredit:
Signing doesn't seem to work (unknown publisher on Windows) :( Perhaps the key is wrong?

edobranov:
Agh yea, that's what I've been trying to figure out for the last couple hours. [This Jira ticket](https://jira.mongodb.org/browse/BUILD-9871) mentions the [`server-Tools` key in this directory](https://www.mongodb.org/static/pgp/), so I don't know what else it could be...

May be a question for the build team? Idk

edobranov:
Okay, so it looks like we had everything right but the notary client has an implicit `--package-file-suffix` that appends `-signed` to the end of the file (`release.msi` is signed as `release-signed.msi` which doesn't get uploaded because of the `release.*` pattern). We can probably just do `--package-file-suffix ""` to overwrite the original.

Not sure how, but sqlproxy seems to get away without mentioning the flag. I'm gonna look into that a bit first to make sure we're not missing anything.

edobranov:
msi

Phew, fortunately that ended up being the fix. Turns out sqlproxy just references `release-signed.msi` directly, so we don't have to worry about that. Here's the output from `signtool` on a Windows host:

```
> signtool verify /pa release_unsigned.msi
File: release_unsigned.msi
Index  Algorithm  Timestamp
========================================
SignTool Error: No signature found.

Number of errors: 1

> signtool verify /pa release.msi
File: release.msi
Index  Algorithm  Timestamp
========================================
0      sha256     None

Successfully verified: release.msi
```

The `/pa` switch uses the "Default Authentication Verification Policy" rather than the "Windows Driver Verification Policy". Or basically what's needed to get signtool to work.

edobranov:
Yep, I think Evergreen gets a bit iffy if you try to unschedule tasks that are running 🤷‍♂

PR #210, opened 2020-02-26
TOOLS-2334: Skip system collections during oplog replay	

varsha804:
`TestSkipSystemCollections` in `mongorestore_test.go` compares all collections before and after replaying the oplog during `mongorestore`. `mongomirror_internal_test.go` currently supports that logic, and so I've borrowed some functions from that file in this testing code. I couldn't make calls directly to functions in the `mongomirror` package due to type differences; therefore, there is currently some redundancy between the `mongorestore` and `mongomirror` testing files.


huan-Mongo:
Did you miss the file testdata/oplog_partial_skips?

varsha804:
> Did you miss the file testdata/oplog_partial_skips?

Ah, yes, sorry about that. I always forget to force add testdata files (they sometimes don't show up when I run `git status`)!

mongorestore/mongorestore_test.go
```diff
@@ -723,3 +979,48 @@ func TestAutoIndexIdNonLocalDB(t *testing.T) {
...
+
+		result := restore.Restore()
+		So(result.Err, ShouldBeNil)
+		So(result.Failures, ShouldEqual, 0)
+
+		// Verify restoration
+		_, err = c1.CountDocuments(nil, bson.M{})
+		So(err, ShouldBeNil)
+
+		assertCollectionEqual(t, collBeforeRestore, session.Database("db3").Collection("c1"), session, ctx)
```
huan-Mongo: I don't think this is correct. collBeforeRestore and the `session.Database("db3").Collection("c1")` are essentially the same thing. They are both the handle for the same collection instead of a snapshot. I think what you should compare is the source collection (the collection from which the dump file was created ) and the destination collection. 
I guess you can bring up a source cluster and then create the dump file in the test. Then use the same file to restore to destin cluster. In this way, you can compare these two collections side by side.

varsha804: Oh right. I'll create two standalones then!

mongorestore/mongorestore_test.go
```diff
@@ -20,6 +25,7 @@ import (
 	"github.com/mongodb/mongo-tools-common/testutil"
 	. "github.com/smartystreets/goconvey/convey"
 	"go.mongodb.org/mongo-driver/bson"
+	driverOpt "go.mongodb.org/mongo-driver/mongo/options"
```
rychipman: for consistency with what we do elsewhere in mongo-tools and mongo-tools-common, let's import this as `mopt`

mongorestore/mongorestore_test.go
```diff
@@ -723,3 +979,48 @@ func TestAutoIndexIdNonLocalDB(t *testing.T) {
 		}
 	})
 }
+
+// TestSkipSystemCollections asserts that certain system collections like "config.systems.sessions" and the transaction
+// related tables aren't applied via applyops when replaying the oplog.
+func TestSkipSystemCollections(t *testing.T) {
```
rychipman: Per offline discussion, let's rework this test to actually query the oplog instead of just comparing collections.

mongorestore/mongorestore_test.go
tfogo: There are a few other places in mongo-tools that still use `driverOpts` as the name. @varsha804 , while you're changing this one, could you also find/replace the others so we can make it consistent?

mongorestore/mongorestore_test.go
```diff
@@ -273,6 +279,256 @@ func generateTestData() error {
...
+
+		indexes = append(indexes, index)
+	}
+	sort.Sort(ByIndexName(indexes))
+	return indexes, nil
+}
+
+// AssertBsonDEqualUnordered, asserts two bson.Ds are equal ignoring the
+// order of the top level fields. Useful for comparing indexes.
+func assertBsonDEqualUnordered(expected, actual *bson.D) {
```
tfogo: You might want to recurse on `assertBsonDEqualUnordered` if you really want to make this robust. The collations can be `bson.D`s so it's theoretically possible that could cause issues. Or you could make `bsonDToM` recurse. On the other hand, I think that's pretty unlikely.

Did you run into issues using `assertBsonDEqual()` with index documents? If `assertBsonDEqual()` works, maybe that's just easier to use.

mongorestore/mongorestore_test.go
varsha804: I actually no longer need this import, but will do, Tim!

varsha804:
Instead of asserting collection equality before and after `mongorestore`, I'm now directly querying the `local.oplog.rs` collection for new oplog entries after `mongorestore`. The query filters for the namespaces which should be skipped.

This obviates the need for the testing functions I borrowed from `mongomirror`. @tfogo , should I still add some of those functions to `mongo-tools-common/bsonutil`?

tfogo:
@varsha804 yeah I think adding both to bsonutil would be good!

pmeredit:
Seems to be having problems connecting to clusters?

varsha804:
@rychipman I tried using `start mongod` instead of `start mongod cluster`, but that led to the following error during `rs.initiate()`: "This node was not started with the replSet option". Turns out we do need a separate mongod function!

mongorestore/oplog.go
```diff
@@ -35,6 +35,15 @@ type oplogContext struct {
 	txnBuffer  *txn.Buffer
 }
 
+// shouldIgnoreNamespace returns true if the given namespace should be ignored during applyOps.
+func shouldIgnoreNamespace(ns string) bool {
+	if strings.HasPrefix(ns, "config.cache.") || ns == "config.system.sessions" {
+		log.Logv(log.Always, "skipping applying the "+ns+" namespace in applyOps")
+		return true
```
pmeredit: The jira ticket seems to say that we should be skipping "config.transactions" as well. Was there something outside of the jira comments that caused us to decide to not skip "config.transactions"?

varsha804: I removed the check for the config.transactions namespace here since transactions seem to be skipped anyway in applyOps.

PR #211, opened 2020-02-26
TOOLS-2442: Automate release uploads

rychipman:
This patch adds a buildvariant and task that takes the artifacts generated by all "sign" tasks, and uploads them to the appropriate s3 locations according to the release type.


release/release.go
```diff
@@ -822,6 +822,10 @@ func buildZip() {
 }
 
 func uploadRelease(v version.Version) {
+	if env.EvgIsPatch() {
```
pmeredit: Nice idea!

release/evergreen/evergreen.go
```diff
@@ -0,0 +1,181 @@
...
+		return "", "", fmt.Errorf("could not obtain evergreen username")
+	}
+	if key == "" {
+		return "", "", fmt.Errorf("could not obtain evergreen key")
+	}
+	return user, key, nil
+}
+
+func getAuthInfoFromConfig() (string, string, error) {
+	/* TODO: not supported yet
```
pmeredit: We normally don't leave things like this in the commit :)

However, I have no problem with it! Assuming we plan to add this soon?

pmeredit: I would actually like to make it formal to have TODOs like this as long as they are ticketed (or plan to be ticketed).

release/platform/platform.go
```diff
@@ -77,6 +81,18 @@ func (p Platform) DebianArch() string {
 	}
 }
 
+func (p Platform) ArtifactExtensions() []string {
+	switch p.OS {
+	case OSLinux:
+		return []string{"tgz", p.Pkg}
```
pmeredit: I like `p.Pkg`, good choice.

release/release.go
```diff
@@ -627,6 +658,19 @@ func copyFile(src, dst string) error {
 	return out.Close()
 }
 
+func downloadFile(url, dst string) {
+	out, err := os.Create(dst)
```
pmeredit: Do we really want to download files via HTTP instead of using the aws interface, now that we are importing it? To me, the answer to that would be whether or not there is a parallel downloader in the golang library.

pmeredit:
I just checked, they *only* have parallel uploaders/downloaders. So we are fine on the upload, and might want to consider using the downloader, depending on how much we care about performance.

release/aws/aws.go
```diff
@@ -0,0 +1,65 @@
...
+
+var awsClient *AWS
+
+type AWS struct {
+	session *session.Session
+}
+
+func initializeClient() error {
+	if awsClient != nil {
+		panic("called initializeClient twice")
```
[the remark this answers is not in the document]
rychipman: if it were public, I would definitely return the error. In this case, I chose to panic because this is only called from `GetClient`, and if called twice, we either raced (which shouldn't be happening in these release scripts) or violated some internal invariant

release/evergreen/evergreen.go
rychipman: Yep, I would prefer that too. In this case, I just completely forgot to come back to this and remove it. I don't think it really adds much value for the effort and extra dependency it would add.

rychipman:
Merging in the interest of getting this one out, since this task is a blocker for a few other parts of the epic. We can always come back to perf improvements in a separate ticket, since I think they're non-critical here!

PR #212, opened 2020-02-27
TOOLS-2464: Update platform support

huan-Mongo:
Added extra platform per request in https://docs.google.com/document/d/1_IsRBSCMNs45H80alkai2HvN3WVOg5JUywuEoJKTD1I/edit#heading=h.8709ewb0mee0

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -1485,6 +1485,17 @@ buildvariants:
 #     Debian x86_64 Buildvariants     #
 #######################################
 
+- name: debian71
+  display_name: Debian 7.1
+  run_on:
+    - debian71-test
+  expansions:
+    build_tags: "sasl gssapi ssl"
+  tasks:
+    - name: dist
+    - name: sign
+      run_on: amazon1-2018-test
+
 - name: debian81
   display_name: Debian 8.1
   run_on:
@@ -1506,6 +1517,17 @@ buildvariants:
   - name: dist
   - name: sign
     run_on: amazon1-2018-test
+
+- name: debian10
+  display_name: Debian 10
+  run_on:
+    - debian10-test
+  expansions:
+    build_tags: "sasl gssapi ssl"
+  tasks:
+    - name: dist
+    - name: sign
+      run_on: amazon1-2018-test
 
 #######################################
 #           macOS Buildvariant        #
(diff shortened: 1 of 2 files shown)


common.yml
```diff
@@ -1690,6 +1766,19 @@ buildvariants:
     USE_SSL: "true"
   tasks: *ubuntu1804_arm64_tasks
 
+- name: ubuntu2004-arm64
+  display_name: ZAP ARM64 Ubuntu 20.04
+  run_on:
+    - ubuntu2004-arm64-test
+  stepback: false
+  batchtime: 10080 # weekly
```
rychipman: What's the rationale for these settings?

release/platform/platform.go
```diff
@@ -191,6 +251,24 @@ var platforms = []Platform{
 		OS:   OSLinux,
 		Pkg:  PkgRPM,
 	},
+	{
+		Name: "rhel80",
```
rychipman:
```suggestion
		Name: "rhel82",
```

release/platform/platform.go
```diff
@@ -161,12 +203,24 @@ var platforms = []Platform{
...
+		Pkg:  PkgDeb,
+	},
 	{
 		Name: "rhel71",
 		Arch: "ppc64le",
 		OS:   OSLinux,
 		Pkg:  PkgRPM,
 	},
+	{
+		Name: "rhel80",
```
rychipman:
```suggestion
		Name: "rhel81",
```

common.yml
huan-Mongo: Frankly, I add them here because they are present in the rest of the tests on the same hardware. I guess we just want to make sure the same test runs at most once/week in order to save the resource? 
@xdg Do you know the exact reason? I saw you added the original settings for the rest of the tests. Thanks!

xdg: Yes, the ZAP platforms are elastic -- they are fixed static hosts, so we want to check them occasionally but leave them available for other projects to test against.

huan-Mongo:
Hi @rychipman, I fixed most of the tests. The rests are qa tests which I don't think is related to this change.

tfogo:
What's the source for what platforms we should be supporting? I don't see some of these platforms on the mongodb-mongo-master waterfall. Just want to make sure I understand what we're trying to keep in sync with.

common.yml
rychipman: SGTM. For some reason, I thought batchtime was "run at least this often, even if there was no new commit".

tfogo:
Okay, where do these come from for example:

- ubuntu 20.04
- rhel 8 on Z and PPC

I don't see these referenced anywhere.

rychipman:
Ah, yes, sorry about that...it's all coming back to me now from when I wrote the design. The 4.4 docs are not yet updated with the 4.4 platforms, so I had to go to the 4.4 platform support scope for those: https://jira.mongodb.org/browse/WRITING-3564

huan-Mongo:
I was waiting on the testing finish running, but in the config, some tests are only run once/week. So I took out the batchTime setting and create a separate evergreen patch run to test the newly added variants. https://evergreen.mongodb.com/version/5e595f499ccd4e62dc395296 With the combination of these two patch builds, I think we are much confident this change works. So I am merging the change now.

PR #215, opened 2020-03-03
TOOLS-2464: Fix missing evergreen distros

rychipman:
Right now, our release task cannot run, because it requires the "sign" tasks from all distros to have run. The "sign" tasks are not running on all distros, because some distros are not yet available in evergreen.

This patch contains two commits. The first changes two distros to match seemingly equivalent distros that _do_ exist in evergreen. The second temporarily removes some distros that are not yet supported in evergreen. This second commit should be reverted in TOOLS-2505 once all the distros in question are available.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -1653,16 +1653,6 @@ buildvariants:
 #######################################
 #     SUSE x86_64 Buildvariants       #
 #######################################
-- name: suse11
-  display_name: SUSE 11
-  run_on:
-    - suse11-test
-  expansions:
-    build_tags: "sasl gssapi ssl"
-  tasks:
-    - name: dist
-    - name: sign
-      run_on: amazon1-2018-test
 
 - name: suse12
   display_name: SUSE 12
--- a/common.yml
+++ b/common.yml
@@ -1642,7 +1642,7 @@ buildvariants:
 - name: rhel80
   display_name: RHEL 8.0
   run_on:
-    - rhel80-small
+    - rhel80-test
   expansions:
     build_tags: "sasl gssapi ssl"
   tasks:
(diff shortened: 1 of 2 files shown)


common.yml
```diff
@@ -1860,10 +1836,10 @@ buildvariants:
     run_kinit: true
   tasks: *rhel71_ppc64le_tasks
 
-- name: rhel82-ppc64le
-  display_name: ZAP PPC64LE RHEL 8.2
+- name: rhel81-ppc64le
```
rychipman: @huan-Mongo was there a particular reason you had for not using the rhel81 distro here?

common.yml
```diff
@@ -1951,10 +1914,10 @@ buildvariants:
   - name: sign
     run_on: amazon1-2018-test
 
-- name: rhel82-s390x
+- name: rhel80-s390x
```
rychipman: @huan-Mongo same question here: was there a reason not to use the existing rhel80 distro? Just want to make sure this isn't undoing a decision you made intentionally.

common.yml
huan-Mongo: In 8.1 image build, I got such error ```fork/exec /opt/mongodbtoolchain/v3/bin/ppc64le-mongodb-linux-gcc: no such file or directory
```
Seems toolchain is missing
https://evergreen.mongodb.com/task/mongo_tools_rhel81_ppc64le_dist_patch_37b7176480b317a7b97077ee2297397df922cf0b_5e5839bc850e6122204d8ce9_20_02_27_21_50_53

common.yml
huan-Mongo: Here I use 8.2 to keep it consistent with the above change.

rychipman: The problem here is that there is no rhel82 distro for s390x (hence why the tasks weren't getting scheduled in the patch or on master). The RHEL8 distro that we do have for s390x is rhel80-zseries

common.yml
huan-Mongo: ah, I see. Good catch!

common.yml
huan-Mongo: got it.

PR #216, opened 2020-03-04
TOOLS-2470: Sign linux packages

varsha804:


varsha804:
The signature of the release.rpm package has been verified:

Screen Shot 2020-03-05 at 4 20 41 PM

common.yml
```diff
@@ -641,6 +641,37 @@ functions:
...
+      params:
+        working_dir: src/github.com/mongodb/mongo-tools/
+        silent: true
+        script: |
+          echo "${signing_auth_token_server_4_4}" > ./signing_auth_token
+    - command: shell.exec
+      type: system
+      params:
+        working_dir: src/github.com/mongodb/mongo-tools/
+        silent: true
```
rychipman:
```suggestion
```
Let's not make this one silent, so that we can see if it prints out any messages!

common.yml
```diff
@@ -641,6 +641,37 @@ functions:
...
+        script: |
+          /usr/local/bin/notary-client.py \
```
rychipman:
```suggestion
          set -o errexit
          /usr/local/bin/notary-client.py \
```
Without this, the task won't actually fail if the first notary-client invocation errors, but the second one succeeds

common.yml
```diff
@@ -641,6 +641,37 @@ functions:
...
+              --key-name "server-4.4" \
+              --auth-token-file ./signing_auth_token \
+              --comment "Evergreen Automatic Signing (mongo-tools)" \
+              --notary-url http://notary-service.build.10gen.cc:5000 \
+              --skip-missing \
+              --package-file-suffix "" \
```
rychipman:
```suggestion
              --package-file-suffix "" \
              --outputs "sig" \
```
This was added to the msi signing in a recent commit on master, and we should do the same here. It makes sure that we only sign the package, and don't generate other extraneous artifacts

common.yml
```diff
@@ -641,6 +641,37 @@ functions:
...
+              release.deb
+          /usr/local/bin/notary-client.py \
+              --key-name "server-4.4" \
+              --auth-token-file ./signing_auth_token \
+              --comment "Evergreen Automatic Signing (mongo-tools)" \
+              --notary-url http://notary-service.build.10gen.cc:5000 \
+              --skip-missing \
+              --package-file-suffix "" \
```
rychipman:
```suggestion
              --package-file-suffix "" \
              --outputs "sig" \
```
same as above

common.yml
varsha804: Should we keep the `silent: true` in the `sign msi installer` task?

common.yml
varsha804: Should this also be set in the `sign msi installer` task?

pmeredit: It should be set in an shell task unless we have a specific reason to *not* set it :)

PR #222, opened 2020-03-06
TOOLS-2421: Maintain test coverage after moving tools tests from server

varsha804:
Tim and I renamed the ticket to better reflect its objective.

Steps:

- remove `test/legacy26` and `test/legacy30`
- copy the [latest tools suite](https://github.com/mongodb/mongo/tree/39c3a5d77b976e131d37476f2e7255d6058f5093/jstests/tool) from server into `test/legacy42`
- remove the legacy30 test from evergreen
- add a legacy42 test to evergreen


varsha804:
`dumprestore_auth.js`, `dumprestore_auth2.js`, and `dumprestore_auth3.js` are the only test files which had to be updated by changing `writeConcern: 0` to `writeConcern: 1`. The server test suite was supposed to skip these test files until TOOLS-1371 was resolved; that ticket is now closed.

PR #228, opened 2020-04-08
TOOLS-2538 TOOLS-2536 Move index conversion logic to mongo-tools-common

tfogo:
Also uses ignoreUnknownIndexOptions option in the createIndexes command for servers >4.1.9.

### DIFF
--- a/mongorestore/restore.go
+++ b/mongorestore/restore.go
@@ -20,38 +20,10 @@ import (
 	"github.com/mongodb/mongo-tools-common/util"
 
 	"go.mongodb.org/mongo-driver/bson"
-	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
 )
 
 const insertBufferFactor = 16
-
-// validIndexOptions are taken from https://github.com/mongodb/mongo/blob/master/src/mongo/db/index/index_descriptor.h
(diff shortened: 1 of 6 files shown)


mongorestore/restore.go
```diff
@@ -335,10 +306,16 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
 
-func convertLegacyIndexes(indexes []IndexDocument) {
+func (restore *MongoRestore) convertLegacyIndexes(indexes []IndexDocument) {
 	for _, index := range indexes {
-		convertLegacyIndexKeys(index)
-		convertLegacyIndexOptions(index)
+		bsonutil.ConvertLegacyIndexKeys(index.Key, index.Options["ns"].(string))
+
+		// It is preferable to use the ignoreUnknownIndexOptions on the createIndex command to
+		// force the server to remove unknown options. But ignoreUnknownIndexOptions was only added in 4.1.9.
```
rychipman, line 314: I don't think it should affect this PR, but just curious to get your thoughts @tfogo:
would it be reasonable to ask for a backport of ignoreUnknownIndexOptions?

tfogo: I thought about this and I don't think there's any point. We'll only get support in a patch release so we'll still have to support multiple pre 4.1.9 patches that don't have it. The logic would be the same.

PR #229, opened 2020-04-09
TOOLS-2472: Automate linux package publishing

rychipman:
The evg patch for this PR won't actually do anything, since the release task doesn't release anything on patchbuilds, but it would be good to get some eyes on it anyways.

### DIFF
--- a/release/env/env.go
+++ b/release/env/env.go
@@ -21,14 +21,20 @@ func mustGet(varname string) (string, error) {
 	return val, nil
 }
 
+// EvgIsPatch returns whether the current evergreen task is a patch,
+// based on the value of an env variable set by the evg project file.
 func EvgIsPatch() bool {
 	return get("EVG_IS_PATCH") != ""
 }
 
+// EvgBuildID returns the build_id of the current evergreen task,
+// based on the value of an env variable set by the evg project file.
 func EvgBuildID() (string, error) {
 	return mustGet("EVG_BUILD_ID")
 }
 
+// EvgVariant returns the variant name for the current evergreen task,
+// based on the value of an env variable set by the evg project file.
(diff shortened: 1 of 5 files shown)


release/release.go
```diff
@@ -912,7 +957,116 @@ func uploadRelease(v version.Version) {
...
+		log.Println("current build is a patch; not performing a linux release")
+		return
+	}
+
+	tasks, err := evergreen.GetTasksForRevision(v.Commit)
+	check(err, "get evergreen tasks")
+
+	distTasks := []evergreen.Task{}
+	for _, task := range tasks {
+		if task.IsPatch() || task.DisplayName != "dist" {
```
pmeredit, line 979: how does `task.IsPatch()` differ from `env.EvgIsPatch()`? Little confused here because I don't see how one task could be a patch when the whole build is not. Some comments would help clarify.

rychipman: The former checks whether a task fetched from the API is a patch, while the latter checks whether the task in which this code is running is a patch. I've added godoc comments to the relevant functions to clarify.

release/env/env.go
```diff
@@ -21,14 +21,20 @@ func mustGet(varname string) (string, error) {
 	return val, nil
 }
 
+// EvgIsPatch returns whether the current evergreen task is a patch,
+// based on the value of an env variable set by the evg project file.
 func EvgIsPatch() bool {
 	return get("EVG_IS_PATCH") != ""
 }
 
+// EvgIsPatch returns the build_id of the current evergreen task,
```
pmeredit, line 30: These all say `EvgIsPatch` :)

rychipman: fixed! That's what I get for trying to rush out the update before standup :)

PR #230, opened 2020-04-09
TOOLS-2529: Output new filename format for long collection names

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2529
Design Doc: [https://docs.google.com/document/d/...](https://docs.google.com/document/d/1LCsXHqjQrRWDzD1Of_pNX41pyQZqd231ni9uF9_noY8/edit#heading=h.qk6j6d5j4oc5)
Related PR: https://github.com/mongodb/mongo-tools/pull/233

This PR allows us to output a new filename format for collections whose names would result in filenames >255 characters long.

The functional change is in `prepare.go`, with a new unit test for `outputPath()`. There's also a tiny change in `mongo-tools-common` that I'll do separately after this PR gets approved.

### DIFF
--- a/mongodump/metadata_dump.go
+++ b/mongodump/metadata_dump.go
@@ -19,9 +19,10 @@ import (
 
 // Metadata holds information about a collection's options and indexes.
 type Metadata struct {
-	Options bson.M   `json:"options,omitempty"`
-	Indexes []bson.D `json:"indexes"`
-	UUID    string   `json:"uuid,omitempty"`
+	Options        bson.M   `json:"options,omitempty"`
+	Indexes        []bson.D `json:"indexes"`
+	UUID           string   `json:"uuid,omitempty"`
+	CollectionName string   `json:"collectionName"`
 }
 
 // IndexDocumentFromDB is used internally to preserve key ordering.
@@ -47,6 +48,8 @@ func (dump *MongoDump) dumpMetadata(intent *intents.Intent, buffer resettableOut
 	// If a collection has a UUID, it was gathered while building the list of
 	// intents.  Otherwise, it will be the empty string.
 	meta.UUID = intent.UUID
+
+	meta.CollectionName = intent.C
 
 	// Second, we read the collection's index information by either calling
 	// listIndexes (pre-2.7 systems) or querying system.indexes.
--- a/mongodump/metadata_dump.go
+++ b/mongodump/metadata_dump.go
@@ -19,10 +19,10 @@ import (
 
 // Metadata holds information about a collection's options and indexes.
 type Metadata struct {
-	Options        bson.M   `json:"options,omitempty"`
-	Indexes        []bson.D `json:"indexes"`
-	UUID           string   `json:"uuid,omitempty"`
-	CollectionName string   `json:"collectionName"`
+	Options        bson.M   `bson:"options,omitempty"`
+	Indexes        []bson.D `bson:"indexes"`
+	UUID           string   `bson:"uuid,omitempty"`
+	CollectionName string   `bson:"collectionName"`
 }
 
 // IndexDocumentFromDB is used internally to preserve key ordering.
@@ -49,6 +49,8 @@ func (dump *MongoDump) dumpMetadata(intent *intents.Intent, buffer resettableOut
 	// intents.  Otherwise, it will be the empty string.
 	meta.UUID = intent.UUID
 
+	// Adding the collection name is useful if a long collection name results in a truncated
+	// bson or metadata file name, in which case the collection name can be found here.
 	meta.CollectionName = intent.C
 
 	// Second, we read the collection's index information by either calling
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -1154,3 +1154,57 @@ func TestMongoDumpViews(t *testing.T) {
 
 	})
 }
+
+func TestMongoDumpCollectionOutputPath(t *testing.T) {
+	// testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+	log.SetWriter(ioutil.Discard)
+
+	Convey("testing output paths for collection names of varying lengths", t, func() {
+		md := simpleMongoDumpInstance()
+
+		Convey("don't change a collection name that results in an output path < 255 bytes", func() {
+			md.OutputOptions.Out = "dump"
+
+			// 26 bytes < 255 bytes
+			// (output path will be under 255 bytes, regardless of file extension)
+			colName := "abcdefghijklmnopqrstuvwxyz"
+
+			fileComponents := strings.Split(md.outputPath(testDB, colName), "/")
+			So(len(fileComponents), ShouldEqual, 3)
+
+			filePath := fileComponents[len(fileComponents)-1]
+			So(len(filePath), ShouldEqual, len(colName))
+			So(filePath, ShouldNotContainSubstring, util.EscapeCollectionName("$"))
+		})
+
+		Convey("don't change a collection name that could result in an output path = 255 bytes", func() {
+			md.OutputOptions.Out = "dump"
+
+			// 17 bytes * 14 = 238 bytes
+			// (output would be exactly 255 bytes with longest possible file extension of .metadata.json.gz)
+			colName := strings.Repeat("abcdefghijklmnopq", 14)
+
+			fileComponents := strings.Split(md.outputPath(testDB, colName), "/")
+			So(len(fileComponents), ShouldEqual, 3)
+
+			filePath := fileComponents[len(fileComponents)-1]
+			So(len(filePath), ShouldEqual, len(colName))
+			So(filePath, ShouldNotContainSubstring, util.EscapeCollectionName("$"))
+		})
+
+		Convey("truncate a collection name that results in an output path > 255 bytes", func() {
+			md.OutputOptions.Out = "dump"
+
+			// 26 bytes * 10 = 260 bytes > 255 bytes
+			// (output path is already over the limit, regardless of file extension)
+			colName := strings.Repeat("abcdefghijklmnopqrstuvwxyz", 10)
+
+			fileComponents := strings.Split(md.outputPath(testDB, colName), "/")
+			So(len(fileComponents), ShouldEqual, 3)
+
+			filePath := fileComponents[len(fileComponents)-1]
+			So(len(filePath), ShouldEqual, 238)
+			So(filePath, ShouldContainSubstring, util.EscapeCollectionName("$"))
+		})
+	})
+}
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -9,6 +9,8 @@ package mongodump
 import (
 	"bytes"
 	"context"
+	"crypto/sha1"
+	"encoding/base64"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
@@ -41,6 +43,12 @@ var (
 
 const (
 	KerberosDumpDirectory = "dump-kerberos"
+	longPrefix            = "aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVery" +
+		"VeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVery" +
+		"VeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVery"
+	longCollectionName = longPrefix + "/Long/Collection/Name/Consisting/Of/Many/Characters"
+	longBsonName       = longPrefix + "%2FLong%2FCollection%2FName%24FUVlwTrb2eHN1RUE1swI1fFzWmA.bson"
+	longMetadataName   = longPrefix + "%2FLong%2FCollection%2FName%24FUVlwTrb2eHN1RUE1swI1fFzWmA.metadata.json"
 )
 
 func simpleMongoDumpInstance() *MongoDump {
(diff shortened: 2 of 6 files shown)


mongodump/mongodump_test.go
```diff
@@ -1154,3 +1154,57 @@ func TestMongoDumpViews(t *testing.T) {
 
 	})
 }
+
+func TestMongoDumpCollectionOutputPath(t *testing.T) {
+	// testtype.SkipUnlessTestType(t, testtype.UnitTestType)
```
edobranov, line 1159: Forgot to uncomment this when pushing! I'll add it back in if it's the right type of test for this change.

mongodump/metadata_dump.go
```diff
@@ -19,9 +19,10 @@ import (
 
 // Metadata holds information about a collection's options and indexes.
 type Metadata struct {
-	Options bson.M   `json:"options,omitempty"`
-	Indexes []bson.D `json:"indexes"`
-	UUID    string   `json:"uuid,omitempty"`
+	Options        bson.M   `json:"options,omitempty"`
+	Indexes        []bson.D `json:"indexes"`
+	UUID           string   `json:"uuid,omitempty"`
+	CollectionName string   `json:"collectionName"`
```
edobranov, line 25: I think we might want this to be `json:"collectionName,omitempty"`? I missed this at first.

mattChiaravalloti: Will it ever be empty? Even for collection names that are < 238 bytes long, don't we want to write the name to the meta object?

mongodump/prepare.go
```diff
@@ -194,9 +196,26 @@ func (dump *MongoDump) outputPath(dbName, colName string) string {
...
+		// First 208 bytes of col name + 3 bytes delimiter + 27 bytes base64 hash = 238 bytes.
+		colName = colNameTruncated + delimiter + colNameHashBase64
+	} else {
+		colName = util.EscapeCollectionName(colName)
+	}
+
+	return filepath.Join(root, dbName, colName)
 }
 
+// Unused.
```
mattChiaravalloti, line 218: Did you add this comment because the `checkStringForPathSeparator` function is not used anywhere? If it is not used anywhere, should we delete it?

mongodump/prepare.go
```diff
@@ -341,7 +354,7 @@ func (dump *MongoDump) NewIntentFromOptions(dbName string, ci *db.CollectionInfo
 					Buffer: &bytes.Buffer{},
 				}
 			} else {
-				path := nameGz(dump.OutputOptions.Gzip, dump.outputPath(dbName, ci.Name+".metadata.json"))
+				path := nameGz(dump.OutputOptions.Gzip, dump.outputPath(dbName, ci.Name)+".metadata.json")
```
mattChiaravalloti, line 357: Was this a pre-existing bug?

mongodump/mongodump_test.go
```diff
@@ -1154,3 +1154,57 @@ func TestMongoDumpViews(t *testing.T) {
...
+			// 26 bytes * 10 = 260 bytes > 255 bytes
+			// (output path is already over the limit, regardless of file extension)
+			colName := strings.Repeat("abcdefghijklmnopqrstuvwxyz", 10)
+
+			fileComponents := strings.Split(md.outputPath(testDB, colName), "/")
+			So(len(fileComponents), ShouldEqual, 3)
+
+			filePath := fileComponents[len(fileComponents)-1]
+			So(len(filePath), ShouldEqual, 238)
+			So(filePath, ShouldContainSubstring, util.EscapeCollectionName("$"))
```
mattChiaravalloti, line 1207: We should also verify that the part preceding the `"$"` is the truncated name and the part follows the `"$"` is the encoded hash of the name; this ensures we are outputting the names in the expected format.

mongodump/prepare.go
edobranov: Right, it's dead code now. I can remove it if Tim agrees we wouldn't need it for anything else.

mongodump/metadata_dump.go
edobranov: (Github won't let me reply to your other comment on this @mattChiaravalloti)

We'd be duplicating the collection name if we do this for short names (once in the .metadata.json filename, once in the .metadata.json `collectionName` field). We definitely could as a redundancy measure, but I'm not sure how much value it adds beyond that. We can see what @tfogo thinks.

mongodump/prepare.go
edobranov: Sort of? It didn't make any functional difference per se. Originally we were primarily joining the `colName` parameter with the `root` and `dbName`, so it didn't matter if you tossed in the filename extension or not.

It makes a difference now though when we count the letters in `colName` and when we hash it. The function's comment also says "sans file extension" for `colName`, so I had to make the adjustment here.

mattChiaravalloti: Ok that makes sense. I took a look before commenting and saw that this was the odd-one-out in terms of uses of `nameGz`/`dump.outputPath` so I'm glad you fixed it. Just wanted to make sure it was something that needed to be changed and you confirmed that!

mongodump/metadata_dump.go
mattChiaravalloti: Yea this is a bit of an ambiguous point in the Detailed Design for this part. One bullet point seems to suggest adding the collection name here unconditionally. The next bullet point implies that the collection name can be taken from the filename if one is not found in the metadata. I'll defer to Tim here like you suggested.

tfogo: I think we should always include it. It won't always be used by mongorestore, but I think it's valuable to always have the same format.

mongodump/prepare.go
```diff
@@ -194,9 +196,26 @@ func (dump *MongoDump) outputPath(dbName, colName string) string {
 		root = dump.OutputOptions.Out
 	}
 
-	return filepath.Join(root, dbName, util.EscapeCollectionName(colName))
+	// Encode a new output path for collection names that would result in a file name greater
+	// than 255 bytes long. This includes the longest possible file extension: .metadata.json.gz
+	// The new format is <truncated-url-encoded-collection-name>%24<collection-name-hash-base64>
+	// where %24 represents a $ symbol delimiter (e.g. aVeryVery...VeryLongName%24oPpXMQ...).
+	if len(colName) > 238 {
```
tfogo, line 203: We need to escape the name before checking the length. Escaping can increase the length, so a collection name shorter than 238 bytes could result in a filename longer than 238 bytes.

mongodump/prepare.go
```diff
@@ -194,9 +196,26 @@ func (dump *MongoDump) outputPath(dbName, colName string) string {
...
+		colNameTruncated := util.EscapeCollectionName(colName)[:208]
+		delimiter := util.EscapeCollectionName("$")
```
tfogo, line 205: I'd just append "%24" here instead of escaping $. Escaping a single character seems a bit overkill. You can add a comment indicating that %24 is $ if you're concerned about readability.

mongodump/prepare.go
tfogo: Yep let's delete it!

mongodump/mongodump_test.go
```diff
@@ -1154,3 +1154,57 @@ func TestMongoDumpViews(t *testing.T) {
...
+			// 26 bytes < 255 bytes
+			// (output path will be under 255 bytes, regardless of file extension)
+			colName := "abcdefghijklmnopqrstuvwxyz"
+
+			fileComponents := strings.Split(md.outputPath(testDB, colName), "/")
+			So(len(fileComponents), ShouldEqual, 3)
+
+			filePath := fileComponents[len(fileComponents)-1]
+			So(len(filePath), ShouldEqual, len(colName))
+			So(filePath, ShouldNotContainSubstring, util.EscapeCollectionName("$"))
```
tfogo, line 1177: I think we should test for exact equality with the expected outcome.

mongodump/mongodump_test.go
```diff
@@ -1154,3 +1154,57 @@ func TestMongoDumpViews(t *testing.T) {
...
+			// 17 bytes * 14 = 238 bytes
+			// (output would be exactly 255 bytes with longest possible file extension of .metadata.json.gz)
+			colName := strings.Repeat("abcdefghijklmnopq", 14)
+
+			fileComponents := strings.Split(md.outputPath(testDB, colName), "/")
+			So(len(fileComponents), ShouldEqual, 3)
+
+			filePath := fileComponents[len(fileComponents)-1]
+			So(len(filePath), ShouldEqual, len(colName))
+			So(filePath, ShouldNotContainSubstring, util.EscapeCollectionName("$"))
```
tfogo, line 1192: Again, I think checking for an exact match is better practice.

mongodump/metadata_dump.go
```diff
@@ -19,10 +19,10 @@ import (
...
 // Metadata holds information about a collection's options and indexes.
 type Metadata struct {
-	Options        bson.M   `json:"options,omitempty"`
-	Indexes        []bson.D `json:"indexes"`
-	UUID           string   `json:"uuid,omitempty"`
-	CollectionName string   `json:"collectionName"`
+	Options        bson.M   `bson:"options,omitempty"`
+	Indexes        []bson.D `bson:"indexes"`
+	UUID           string   `bson:"uuid,omitempty"`
+	CollectionName string   `bson:"collectionName"`
```
edobranov, line 25: Funnily enough I don't believe the struct tags weren't doing anything before. The [2nd point in the BSON doc here](https://godoc.org/go.mongodb.org/mongo-driver/bson#hdr-Structs) suggests that the tags start with `bson: ...`

mongodump/prepare.go
```diff
@@ -200,29 +200,17 @@ func (dump *MongoDump) outputPath(dbName, colName string) string {
 	// than 255 bytes long. This includes the longest possible file extension: .metadata.json.gz
 	// The new format is <truncated-url-encoded-collection-name>%24<collection-name-hash-base64>
 	// where %24 represents a $ symbol delimiter (e.g. aVeryVery...VeryLongName%24oPpXMQ...).
-	if len(colName) > 238 {
-		colNameTruncated := util.EscapeCollectionName(colName)[:208]
-		delimiter := util.EscapeCollectionName("$")
+	escapedColName := util.EscapeCollectionName(colName)
+	if len(escapedColName) > 238 {
+		colNameTruncated := escapedColName[:208]
 		colNameHashBytes := sha1.Sum([]byte(colName))
```
edobranov, line 206: Minor question -- do we want a hash of the regular collection name, or the escaped collection name? Currently it's the hash of just the collection name, but I don't think it really makes a difference.

mongodump/mongodump_test.go
```diff
@@ -1188,23 +1239,44 @@ func TestMongoDumpCollectionOutputPath(t *testing.T) {
...
+
+				hashDecoded, _ := base64.RawURLEncoding.DecodeString(filePath[211:])
+				hash := sha1.Sum([]byte(colName))
+				So(bytes.Compare(hashDecoded, hash[:]), ShouldEqual, 0)
+			})
+
+			Convey("with special characters", func() {
+				// (26 bytes + 3 special bytes) * 8 = 232 bytes < 238 bytes
+				// (output path is under the limit, but will go over when we escape the special symbols)
+				colName := strings.Repeat("abcdefghijklmnopqrstuvwxyz+/@", 8)
```
edobranov, line 1268: I'm grouping the unit tests for a '/' symbol in the collection name and encoding to a length >238 bytes. That seems okay to me, but I can separate them if you guys think that's better.

mongodump/mongodump_test.go
```diff
@@ -579,9 +587,47 @@ func TestMongoDumpBSON(t *testing.T) {
...
+
+				absDumpDir := util.ToUniversalPath(filepath.Join(path, "dump_slash"))
+				So(os.RemoveAll(absDumpDir), ShouldBeNil)
+				So(fileDirExists(absDumpDir), ShouldBeFalse)
+
+				dumpDBDir := util.ToUniversalPath(filepath.Join("dump_slash", testDB))
+				So(fileDirExists(dumpDBDir), ShouldBeFalse)
+
+				md.OutputOptions.Out = "dump_slash"
+				err = md.Dump()
```
edobranov, line 616: Unfortunately for this test I couldn't use `testDumpOneCollection()` as it's written, since it hinges on the fact that the bson/metadata file names are 1-to-1 with the collection they came from. It does this when it checks that the dumped bson data is correct.

I was thinking if it's possible to adjust it, but I'd probably end up copying all the new mongorestore code over. It doesn't seem worth it for solely checking that the file names are correct, so I do that manually a few lines down.

tfogo: I think it's fine doing the test here :+1:

mongodump/mongodump_test.go
```diff
@@ -579,9 +587,47 @@ func TestMongoDumpBSON(t *testing.T) {
...
+				So(err, ShouldBeNil)
+				So(fileDirExists(dumpDBDir), ShouldBeTrue)
+
+				Convey("to a bson file", func() {
+					oneBsonFile, err := os.Open(util.ToUniversalPath(filepath.Join(dumpDBDir, longBsonName)))
+					So(err, ShouldBeNil)
+					defer oneBsonFile.Close()
```
tfogo, line 623: Deferring this right at the end of the function seems a bit pointless. I think you can just close it.

mongodump/mongodump_test.go
```diff
@@ -579,9 +587,47 @@ func TestMongoDumpBSON(t *testing.T) {
 					md.OutputOptions.Archive = "dump_slash.archive"
 					err = md.Init()
 					So(err, ShouldBeNil)
+				})
+			})
+
+			Convey("and that it dumps a collection with a name >238 bytes in the right format", func() {
```
tfogo, line 593: This should only be run on >4.4. We'll also need to add a 4.4 task to evergreen. As long as this wrks locally on 4.4, we can add the 4.4 task in another PR. But do make sure to add the 4.4 check on this test now.

mongodump/prepare.go
tfogo: Yeah it doesn't matter as long as we're consistent. Let's do the unescaped collection name.

mongodump/metadata_dump.go
edobranov: Yep that was a good idea. Seems like bson/json struct tags don't completely align in terms of the tags available, among some other use cases. The `MarshallExtJSON()` func is in the bson package anyway, so I suppose that's counter-counterintuitive enough for me :)

mongodump/mongodump_test.go
```diff
@@ -734,6 +694,70 @@ func TestMongoDumpBSON(t *testing.T) {
 	})
 }
 
+func TestMongoDumpBSONLongCollectionName(t *testing.T) {
```
edobranov, line 697: I had to toss this under a new test since I don't believe we can do `t.Skip()` for just one sub-call of `Convey()`.

PR #233, opened 2020-04-16
TOOLS-2530: Restore from new filename format for long collection names

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2530
Design Doc: [https://docs.google.com/document/d/...](https://docs.google.com/document/d/1LCsXHqjQrRWDzD1Of_pNX41pyQZqd231ni9uF9_noY8/edit#heading=h.qk6j6d5j4oc5)
Related PR: https://github.com/mongodb/mongo-tools/pull/230

This PR allows us to restore from the new filename format for collections whose names would result in filenames >255 characters long.

The big functional change is in `filepath.go` in `getCollectionNameFromFile()` (formerly `getInfoFromFilename()`) and its two helpers right below it. I had to restructure it a good amount to allow us to open & inspect metadata files when we need to.

The other changes in the same file are for how we call `getCollectionNameFromFile()`, since there was a lot of assumption that the collection name would always be 1-to-1 with the bson filename.

The rest of the changes are for the tests. For those I added three new files whose names are in the new format:
* a bson file
* ^ a metadata file for the above bson file
* a bson file with no corresponding metadata file


mongorestore/filepath.go
```diff
@@ -220,25 +224,119 @@ func (restore *MongoRestore) getInfoFromFilename(filename string) (string, FileT
...
+// Otherwise the bson file name is truncated, so it returns the collection name found in
+// the corresponding metadata file in the same directory as the bson file.
+//
+// Intended as a helper to getCollectionNameFromFile.
+func (restore *MongoRestore) getCollectionNameFromBson(bsonFullPath string) (string, error) {
+	if bsonFullPath == "" {
+		return "", fmt.Errorf("bson file path is empty")
+	}
+
+	// Return the collection name directly from from the bson file name for files that have
```
edobranov, line 271: "from from" 🙈

mongorestore/testdata/testdirs/db1/aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongCollectionNameC%24Ax2lHbMK_bh19s4f_JCieskppac.metadata.json
```diff
@@ -0,0 +1 @@
+{"options":{},"indexes":[{"v":{"$numberInt":"2"},"key":{"_id":{"$numberInt":"1"}},"name":"_id_"}],"uuid":"5fa8904aa1b543b0bcc9f789192cc92c","collectionname":"aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongCollectionNameConsistingOfExactlyTwoHundredAndFiftyNineCharacters"}
```
edobranov, line 1: Per Slack discussion, the server namespace (\<db\>.\<collection\>) limit is 255 bytes, so the collection name here should be changed from 259 bytes to something like 240-245 bytes -- still hitting the limit to qualify for truncation (238 bytes), but not causing an error on 4.4 with the additional `db1` prefix in the tests.

mongorestore/filepath.go
```diff
@@ -204,13 +204,17 @@ func (f *stdinFile) Close() error {
 	return nil
 }
 
-// getInfoFromFilename pulls the base collection name and FileType from a given file.
-func (restore *MongoRestore) getInfoFromFilename(filename string) (string, FileType, error) {
+// getCollectionNameFromFile returns the collection name and FileType for a bson or metadata file.
+// The collection name may be pulled from either the file name itself, or from the content of a
+// .metadata.json file if the file name is truncated.
+func (restore *MongoRestore) getCollectionNameFromFile(filename string) (string, FileType, error) {
```
mattChiaravalloti, line 210: I have a couple thoughts and questions about this method and its new helpers.

1. The `getCollectionNameFromMetadata` and `getCollectionNameFromBson` methods are very similarly implemented so I wonder if that's the right breakdown of helper methods. The [detailed design section](https://docs.google.com/document/d/1LCsXHqjQrRWDzD1Of_pNX41pyQZqd231ni9uF9_noY8/edit#heading=h.ddsw2s91f9e6) actually suggests always checking the metadata file no matter what. If you fail to find the collection name by doing that, then return an error if the `filename` contains `"%24"`; otherwise you can trim the suffix and what's left is the collection name. This implies breaking down the helpers as `getCollectionNameFromMetadata` and `getCollectionNameFromFilename`. The former would need to ensure the `.metadata.json[.gz]` file exists, can be opened, and contains a `collectionName` field. The latter would need to strip away the suffix. Notice that there is no overlap of functionality between those two, whereas the current two helpers have a lot of the same code duplicated.
2. Given that, the body of this method doesn't exactly suit what we need to do now. I think we can slightly change the suggestion in the design doc to limit the number of places we call helper methods:
	1. Get the suffix and `FileType` from the `filename`. This can be done with the same conditional structure that already exists, just alter the bodies to store the `FileType` and suffix instead of calling helper functions.
	2. If the `filename` contains `"%24"`, get the collection name from the metadata file (`getColllectionNameFromMetadata`). Return an error if we can't get it for some reason.
	3. Otherwise, get the collection name from the file name by trimming the suffix (`getCollectionNameFromFilename`).
	4. Unescape (i.e. the last 6 lines of this method don't need to change).

mattChiaravalloti: (This is just trying to cut back on the duplicated code between the two helpers, but that is not ultimately very important!)

mongorestore/metadata.go
```diff
@@ -35,9 +35,10 @@ type authVersionPair struct {
 
 // Metadata holds information about a collection's options and indexes.
 type Metadata struct {
-	Options bson.D          `bson:"options,omitempty"`
-	Indexes []IndexDocument `bson:"indexes"`
-	UUID    string          `bson:"uuid"`
+	Options        bson.D          `bson:"options,omitempty"`
+	Indexes        []IndexDocument `bson:"indexes"`
+	UUID           string          `bson:"uuid"`
+	CollectionName string          `bson:"collectionname"`
```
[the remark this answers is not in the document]
edobranov: Hmm, it wasn't coming out camelcased as `collectionName` in the metadata file when I was working on the mongodump half, even though I'm pretty sure I wrote it that way.

But yes I totally agree with you. I'll look at them again just to be sure!

mongorestore/filepath.go
edobranov: I tend to agree with your points. If we move the `%24` check up, it can definitely eliminate some duplicated code since we know we'll be looking at a metadata file next.

As a side note, I might've separated them out because I had this idea that they could also be used directly as non-helpers (e.g. if you're in a function that only expects metadata files). There's currently barely any use cases like that though, and likewise I think it's better to keep it compact than offer functions that don't really accomplish anything different than what `getCollectionNameFromFilename` accomplishes.

tfogo: I agree with what Matt laid out. Although the design doc says to always check the metadata file for the collection name, that's inefficient and I think we should only do that if %24 is in the filename.

I don't agree with renaming this function from `getInfoFromFilename` to `getCollectionNameFromFile` by the way. It returns more than just a filename.

edobranov:
I made the changes with the `%24` check you guys suggested and think it looks a lot better.

I wanted to clarify a couple points you made on the tests @tfogo. It might get a bit tedious covering all the cases, so I want to make sure I understand what we want to add before I add them:

> It would be better to create a new dump dir in testdata for the long filename tests. I think doing it in TestMongorestore() is fine, just with new convey blocks.

I created the new dump dir, but we might have to put this in a new test altogether because I think we still need to check the FCV version. If this is right, I'll do it this way for every new test we add.

> We also need to ensure that we properly test interactions with --nsInclude, --nsExclude, --nsFrom and --nsTo.

Right, so I don't see existing tests explicitly testing these. Would the approach then be to write new tests for (1) nsInclude, (2) nsExclude, and (3) nsFrom/nsTo, for normal and also truncated collection names? This could maybe be 2 new tests, with 3 `Convey` blocks each.

> We need to make sure restoring from a single BSON file works for small names without metadata files but fails for long names without metadata files.

This is/was sort of covered in `filepath_test.go` which I temporarily reverted to the original because I figured you meant an integration test. I also reverted it since we were going to separate out the tests anyway. Should we then skip adding tests to `filepath_test.go` or do you still think it's worth it?

tfogo:
> I created the new dump dir, but we might have to put this in a new test altogether because I think we still need to check the FCV version. If this is right, I'll do it this way for every new test we add.

Yeah that's fine.

> Right, so I don't see existing tests explicitly testing these. Would the approach then be to write new tests for (1) nsInclude, (2) nsExclude, and (3) nsFrom/nsTo, for normal and also truncated collection names? This could maybe be 2 new tests, with 3 Convey blocks each.

Yes, that sounds good. The Go tests are incomplete, so any opportunity we get to add test coverage is good. 

> Should we then skip adding tests to filepath_test.go or do you still think it's worth it?

No sorry, what you had in `filepath_test.go` for the single bson file was perfect. No need to add more tests for that. My mistake.

mongorestore/filepath_test.go
```diff
@@ -112,6 +112,47 @@ func TestCreateAllIntents(t *testing.T) {
 	})
 }
 
+func TestCreateAllIntentsLongCollectionName(t *testing.T) {
+	// This tests creates intents based on the test file tree:
+	//   testdata/longcollectionname
+	//   testdata/longcollectionname/db1
```
edobranov, line 118: I think we have to keep all the long collection name tests separate. Both the unit tests and integration tests use the same `testdata/testdirs/` directory, so if we're going to put the truncated bson/metadata files in another spot, that's going to affect how we do both types of tests.

It's kind of annoying with respect to code duplication, but there's no easy way to otherwise test functions like `CreateAllIntents()` that take a single directory.

mongorestore/mongorestore_test.go
```diff
@@ -107,12 +117,86 @@ func TestMongorestore(t *testing.T) {
...
+			So(result.Err, ShouldBeNil)
+			count, err := c1.CountDocuments(nil, bson.M{})
+			So(err, ShouldBeNil)
+			So(count, ShouldEqual, 100)
+
+			restore.NSOptions.Collection = ""
+			restore.NSOptions.DB = ""
+			restore.InputReader = nil
+			restore.TargetDirectory = ""
+			c1.Drop(nil)
```
edobranov, line 131: These options should be reset after each Convey block (otherwise we get conflicting options later on). Dropping the collections also lets us run tests repeatedly without manually dropping them on the command line each time.

mongorestore/mongorestore_test.go
```diff
@@ -107,12 +117,86 @@ func TestMongorestore(t *testing.T) {
...
+			restore.NSOptions.NSInclude[0] = "db1.c3"
+
+			result := restore.Restore()
+			So(result.Err, ShouldBeNil)
+			So(result.Successes, ShouldEqual, 0)
+			So(result.Failures, ShouldEqual, 0)
+
+			count, err := c3.CountDocuments(nil, bson.M{})
+			So(err, ShouldBeNil)
+			So(count, ShouldEqual, 0)
```
edobranov, line 167: Looking at it now, "nsIncluding" a collection with 0 documents might not be the best thing to do :P 

Maybe it's better to add a couple other collections to restore from if y'all think this isn't sufficient (meaning add new bson/metadata files to `testdata/testdirs/db1/`)

mongorestore/filepath.go
```diff
@@ -204,41 +204,105 @@ func (f *stdinFile) Close() error {
...
+	collName, err = restore.getCollectionNameFromMetadata(metadataFullPath)
+	if err != nil {
+		return "", UnknownFileType, err
+	}
+	unescapedCollName, err = util.UnescapeCollectionName(collName)
+	if err != nil {
+		return "", UnknownFileType, fmt.Errorf("error parsing collection name from filename \"%v\": %v", baseFileName, err)
+	}
+
+	return unescapedCollName, fileType, nil
```
mattChiaravalloti, line 266: Notice that lines 249-253 are the same as lines 26-266. You can refactor this so the conditional checks the positive case and acts on that:
```
if strings.Contains(collName, "%24") {
    collName, err = restore.getCollectionNameFromMetadata(metadataFullPath)
    if err != nil { return "", UnknownFileType, err }
}
... // the duplicated lines I mentioned
```

mongorestore/filepath_test.go
```diff
@@ -215,7 +295,7 @@ func TestHandlingBSON(t *testing.T) {
 	Convey("With a test MongoRestore", t, func() {
 		mr = newMongoRestore()
 
-		Convey("with a target path to a bson file instead of a directory", func() {
+		Convey("with a target path to a regular bson file instead of a directory", func() {
```
mattChiaravalloti, line 298:
```suggestion
		Convey("with a target path to a non-truncated bson file instead of a directory", func() {
```

mongorestore/mongorestore_test.go
mattChiaravalloti: I agree that restoring some documents would make this test a little more meaningful. Having no error and getting 0s for successes and failures doesn't indicate much.

mongorestore/filepath.go
edobranov: Yes great point. I was wondering why it was looking a lil long the way I had it 🙃

mattChiaravalloti: I generally prefer conditionals that read as "if true do this" anyway, instead of "if _not_ something, do this"!

mongorestore/filepath_test.go
```diff
@@ -122,6 +170,8 @@ func TestCreateIntentsForDB(t *testing.T) {
 	//   db1/c2.bson
 	//   db1/c3.bson
 	//   db1/c3.metadata.json
+	//   db1/c4.bson
+	//   db1/c4.metadata.json
```
tfogo, line 174: What is c4 testing?

mongorestore/mongorestore_test.go
tfogo: I don't think you should need to reset the options. Take a look at this page in the goconvey docs: https://github.com/smartystreets/goconvey/wiki/Execution-order

In general teardown logic such as dropping collections should be in the Reset function (https://github.com/smartystreets/goconvey/wiki/Reset) or, if you prefer, in a defer statement in the parent convey block.

mongorestore/filepath_test.go
```diff
@@ -112,6 +119,47 @@ func TestCreateAllIntents(t *testing.T) {
...
+			Convey("and reading the intents should show a long collection name", func() {
+				i0 := mr.manager.Pop()
+				So(i0.DB, ShouldEqual, "db1")
+				So(i0.C, ShouldEqual, longCollectionName)
+
+				Convey("with all the proper metadata + bson merges", func() {
+					So(i0.Location, ShouldNotEqual, "")
+					So(i0.MetadataLocation, ShouldNotEqual, "")
+
+					Convey("and skipped files all present in the logs", func() {
```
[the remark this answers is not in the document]
edobranov: Mmm yes that's a good point. I was following the structure of the test above this one, but I can remove this Convey call since it doesn't directly relate to long collection names.

mongorestore/mongorestore_test.go
edobranov: That's really interesting actually! I didn't realize that it ran like the Convey blocks like nodes in a tree :)

I'll just remove these lines then. The collections will persist in `db1` as before, but that doesn't affect successive runs (I was probably doing something wrong before).

mongorestore/filepath_test.go
edobranov: I added it because of [a concern I had here](https://github.com/mongodb/mongo-tools/pull/233#discussion_r416168450). In short, it's to give us more meaningful results for testing --nsInclude and --nsExclude.

I was worried that since c1 is the only collection with docs in it, there might be a case where `--nsInclude db1.c1` could include more than it should and we wouldn't know since c2 and c3 have 0 docs. Same case with `--nsExclude db1.c1` but with opposite results. So I added c4 as a second collection with a non-zero number of docs.

I have to account for c4 in filepath_test.go since it uses the same `testdata/testdirs/db1` directory as mongorestore_test.go

mongorestore/mongorestore_test.go
```diff
@@ -173,13 +155,13 @@ func TestMongorestore(t *testing.T) {
 			So(result.Successes, ShouldEqual, 10)
 			So(result.Failures, ShouldEqual, 0)
 
-			count, err := c4.CountDocuments(nil, bson.M{})
+			count, err := c1.CountDocuments(nil, bson.M{})
```
edobranov, line 158: Forgot to check that c1 isn't being included when we only include c4.

mongorestore/filepath_test.go
tfogo: Ah I see, good solution!

PR #236, opened 2020-04-17
TOOLS-2534: Ignore startIndexBuild and abortIndexBuild oplog entries in oplog replay

varsha804:
In order to get the `startIndexBuild` and `abortIndexBuild` entries in the `oplog.bson` used for testing, I enabled the `hangAfterStartingIndexBuild` failpoint and then killed the index build process.


mongorestore/oplog.go
```diff
@@ -110,6 +110,15 @@ func (restore *MongoRestore) RestoreOplog() error {
 			//skip no-ops
 			continue
 		}
+
+		if restore.serverVersion.GTE(db.Version{4, 4, 0}) {
+			entryName := entryAsOplog.Object[0].Key
```
huan-Mongo, line 115: I am not sure if this is risky always assume `Object` field is not empty. Maybe worth checking it first and add extra condition _entryAsOplog.Operation == "c"_ ? What do you think @tfogo

mongorestore/mongorestore_test.go
```diff
@@ -1084,3 +1084,71 @@ func TestSkipSystemCollections(t *testing.T) {
...
+			result := restore.Restore()
+			So(result.Err, ShouldBeNil)
+
+			Convey("RestoreOplog() should ignore all startIndexBuild and abortIndexBuild oplog entries", func() {
+				queryObj := bson.D{
+					{"$and",
+						bson.A{
+							bson.D{{"ts", bson.M{"$gte": primitive.Timestamp{T: currentTS, I: 1}}}},
+							bson.D{{"$or", bson.A{
+								bson.D{{"o.startIndexBuild", primitive.Regex{Pattern: "skip_index_entries"}}},
```
huan-Mongo, line 1137: This test is not ideal since it's just checking if we didn't create startIndexBuild or abortIndexBuild in the target oplog. But what should be really checked is such entries are ignored.
For example, if we convert the`startIndexBuild` to `createIndexes` in the target this test won't be able to find that. I am not sure if you can create a bson file with startIndexBuild and abortIndexBuild alone and prove we don't create any oplog on the destination?

mongorestore/oplog.go
tfogo: Yeah, I agree. a) check it's a command and b) to be defensive, check that the length of entryAsOplog.Object != 0.

mongorestore/oplog.go
```diff
@@ -110,6 +110,15 @@ func (restore *MongoRestore) RestoreOplog() error {
 			//skip no-ops
 			continue
 		}
+
+		if restore.serverVersion.GTE(db.Version{4, 4, 0}) {
```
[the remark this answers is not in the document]
varsha804: Oh yeah, I don't think we need this check!

mongorestore/mongorestore_test.go
varsha804: I've modified `oplog.bson` so that it only has `startIndexBuild` and `abortIndexBuild` entries, and have modified the test to ensure that nothing is added to the oplog after running mongorestore.

varsha804: I just noticed that the test is failing on evergreen despite passing locally. Currently trying to figure out why the number of documents in `db.local.oplog.rs` > 0 even though the logs say "applied 0 oplog entries".

mongorestore/mongorestore_test.go
```diff
@@ -1084,3 +1084,59 @@ func TestSkipSystemCollections(t *testing.T) {
...
+	session, err := sessionProvider.GetSession()
+	if err != nil {
+		t.Fatalf("No client available")
+	}
+
+	if ok, _ := sessionProvider.IsReplicaSet(); !ok {
+		t.SkipNow()
+	}
+
+	sessionProvider.GetNodeType()
```
tfogo, line 1107: This isn't doing anything?

mongorestore/mongorestore_test.go
```diff
@@ -1084,3 +1084,59 @@ func TestSkipSystemCollections(t *testing.T) {
...
+			DropOption,
+		}
+
+		restore, err := getRestoreWithArgs(args...)
+		So(err, ShouldBeNil)
+
+		if restore.serverVersion.GTE(db.Version{4, 4, 0}) {
+			// Run mongorestore
+			dbLocal := session.Database("local")
+			countBeforeRestore, err := dbLocal.Collection("oplog.rs").CountDocuments(ctx, bson.D{})
```
tfogo, line 1128: I think you should add a filter to these counts to filter out no-ops, just in case one sneaks through.

huan-Mongo: Agree!

varsha804:
Moving this ticket to blocked because integration-4.4-cluster is failing on master. I've created a build failure [ticket](https://jira.mongodb.org/browse/TOOLS-2604) to address this issue.

mongorestore/mongorestore_test.go
```diff
@@ -1128,11 +1126,21 @@ func TestSkipStartAndAbortIndexBuild(t *testing.T) {
 			countBeforeRestore, err := dbLocal.Collection("oplog.rs").CountDocuments(ctx, bson.D{})
 			So(err, ShouldBeNil)
 
+			currentTS := uint32(time.Now().UTC().Unix())
```
tfogo: I think it would be better to get the timestamp from the latest oplog entry. Currently mongod and the test process are on the same host so it's probably fine. But if we ever want to run the tests on a remote host, there could be clock skew issues. So getting the timestamp from the oplog is a little more reliable and future-proof.

mongorestore/mongorestore_test.go
```diff
@@ -1084,3 +1087,59 @@ func TestSkipSystemCollections(t *testing.T) {
...
+			DropOption,
+		}
+
+		restore, err := getRestoreWithArgs(args...)
+		So(err, ShouldBeNil)
+
+		if restore.serverVersion.GTE(db.Version{4, 4, 0}) {
+			// Run mongorestore
+			dbLocal := session.Database("local")
+			queryObj := bson.D{{"op", bson.M{"$ne": "n"}}}
```
tfogo, line 1129: Just realized, I also think you want to ignore `ns: "config.system.sessions"`. Session oplog entries can occur at any time: https://docs.mongodb.com/manual/reference/config-database/#config.system.sessions

mongorestore/mongorestore_test.go
varsha804: I was getting a weird error when I was filtering with `currentTS`, so I changed the method of testing so that now I'm counting _all_ of the documents before and after restore.

mongorestore/mongorestore_test.go
```diff
@@ -1084,3 +1087,65 @@ func TestSkipSystemCollections(t *testing.T) {
...
+
+		restore, err := getRestoreWithArgs(args...)
+		So(err, ShouldBeNil)
+
+		if restore.serverVersion.GTE(db.Version{4, 4, 0}) {
+			// Run mongorestore
+			dbLocal := session.Database("local")
+			queryObj := bson.D{{
+				"$or", bson.A{
+					bson.D{{"ns", primitive.Regex{Pattern: "^config.system.sessions*"}}},
```
tfogo, line 1131: But this is including config.system.sessions, not excluding it. Currently this query is matching `ns == /^config.system.sessions*/ || op != "n"`. I think we need `ns != "config.system.sessions" && op != "n"`

mongorestore/mongorestore_test.go
tfogo: Okay, that works :+1:

mongorestore/mongorestore_test.go
varsha804: Oh, my mistake! I guess I didn't notice the error since there were no `config.system.sessions` logs in the oplog when testing 😅 I think the two conditions should be `or`'d, though.

mongorestore/mongorestore_test.go
```diff
@@ -1084,3 +1087,65 @@ func TestSkipSystemCollections(t *testing.T) {
...
+		}
+
+		restore, err := getRestoreWithArgs(args...)
+		So(err, ShouldBeNil)
+
+		if restore.serverVersion.GTE(db.Version{4, 4, 0}) {
+			// Run mongorestore
+			dbLocal := session.Database("local")
+			queryObj := bson.D{{
+				"$or", bson.A{
```
tfogo, line 1130: This should be `$and`, right? You can merge once that change is made.

PR #239, opened 2020-04-22
TOOLS-2441: Update release process documentation

rychipman:
This PR adds a `RELEASE.md` describing the versioning policy and release process for the tools

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,57 @@
+# Releasing the MongoDB Database Tools
+
+This document describes the version policy and release process for the MongoDB Database Tools.
+
+## Versioning
+
+The MongoDB Database Tools use [Semantic Versioning](https://semver.org/).
+
+We will use the following guidelines to determine when each version component will be updated:
+- **major**: backwards-breaking changes to the CLI API or other behaviors (e.g. exit codes) that could break scripts using the Tools
+- **minor**: new features, including new server version support
+- **patch**: bug fixes and cosmetic changes
+
+The team recognizes that log messages, while intended to be informational, may cause breakage if scripts attempt to parse output.
+While we will not commit to a formal versioning policy for log message changes, we will attempt to adhere to the following guidelines:
+- **log.Always** level (the default): adding logging is a "minor" change; changing or removing logging is a "major" change.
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,56 @@
...
+At the moment, there are no pre-release (alpha, beta, rc, etc.) versions of the Tools.
+If there is a need to support pre-release versions of the Tools, we will need to update our release infrastructure to support them.
+
+## Releasing
+
+This section describes the steps for releasing a new version of the Tools.
+
+### Major Release
+
+We do not expect to make any breaking changes to the tools.
```
tfogo, line 28: I would expect breaking changes as part of CLI option rationalization at least. Saying the release process doesn't support major releases right now is fine, but I think we should expect to make breaking changes in the coming quarters. I could imagine the index work could result in logging changes for example.

pmeredit: agreed

RELEASE.md
```diff
@@ -0,0 +1,56 @@
...
+### Minor/Patch Release
+
+#### Determine the new version
+
+First, determine the version number of the new release.
+The new version should be the smallest possible minor or patch increment from the current version.
+
+For example, consider the version `<x>.<y>.<z>`.
+Bumping the patch version would give version `<x>.<y>.<z+1>`.
+Bumping the minor version would give version `<x>.<y+1>.0`.
```
pmeredit, line 40: At my old job I had the problem of occasionally incrementing versions incorrectly because we cut _so many tags_. We might consider adding a small utility to correctly determine that appropriate tag, like I did there (it will parse the latest tag out of git and increment as appropriate, and create the tag, it required about 80 lines of python... probably even less go since we already have a lot of the plumbing needed for this sort of tool in our release go code).

pmeredit: more specifically `go run release/release.go {patch-tag, minor-tag, major-tag}`

RELEASE.md
rychipman: updated!

RELEASE.md
rychipman: I think that's out of scope for this project, but certainly something you could file a TOOLS ticket for!

pmeredit: sgtm

PR #242, opened 2020-04-30
TOOLS-2537: Ignore config.system.indexBuilds namespace

huan-Mongo:


mongorestore/oplog_test.go
```diff
@@ -191,3 +191,47 @@ func TestOplogRestoreTools2002(t *testing.T) {
...
+			output: false,
+		},
+		{
+			ns:     "test.test",
+			output: false,
+		},
+		{
+			ns:     "config.cache.any",
+			output: true,
+		},
```
mattChiaravalloti, line 230: Is it worth adding one more case with the `config.cache.` prefix?

huan-Mongo: Thanks for reviewing. I guess not much extra value to be added by one more test case.

PR #245, opened 2020-05-05
TOOLS-2498: Nil pointer error mongodump

huan-Mongo:


mongodump/mongodump_test.go
```diff
@@ -1186,7 +1213,9 @@ func TestMongoDumpViews(t *testing.T) {
 		err := setUpMongoDumpTestData()
 		So(err, ShouldBeNil)
 
-		err = setUpDBView()
+		colName := "dump_views"
+		dbName := testDB
+		err = setUpDBView(colName, dbName)
```
rychipman, line 1218: I see some test refactoring in here, but I can't tell if there's anything in here that is actually regression testing the nil pointer bug

mongodump/mongodump_test.go
```diff
@@ -1166,6 +1179,20 @@ func TestMongoDumpViewsAsCollections(t *testing.T) {
 
 			})
 
+			Convey("testing dumping a view query hints", func() {
```
rychipman, line 1182: Two questions:
- Isn't it only supposed to hint the index when the storage engine is mmapv1?
- Should we also be testing this same assertion in `TestMongoDumpViews`?

common.yml
```diff
@@ -20,6 +20,7 @@ mongo_tools_variables:
       - name: integration-4.0
       - name: integration-4.0-auth
       - name: integration-4.0-cluster
+      - name: integration-4.0-mmapv1
```
rychipman, line 23: Does this address https://jira.mongodb.org/browse/TOOLS-2446?

mongodump/mongodump_test.go
huan-Mongo: I was testing another behavior of this bug that we don't hint index for mmapv1 cause we always consider them as views. TestMongoDumpTOOLS1952 has already covered this but we don't have mmapv1 set in evergreen. So I don't need to modify the test after adding mmapv1 to evergreen. 
For the Nil pointer error, I added test TestMongoDumpTOOLS2498 to reproduce it.

mongodump/mongodump.go
```diff
@@ -534,6 +534,8 @@ func (dump *MongoDump) DumpIntent(intent *intents.Intent, buffer resettableOutpu
 	// determine if a collection is a view.
 	collInfo, err := db.GetCollectionInfo(coll)
 	if err != nil {
+		return err
+	} else if collInfo != nil {
```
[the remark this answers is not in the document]
huan-Mongo:
```
	var foundCollInfo *CollectionInfo
	for iter.Next(nil) {
		collInfo := &CollectionInfo{}
		err = iter.Decode(collInfo)
		if err != nil {
			return nil, err
		}
		if collInfo.Name == comparisonName {
			foundCollInfo = collInfo
			break
		}
	}
	if err := iter.Err(); err != nil {
		return nil, err
	}
	return foundCollInfo, nil
```
So if we didn't find the collection name and at the same time, there is no decode error, the `err` and `collInfo` would both be nil?

common.yml
huan-Mongo: Yes, this should have addressed TOOLS-2446

common.yml
```diff
@@ -545,6 +556,20 @@ functions:
         echo "Starting mongod..."
         PATH=$PWD/bin:$PATH ./bin/mongod${extension}  --port ${mongod_port} ${mongod_args} ${additional_args} --dbpath mongodb/db_files --setParameter=enableTestCommands=1
 
+  "start mmapv1 instance":
```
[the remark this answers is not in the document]
huan-Mongo: Sure, I removed the function and added the variable as in the BIC file.

mongodump/mongodump_test.go
```diff
@@ -31,6 +23,13 @@ import (
 	. "github.com/smartystreets/goconvey/convey"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
```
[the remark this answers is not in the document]
huan-Mongo: I accidentally removed the empty line inside and go fmt change the ordering. Reverted the change.

mongodump/mongodump_test.go
```diff
@@ -288,6 +284,40 @@ func setUpDBView() error {
...
+	sessionProvider, _, err := testutil.GetBareSessionProvider()
+	if err != nil {
+		return err
+	}
+
+	profileCmd := bson.D{
+		{"profile", 2},
+	}
+
+	var r2 bson.M
```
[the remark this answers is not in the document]
huan-Mongo: Done!

mongodump/mongodump_test.go
```diff
@@ -265,15 +264,12 @@ func setUpMongoDumpTestData() error {
 	return nil
 }
 
-func setUpDBView() error {
+func setUpDBView(collName string, dbName string) error {
```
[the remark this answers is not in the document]
huan-Mongo: Good idea! Made a change to the signature.

huan-Mongo:
I added a test case to reproduce the nil pointer bug, but it uses failpoints and I need a way to reset it for different test cases. https://github.com/mongodb/mongo-tools-common/pull/35 is the change in tools-common for the failpoints.

mongodump/mongodump_test.go
```diff
@@ -12,13 +12,15 @@ import (
 	"crypto/sha1"
 	"encoding/base64"
 	"fmt"
+	"github.com/mongodb/mongo-tools-common/failpoint"
```
rychipman, line 15: doesn't look like this line is in the right place. This should be grouped with the other non-stdlib imports

mongodump/mongodump_test.go
```diff
@@ -1068,6 +1080,56 @@ func TestMongoDumpTOOLS1952(t *testing.T) {
...
+func TestMongoDumpTOOLS2498(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.IntegrationTestType)
+	log.SetWriter(ioutil.Discard)
+
+	sessionProvider, _, err := testutil.GetBareSessionProvider()
+	if err != nil {
+		t.Fatalf("No cluster available: %v", err)
+	}
+
+	collName := "tools-1952-dump"
```
rychipman, line 1093:
```suggestion
	collName := "tools-2498-dump"
```

mongodump/mongodump_test.go
```diff
@@ -1068,6 +1080,56 @@ func TestMongoDumpTOOLS1952(t *testing.T) {
...
+	createCmd := bson.D{
+		{"create", collName},
+	}
+	var r2 bson.M
+	err = sessionProvider.Run(createCmd, &r2, dbName)
+	if err != nil {
+		t.Fatalf("Error creating collection: %v", err)
+	}
+
+	Convey("testing dumping a collection query hints", t, func() {
```
rychipman, line 1108:
```suggestion
	Convey("failing to get collection info should error, but not panic", t, func() {
```

mongodump/mongodump.go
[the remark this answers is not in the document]
rychipman: yep, makes sense!

PR #246, opened 2020-05-06
TOOLS-2495: Allow mongorestore oplog replay to handle 16 MiB docs

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2495

This PR adds a 16 KiB overhead to the `bsonSource` to accommodate 16 MiB documents and any oplog metadata for the entry.

For the test, I opted in to generate the ~16 MiB `oplog.bson` file when the test is run and then delete it at the end. Figured it wouldn't be super ideal to add a 16 MiB file to the repo for one test.


mongorestore/oplog_test.go
```diff
@@ -167,6 +172,106 @@ func TestOplogRestore(t *testing.T) {
 	})
 }
 
+func TestOplogRestoreMaxDocumentSize(t *testing.T) {
+	// testtype.SkipUnlessTestType(t, testtype.IntegrationTestType)
```
edobranov, line 176: Left this out again 🙃Will uncomment before merging!

mongorestore/oplog_test.go
```diff
@@ -167,6 +172,106 @@ func TestOplogRestore(t *testing.T) {
...
+	fcv := testutil.GetFCV(session)
+	var shouldPreserveUUID bool
+	if cmp, err := testutil.CompareFCV(fcv, "3.6"); err != nil || cmp >= 0 {
+		shouldPreserveUUID = true
+	}
+
+	c1 := session.Database("db1").Collection("c1")
+	c1.Drop(nil)
+
+	Convey("With a test MongoRestore", t, func() {
```
tfogo, line 191: This should be more descriptive of the test that's happening.

PR #249, opened 2020-05-15
TOOLS-2593: Clarify where $admin dbs are skipped

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2593

This just adds a clarifying comment for the change [here in this commit](https://github.com/mongodb/mongo-tools/commit/60114507c1f63057163eb415e87fd7f91a12bf6f#diff-c00634e277d3e3e4a379a8318bcd17d1R248). Lmk if you think something should be rephrased!

### DIFF
--- a/mongorestore/filepath.go
+++ b/mongorestore/filepath.go
@@ -245,6 +245,12 @@ func (restore *MongoRestore) getInfoFromFile(filename string) (string, FileType,
 	}
 
 	// If the collection name is truncated, parse the full name from the metadata file.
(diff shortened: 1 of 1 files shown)


edobranov:
For sure! I agree if they aren't collections, then we shouldn't call them that.

We should then change [this comment about them](https://github.com/mongodb/mongo-tools/blob/d62ed5596c4ade75147d81e98768708bcb8b1091/mongorestore/filepath.go#L419) too. Maybe to something like "db-specific files" or "special db files", if one sounds better to you.

tfogo:
Yeah I agree. I think either option is good.

PR #250, opened 2020-05-19
TOOLS-2369: IAM Role-based authentication

varsha804:
[Technical Design](https://docs.google.com/document/d/1LCsXHqjQrRWDzD1Of_pNX41pyQZqd231ni9uF9_noY8/edit#)

I've added integration tests for `mongodump`, `mongoimport`, `mongoexport`, `mongofiles`, and `mongorestore`, but not the other tools (`mongotop`, `mongostat`, and `bsondump`) since they either don't have any integration tests, or don't test any kind of auth mechanism (e.g. Kerberos or username/password).

The `aws-auth` Evergreen task that I created isn't being run on Windows or RHEL 6.2 since they don't have python3, which is needed for `boto3`.


Gopkg.toml
```diff
@@ -29,7 +29,7 @@ required = ["github.com/3rf/mongo-lint"]
 
 [[constraint]]
   name = "github.com/mongodb/mongo-tools-common"
-  version = "v3.0.7"
+  revision = "2ecb309551185b1ca80cb3c52e2cafe19e0e607c"
```
varsha804, line 32: Temporarily pointing to [TOOLS-2369 in MTC](https://github.com/mongodb/mongo-tools-common/pull/36).

common.yml
```diff
@@ -959,6 +974,106 @@ functions:
...
+          fi;
+          if [ "${AUTH_TEST}" = "true" ]; then
+            export TOOLS_TESTING_AUTH="true"
+          fi;
+          if [ "${create_mongod_users_command}" != "" ]; then
+            export TOOLS_TESTING_AUTH_USERNAME=${auth_username}
+            export TOOLS_TESTING_AUTH_PASSWORD=${auth_password}
+            echo "${create_mongod_users_command}" | ./bin/mongo${extension} $MONGO_ARGS admin
+          fi;
+          export ON_EVERGREEN="true"
```
huan-Mongo, line 1072: Should you set           ```export TOOLS_TESTING_INTEGRATION="true"```, so that aws tests can skip unit tests?

mongodump/mongodump_test.go
```diff
@@ -1400,3 +1401,51 @@ func TestMongoDumpCollectionOutputPath(t *testing.T) {
 		})
 	})
 }
+
+func TestMongoDumpAwsAuth(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.AWSAuthTestType)
+	Convey("Should be able to run mongodump with AWS STS AssumeRole auth", t, func() {
```
huan-Mongo, line 1407: Imo, you don't need to create extra test cases for AWS Auth since the change only applies during connection. You can change `simpleMongoDumpInstance` to make a special case for opts if a special `AWS_AUTH` flag is set just like what `GetSSLOptions` did. In such way, you don't need `testtype.AWSAuthTestType` either, you can just run any integration test in the AWS_AUTH mode.
Same applies to other tests.

mongofiles/mongofiles_test.go
```diff
@@ -32,24 +32,46 @@ var (
...
-		Verbosity:  &options.Verbosity{},
-		URI:        &options.URI{},
-	}
-	testFiles = map[string]primitive.ObjectID{"testfile1": primitive.NewObjectID(), "testfile2": primitive.NewObjectID(), "testfile3": primitive.NewObjectID()}
+	ssl         = testutil.GetSSLOptions()
+	toolOptions = setupToolOptions()
+	testFiles   = map[string]primitive.ObjectID{"testfile1": primitive.NewObjectID(), "testfile2": primitive.NewObjectID(), "testfile3": primitive.NewObjectID()}
 )
 
+func setupToolOptions() *options.ToolOptions {
```
huan-Mongo, line 40: Is there any reason it cannot reuse the function in tools-common?

mongodump/mongodump_test.go
varsha804: Hmm, I was following the example set by the Kerberos tests -- any reason why there's a separate test for that kind of connection, as opposed to just setting a flag? I figured we didn't really need to run every integration test against a mongod in AWS mode, since we are just testing the connection, as you said, and a simple test for each tool would suffice. I think keeping it this way would simplify the structure of testing; toggling a flag might be less convenient. Thoughts?

huan-Mongo: I got your reason for creating a new test case for the AWS auth now. Not sure about the reason for Kerberos to implement its own test case, but from what I see, reusing the existing test cases and added a flag for AWS Auth specific options will definitely simplifying the test structures. You might not need `run-mongodb-aws-test.sh`at all (with some tweak in the runTest.sh) and you don't need to create extra test cases for each Tools product. 
But definitely current implementation also works. 
I'd like to hear @tfogo 's thought on this matter.

varsha804:
Discussed offline: we will merge the IAM auth PRs as they are right now, and then do all of the refactoring work in a separate ticket. This way, we won't have to rush to fix the tests in order to get the feature in the next release.

mongodump/mongodump_test.go
huan-Mongo: As discussed, this would be a separate ticket.

PR #253, opened 2020-06-10
TOOLS-2505: Add missing 4.4 platforms

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2505

This PR adds in the missing 4.4 platforms on Evergreen.

It's not totally clear to me which of the added variants we want to keep, but it looks like some of the ZAP variants haven't been configured yet [from this patch build I ran](https://evergreen.mongodb.com/version/5ee13cee5623435feed3a1a1). We can decide that now or ask the Evergreen team about it.


rychipman:
whoops; never mind! just saw that you linked them in your first comment!

release/platform/platform.go
```diff
@@ -215,32 +222,49 @@ var platforms = []Platform{
...
 	{
-		Name:  "rhel67",
-		Arch:  "s390x",
-		OS:    OSLinux,
-		Pkg:   PkgRPM,
-		Repos: []string{RepoOrg, RepoEnterprise},
+		Name: "rhel67",
+		Arch: "s390x",
+		OS:   OSLinux,
+		Pkg:  PkgRPM,
```
rychipman, line 267: why was the "repos" key removed from this platform and multiple others?

edobranov: Hmm, I didn't notice this at first so good catch! Most of the changes to this file were from when I reverted one of the commits mentioned in the ticket, so something might've gotten lost in translation there on my end. I'll double-check everything to make sure it's as-expected.

edobranov:
Adding the repos key back in seemed to fix some of the weird diff results too, probably because the auto spacing resulted in a lot of changed lines the first time.

Regarding Evergreen zseries tasks, it seems like there's some networking issues the team is working through right now (see https://jira.mongodb.org/browse/BUILD-10986). I asked on Slack if they had an estimate on when this would be done / if we should hold off on merging this PR while we wait, but they recommended not waiting on it and merging without those tasks.

If that's what we wanna do, I can remove the tasks that aren't set up yet and make a ticket to add them in later that'll be blocked on the above one?

rychipman:
yep, if those variants can't run yet, then they'll prevent us from cutting any tools releases. So I think your suggestion is a good one!
- removing them from this PR
- creating a new ticket for adding those platforms
- marking that new ticket as blocked on the BUILD ticket

edobranov:
Cool, lmk how those changes seem! I also created https://jira.mongodb.org/browse/TOOLS-2611

release/platform/platform.go
```diff
@@ -235,6 +256,13 @@ var platforms = []Platform{
 		Pkg:   PkgDeb,
 		Repos: []string{RepoOrg, RepoEnterprise},
 	},
+	{
+		Name:  "ubuntu2004",
+		Arch:  "ppc64le",
+		OS:    OSLinux,
+		Pkg:   PkgDeb,
+		Repos: []string{RepoOrg, RepoEnterprise},
+	},
```
rychipman, line 265: looks like this wasn't added to the evg yml file

edobranov:
Ok, so I reconfigured the PR Evergreen patch here to run the following three:
- Ubuntu 20.04
- ZAP ARM64 Ubuntu 20.04
- ZAP PPC64LE RHEL 8.1

Like you said I had an entry in platforms.go for `ubuntu2004-ppc64le` and not in common.yml, but upon looking at the Evergreen distros, it doesn't look like we actually have an `ubuntu2004-power8` distro in the list.

I'm still waiting on a response to see if that's something that we expect to be available. I removed the `ubuntu2004-ppc64le` entry in platforms.go for the moment, so we can either wait to see what they say or merge now if we want to add it as part of [TOOLS-2611](https://jira.mongodb.org/browse/TOOLS-2611).

PR #256, opened 2020-06-23
TOOLS-2363: Update warning message for "mongorestore"

hariamoor:
[TOOLS-2363](https://jira.mongodb.org/browse/TOOLS-2363) is a simple ticket to update the warning messages provided in `mongorestore` regarding the `--db` and `--collection` flags. The two are deprecated in favor of `--nsInclude` in all cases but that in which the user wants to restore a **simple BSON file**, as specified in [the Tools' public docs](https://docs.mongodb.com/database-tools/mongorestore/#cmdoption-mongorestore-db).

The correct behavior is as follows:

If either of `--db` or `--collection` are specified, then check if the input is a single BSON file. If so, then proceed without warning. Otherwise, issue the following deprecation warning:

> The --db and --collection flags are deprecated for this use-case; please use `--nsInclude` instead, i.e. with `--nsInclude=${DATABASE}.${COLLECTION}`

**Test Plan**

A test case for this behavior is provided together with the other Golang integration tests.

~~A small caveat is that this doesn't necessarily _need_ to be an integration test; at the very least, it doesn't depend on a database instance. However, the code to validate command-line options is coupled with the core application logic of `mongorestore`, which, naturally, _does_ depend on integration tests.~~

~~A further "TODO" is supplied to investigate whether it would be possible to decouple the parsing/validation logic from the core application logic. It might be possible, with sufficient refactoring, to have any number of _unit_ tests on the validation logic without the logical overhead of coupling with the application logic.~~

With assistance from @mattChiaravalloti, the previous caveats should no longer be a concern.

Edit: I investigated the CI failure. It pertains to a JS test regarding sharded secondary reads, which I'm almost certain has nothing to do with the contents of this PR.

### DIFF
--- a/mongorestore/mongorestore.go
+++ b/mongorestore/mongorestore.go
@@ -111,7 +111,6 @@ func New(opts Options) (*MongoRestore, error) {
 		ProgressManager: progressManager,
 		serverVersion:   serverVersion,
 	}
-
 	return restore, nil
 }
 
@@ -215,22 +214,11 @@ func (restore *MongoRestore) ParseAndValidateOptions() error {
 
 	// deprecations with --nsInclude --nsExclude
 	if restore.NSOptions.DB != "" || restore.NSOptions.Collection != "" {
-		// log a deprecation message if restoring from a directory, or
-		// from a non-bson file with the --archive option
-		_, fileType, err := restore.getInfoFromFile(restore.TargetDirectory)
-		if err != nil {
-			return err
-		}
-
-		isDir := false
-		if target, err := newActualPath(restore.TargetDirectory); err == nil {
-			isDir = target.IsDir()
-		}
-
-		if isDir || (fileType != BSONFileType && restore.InputOptions.Archive != "") {
-			log.Logvf(log.Always, "the --db and --collection args should only be used when "+
-				"restoring from a BSON file. Other uses are deprecated and will not exist "+
-				"in the future; use --nsInclude instead")
+		if filepath.Ext(restore.TargetDirectory) != ".bson" {
+			log.Logvf(log.Always,
+				`The --db and --collection flags are deprecated for 
+                                 this use-case; please use --nsInclude instead, 
+                                 i.e. with --nsInclude=${DATABASE}.${COLLECTION}`)
 		}
 	}
 	if len(restore.NSOptions.ExcludedCollections) > 0 ||
--- a/mongorestore/mongorestore.go
+++ b/mongorestore/mongorestore.go
@@ -35,7 +35,7 @@ const (
 )
 
 const (
-	DEPRECATED_DB_AND_COLLECTION_WARNING = "The --db and --collection flags are deprecated for " +
+	deprecated_db_and_collection_options_warning = "The --db and --collection flags are deprecated for " +
 		"this use-case; please use --nsInclude instead, " +
 		"i.e. with --nsInclude=${DATABASE}.${COLLECTION}"
 )
@@ -221,7 +221,7 @@ func (restore *MongoRestore) ParseAndValidateOptions() error {
 	// deprecations with --nsInclude --nsExclude
 	if restore.NSOptions.DB != "" || restore.NSOptions.Collection != "" {
 		if filepath.Ext(restore.TargetDirectory) != ".bson" {
-			log.Logvf(log.Always, DEPRECATED_DB_AND_COLLECTION_WARNING)
+			log.Logvf(log.Always, deprecated_db_and_collection_option_warning)
 		}
 	}
 	if len(restore.NSOptions.ExcludedCollections) > 0 ||
--- a/mongorestore/mongorestore.go
+++ b/mongorestore/mongorestore.go
@@ -34,6 +34,12 @@ const (
 	progressBarWaitTime = time.Second * 3
 )
 
+const (
+	DEPRECATED_DB_AND_COLLECTION_WARNING = "The --db and --collection flags are deprecated for " +
+		"this use-case; please use --nsInclude instead, " +
+		"i.e. with --nsInclude=${DATABASE}.${COLLECTION}"
+)
+
 // MongoRestore is a container for the user-specified options and
 // internal state used for running mongorestore.
 type MongoRestore struct {
@@ -215,10 +221,7 @@ func (restore *MongoRestore) ParseAndValidateOptions() error {
 	// deprecations with --nsInclude --nsExclude
 	if restore.NSOptions.DB != "" || restore.NSOptions.Collection != "" {
 		if filepath.Ext(restore.TargetDirectory) != ".bson" {
-			log.Logvf(log.Always,
-				`The --db and --collection flags are deprecated for 
-                                 this use-case; please use --nsInclude instead, 
-                                 i.e. with --nsInclude=${DATABASE}.${COLLECTION}`)
+			log.Logvf(log.Always, DEPRECATED_DB_AND_COLLECTION_WARNING)
 		}
 	}
 	if len(restore.NSOptions.ExcludedCollections) > 0 ||
--- a/mongorestore/mongorestore.go
+++ b/mongorestore/mongorestore.go
@@ -35,7 +35,7 @@ const (
 )
 
 const (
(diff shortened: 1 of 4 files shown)


mongorestore/mongorestore.go
```diff
@@ -215,22 +214,11 @@ func (restore *MongoRestore) ParseAndValidateOptions() error {
...
-
-		if isDir || (fileType != BSONFileType && restore.InputOptions.Archive != "") {
-			log.Logvf(log.Always, "the --db and --collection args should only be used when "+
-				"restoring from a BSON file. Other uses are deprecated and will not exist "+
-				"in the future; use --nsInclude instead")
+		if filepath.Ext(restore.TargetDirectory) != ".bson" {
+			log.Logvf(log.Always,
+				`The --db and --collection flags are deprecated for 
+                                 this use-case; please use --nsInclude instead, 
+                                 i.e. with --nsInclude=${DATABASE}.${COLLECTION}`)
```
mattChiaravalloti, line 221: The use of backticks here for the string literal means that the log message will print with all of the spaces. i.e., it'll print out like this:
```
$ mongorestore --db d --coll c dir/
The --db and --collection flags are deprecated for
                            this use-case; please use --nsInclude instead,
                            i.e. with --nsInclude=${DATABASE}.${COOLECTION}
```

I suggest using the old method of multi-line strings: double-quotes and breaking it up across newlines with `+`.

mongorestore/mongorestore_test.go
```diff
@@ -69,6 +70,57 @@ func TestMongorestore(t *testing.T) {
...
 
 	Convey("With a test MongoRestore", t, func() {
+		// As specified in TOOLS-2363, the --db and --collection options
+		// are well-defined only for restoration of a single BSON file
+		Convey("and proper warning is issued if --db and --collection are not supported", func() {
+			// Hack to make log readable when testing
+
+			// TODO: Handle all logging in main so that ParseOptions
+			// and ParseAndValidateOptions can idiomatically return
+			// testable error values
```
mattChiaravalloti, line 80: We avoid leaving "TODOs" in production code. If you find yourself adding TODOs for future work, you should file a ticket so we can triage and schedule it.

mongorestore/mongorestore_test.go
```diff
@@ -69,6 +70,57 @@ func TestMongorestore(t *testing.T) {
 	}
 
 	Convey("With a test MongoRestore", t, func() {
+		// As specified in TOOLS-2363, the --db and --collection options
+		// are well-defined only for restoration of a single BSON file
+		Convey("and proper warning is issued if --db and --collection are not supported", func() {
```
mattChiaravalloti, line 75: I don't think this is the best spot for these tests. Look at the other tests in this file, i.e. `TestDeprecatedIndexOptions`. Perhaps you can model your new tests off of that. You aren't using any of the setup code in the outer convey block so these tests probably shouldn't be here. My suggestion is to create a new test function, something like `TestDeprecatedDbAndCollectionOptions`, and move these convey blocks there.

mongorestore/mongorestore_test.go
```diff
@@ -69,6 +70,57 @@ func TestMongorestore(t *testing.T) {
...
+
+				restore, err := getRestoreWithArgs(args...)
+
+				if err != nil {
+					t.Errorf("Cannot connect to database: %v", err.Error())
+				}
+				err = restore.ParseAndValidateOptions()
+
+				So(err, ShouldBeNil)
+				So(buffer.String(), ShouldNotBeEmpty)
```
mattChiaravalloti, line 120: I know it is hacky and not ideal, but should we do more than check that the output is non-empty? It looks like we do some direct error message matching in other tests. Again, that's not ideal but is relevant. See my comment above about the output of this error message. Without a direct test, you didn't catch that it is oddly formatted.

huan-Mongo:
Agree with Matt on the new test cases! I believe it's better to create a separate one for the this change. Other parts look good!

mongorestore/mongorestore_test.go
hariamoor: The other tests are of the form `So(err.Error(), ShouldStartWith, "indextest.test collection: error...")`. I don't like this because you need one string literal for the message itself in core logic and then another for the message prefix in the test. This essentially duplicates the same invariant, which would make refactoring rather tedious in the future (you have to move things around in both places).

I think I can just write the warning as a global `const string` or some such, and then use it to directly check for equality. That way, we can maintain correctness of the code without depending on duplicated invariant literals like `TestDeprecatedIndexOptions` does.

mongorestore/mongorestore_test.go
hariamoor: Fair point. Consider it done!

mongorestore/mongorestore.go
hariamoor: Great catch! Thanks a lot for noticing this! I'll push a fix before merging.

mongorestore/mongorestore_test.go
hariamoor: Thanks for the suggestion! Consider it done!

mongorestore/mongorestore_test.go
```diff
@@ -61,6 +62,62 @@ func getRestoreWithArgs(additionalArgs ...string) (*MongoRestore, error) {
 	return restore, nil
 }
 
+func TestDeprecatedDBAndCollectionOptions(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	Convey("and proper warning is issued if --db and --collection are not supported", func() {
```
mattChiaravalloti, line 68: [nit] can you reword this so it doesn't start with "and"?

mongorestore/mongorestore.go
```diff
@@ -34,6 +34,12 @@ const (
 	progressBarWaitTime = time.Second * 3
 )
 
+const (
+	DEPRECATED_DB_AND_COLLECTION_WARNING = "The --db and --collection flags are deprecated for " +
```
mattChiaravalloti, line 38: I don't think this needs to be exported. In go, if an element starts with a capital letter, it is "exported", meaning consumers of the package defining that element can use it. In this case, there's no need for anything outside this package to use this error message (the tests are in the same package for `mongorestore`), so let's no export it.

mongorestore/mongorestore_test.go
```diff
@@ -69,6 +126,9 @@ func TestMongorestore(t *testing.T) {
 	}
 
 	Convey("With a test MongoRestore", t, func() {
+		// As specified in TOOLS-2363, the --db and --collection options
```
huan-Mongo, line 129: Do we still need the comment here?

hariamoor: We need the comment, but it seems to be in the wrong location. I'll move it to the test case that verifies this behavior so that it's clearer what that's regarding.

mongorestore/mongorestore.go
```diff
@@ -34,6 +34,12 @@ const (
 	progressBarWaitTime = time.Second * 3
 )
 
+const (
+	deprecated_db_and_collection_options_warning = "The --db and --collection flags are deprecated for " +
```
mattChiaravalloti, line 38: [nit] golang typically does not use snake_case
```suggestion
	deprecatedDbAndCollectionOptionsWarning = "The --db and --collection flags are deprecated for " +
```

mongorestore/mongorestore_test.go
```diff
@@ -61,6 +62,65 @@ func getRestoreWithArgs(additionalArgs ...string) (*MongoRestore, error) {
 	return restore, nil
 }
 
+func TestDeprecatedDBAndCollectionOptions(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	/* As specified in TOOLS-2363, the --db and --collection options
+	are well-defined only for restoration of a single BSON file */
```
mattChiaravalloti, line 69: [nit]
```suggestion
	// As specified in TOOLS-2363, the --db and --collection options
	// are well-defined only for restoration of a single BSON file.
```
[nit] We typically write multi-line comments with the `//` on each line, as opposed to `/* ... */`. This also applies to the comment below.

PR #259, opened 2020-06-26
TOOLS-2622: Update readme build instructions

sphanse99:


mattChiaravalloti:
> The change is not correct since go clone actually creates `github.com/mongodb/mongo-tools` folder. you just need to cd into the $GOPATH to run `go clone`. Maybe you can try with the command locally first?

I don't think that's correct. `git clone` just creates the `mongo-tools` directory when I run it.

PR #260, opened 2020-07-01
TOOLS-2638: Fix TestDeprecatedDBAndCollectionOptions

hariamoor:
[TOOLS-2363](https://jira.mongodb.org/browse/TOOLS-2363) introduced a subtle bug in the test case - the fix is given in a single character.


rychipman:
> TOOLS-2638: Trivial one-character fix of TestDeprecatedDBAndCollectionOptions
Commit messages should be in the imperative mood. Additionally, they aren't really the appropriate place for this sort of editorialization; `Fix TestDeprecatedDBAndCollectionOptions` would be just fine here.

hariamoor:
> > TOOLS-2638: Trivial one-character fix of TestDeprecatedDBAndCollectionOptions
> 
> Commit messages should be in the imperative mood. Additionally, they aren't really the appropriate place for this sort of editorialization; `Fix TestDeprecatedDBAndCollectionOptions` would be just fine here.

Thanks! Will fix the commit message!

hariamoor:
The current build failures and known and should be solved in a different PR. See the base vs. patch status of Evergreen tests to ensure that this does indeed fix the pertinent integration test.

PR #261, opened 2020-07-01
 TOOLS-2287: URI parser incorrectly prints unsupported parameter warnings

hariamoor:
Removes some "deprecated" code associated with URI options.

Depends [#40 in MTC](https://github.com/mongodb/mongo-tools-common/pull/40) - remove all uses of the "deprecated" interface before "dropping support" for it (I put this in quotes because we don't technically apply versioning to MTC in a way where these things matter)


tfogo:
> we don't technically apply versioning to MTC in a way where these things matter
mongo-tools-common does use semver. 

We probably want this PR to include the version of mtc being submitted in mongodb/mongo-tools-common#40, that way we know that the combination works correctly together and we won't have to submit a different PR for the revendor. So you'll need to do the following:

1. In mongo-tools, temporarily point dep to your new version of mongo-tools-common. You can use the [source](https://golang.github.io/dep/docs/Gopkg.toml.html#source) and [revision](https://golang.github.io/dep/docs/Gopkg.toml.html#revision) rules to do this. 
2. Use dep ensure to revendor your version to mtc in mongo-tools.
3. Push this so that it's part of the tools PR
4. Once the tools PR and mtc PR have both been approved, merge the mtc PR.
5. Cut a new version of mtc. In this case, since you're making a breaking change to mtc, you will need to bump the major version number.
6. Go back to tools, and make dep point to the new mtc version.
7. Push this change, get approval, then merge.

This revendoring dance can be a bit awkward, so let us know if you need a hand with the order of operations.

By the way, I notice you made these changes on `hariamoor:master`. I'd suggest always making changes on a feature branch instead of master. Otherwise your repo could get into a messy state.

hariamoor:
@tfogo Sounds good. I've put this PR back into draft so it stays out of the way of any future reviews.

hariamoor:
@tfogo I've implemented Steps 1-3 of your comment, with some help from Huan. I had `dep` point to the branch associated with [mongodb/mongo-tools-common#40](https://github.com/mongodb/mongo-tools-common/pull/40) in my personal repo, since it's been approved, but not yet merged.

I assume that this will be fine for testing purposes, but I do plan to formally cut a new major version of MTC and have it point to that after this PR is approved, as specified in Steps 5-6 of your comment:

> 5. Cut a new version of mtc. In this case, since you're making a breaking change to mtc, you will need to bump the major version number.
> 6. Go back to tools, and make dep point to the new mtc version.

@huan-Mongo @mattChiaravalloti I can move forward once this PR is approved, and it's currently ready for review.

dragonlee8:
Hi @hariamoor , I saw there are a bunch of dependencies other than MTC were also updated in this PR, such as "google/go-cmp". Did you do that manually?  Can you just run `dep ensure -update github.com/mongodb/mongo-tools-common` so that it only updates MTC as needed?

hariamoor:
@dragonlee8 Yes, I ran `dep ensure -update`, without specifying MTC as the only dependency to update. I'll make the change requested. Thanks for spotting this.

huan-Mongo:
Cool. thx. Sorry I wasn't aware I was in another github account when doing the update :-)

PR #265, opened 2020-07-08
TOOLS-2644: authenticate to barque using API key

Kimchelly:
JIRA: https://jira.mongodb.org/browse/TOOLS-2644

This change just uses the barque service user's API key instead of LDAP username/password to auth because the LDAP server is not supported anymore.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -1075,7 +1075,7 @@ pre:
           export NOTARY_KEY_NAME='server-4.4'
           export NOTARY_TOKEN='${signing_auth_token_server_4_4}'
           export BARQUE_USERNAME='${evg_user}'
-          export BARQUE_PASSWORD='${evg_password}'
+          export BARQUE_API_KEY='${barque_api_key}'
           set -o xtrace
           set -o verbose
           set -o errexit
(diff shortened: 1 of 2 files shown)


common.yml
```diff
@@ -1075,7 +1075,7 @@ pre:
           export NOTARY_KEY_NAME='server-4.4'
           export NOTARY_TOKEN='${signing_auth_token_server_4_4}'
           export BARQUE_USERNAME='${evg_user}'
```
rychipman, line 1077:
```suggestion
```
since this isn't used anymore, looks like we can delete it

release/release.go
```diff
@@ -1151,8 +1151,6 @@ func linuxRelease(v version.Version) {
 							"--edition", me,
 							"--version", mv,
 							"--packages", packagesURL,
-							"--username", os.Getenv("BARQUE_USERNAME"),
-							"--password", os.Getenv("BARQUE_PASSWORD"),
```
rychipman, line 1155: can we still pass the api key via a flag? I believe the reason we were passing these as flags is that go subprocesses don't inherit the env by default

common.yml
Kimchelly: The username is actually still required because it verifies that the API key matches the user.

release/release.go
Kimchelly: Done.

PR #266, opened 2020-07-10
TOOLS-2531 Don't fail when there is no UUID in metadata and --preserveUUID is set

tfogo:
This is a small code change but I want to do a sanity check on the new behavior.

Previously `--preserveUUID` will abort whenever there is a metadata file without a UUID in the dump. This includes pre-3.6 dumps, any dump that includes a view, and any dump that includes a view dumped as a collection. 

I'm proposing that mongorestore will just restore any collection/view without a UUID even if `--preserveUUID` is specified. I don't see any reason why we would need to abort. @xdg is there anything I've missed?


mongorestore/restore.go
```diff
@@ -207,7 +207,7 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
 			indexes = metadata.Indexes
 			if restore.OutputOptions.PreserveUUID {
 				if metadata.UUID == "" {
-					return Result{Err: fmt.Errorf("--preserveUUID used but no UUID found in %v", intent.MetadataLocation)}
+					log.Logvf(log.Always, "--preserveUUID used but no UUID found in %v, generating new UUID for %v", intent.MetadataLocation, intent.Namespace())
```
sphanse99, line 210: Good idea to generate a UUID for the null case! However, where is the UUID generated since it would be set as null in line 212?

sphanse99: Moreover, we should probably add a test case that ensures that a UUID is correctly generated since this is not the default behavior (correct me if I'm wrong here) of mongorestore.

tfogo: Great question! The UUID for collections is generated by the server. 

If we have a zero-value UUID, then we create the collection with a `create` command. There's no way of passing a UUID to a `create` command, it will always generate a new one itself on MongoDB 3.6+. 

If we have a UUID value, then we use an `applyOps` command to create the collection. `applyOps` lets us bypass a lot of regular restrictions with commands so we can use it to manually set a UUID that the server has to use.

I think we don't need to check the UUIDs that eventually get created, since that's a server thing.

sphanse99: Ahh okay, this makes sense! In that case, LGTM :)

xdg:
No reason to abort on views.

PR #270, opened 2020-07-13
TOOLS-2622: Modified README Instructions to explain GOPATH and specif…

sphanse99:
…y supported go  version

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,13 @@ Report any bugs, improvements, or new feature requests at https://jira.mongodb.o
 
 Building Tools
 ---------------
-To build the tools, you need to have Go version 1.9 and up. `go get` will not work; you
-need to clone the repository to build it. Be sure to clone the repository into your Go workspace.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,8 +14,13 @@ Report any bugs, improvements, or new feature requests at https://jira.mongodb.o
 
 Building Tools
 ---------------
-To build the tools, you need to have Go version 1.9 and up. `go get` will not work; you
-need to clone the repository to build it. Be sure to clone the repository into your Go workspace.
+To build the tools, you need to have Go version 1.9 and up. Note, we only support Go version 1.12. `go get` will not
```
tfogo, line 17: We want to remove "To build the tools, you need to have Go version 1.9 and up." The reason we're making this change is because v1.9 no longer works and we don't want to make any promises about versions other than the one we currently use.

> we only support Go version 1.12

I think we can say something more broad here like "we currently build the tools with Go version 1.12, other Go versions may work but they are untested."

sphanse99: Definitely a better way of phrasing it. Will do :)

PR #106, opened 2017-11-06
TOOLS-1856 - use a memory pool in mongorestore

feliixx:
use a memory pool to avoid allocation for each
document to restore, see [TOOLS-1856](https://jira.mongodb.org/browse/TOOLS-1856) for details

PR #274, opened 2020-07-15
TOOLS-1856 Use a memory pool in mongorestore

feliixx:
Closes TOOLS-1856

( replaces https://github.com/mongodb/mongo-tools/pull/106 )

Not sure that it's still worth it though. Did some benchmarking with the following code: 

```
func BenchmarkRestore(b *testing.B) {

	opts, err := mongorestore.ParseOptions([]string{"/home/feliixx/dump/dump"}, "", "")

	restore, err := mongorestore.New(opts)
	if err != nil {
		b.Error(err)
	}

	b.StartTimer()

	for i := 0; i < b.N; i++ {
		b.StopTimer()
		restore.SessionProvider.DB("test").Drop(nil)
		b.StartTimer()

		restore.Restore()
	}
}
```
New code is a little bit faster, but most of the allocation ( > 90%) comes from this [line](https://github.com/mongodb/mongo-tools/blob/master/mongorestore/restore.go#L408) : 

`result.combineWith(NewResultFromBulkResult(bulk.InsertRaw(rawDoc)))`

so it doesn't make a huge difference for allocated bytes

### DIFF
--- a/mongorestore/restore.go
+++ b/mongorestore/restore.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/mongodb/mongo-tools-common/bsonutil"
@@ -371,7 +372,16 @@ func (restore *MongoRestore) RestoreCollectionToDB(dbName, colName string,
 
 	collection := session.Database(dbName).Collection(colName)
 
-	documentCount := int64(0)
+	pool := sync.Pool{
+		New: func() interface{} {
+			documents := make([]bson.Raw, restore.OutputOptions.BulkBufferSize)
+			for i := range documents {
+				documents[i] = make([]byte, 256)
+			}
+			return documents
+		},
+	}
+
 	watchProgressor := progress.NewCounter(fileSize)
 	if restore.ProgressManager != nil {
 		name := fmt.Sprintf("%v.%v", dbName, colName)
@@ -381,11 +391,15 @@ func (restore *MongoRestore) RestoreCollectionToDB(dbName, colName string,
 
 	maxInsertWorkers := restore.OutputOptions.NumInsertionWorkers
 
-	docChan := make(chan bson.Raw, insertBufferFactor)
+	docsBatchChan := make(chan []bson.Raw, insertBufferFactor)
 	resultChan := make(chan Result, maxInsertWorkers)
 
 	// stream documents for this collection on docChan
 	go func() {
+
+		count := 0
+		docsBatch := pool.Get().([]bson.Raw)
+
 		for {
 			doc := bsonSource.LoadNext()
 			if doc == nil {
@@ -395,16 +409,31 @@ func (restore *MongoRestore) RestoreCollectionToDB(dbName, colName string,
 			if restore.terminate {
 				log.Logvf(log.Always, "terminating read on %v.%v", dbName, colName)
 				termErr = util.ErrTerminated
-				close(docChan)
+				close(docsBatchChan)
 				return
 			}
 
-			rawBytes := make([]byte, len(doc))
-			copy(rawBytes, doc)
-			docChan <- bson.Raw(rawBytes)
-			documentCount++
+			if count == restore.OutputOptions.BulkBufferSize {
+				docsBatchChan <- docsBatch
+				count = 0
+				docsBatch = pool.Get().([]bson.Raw)
+			}
+
+			if len(doc) > cap(docsBatch[count]) {
+				docsBatch[count] = make([]byte, len(doc))
+			} else {
+				docsBatch[count] = docsBatch[count][0:len(doc)]
+			}
+
+			copy(docsBatch[count], doc)
+			count++
 		}
-		close(docChan)
+
+		if count > 0 {
+			docsBatchChan <- docsBatch[0:count]
+		}
+
+		close(docsBatchChan)
 	}()
 
 	log.Logvf(log.DebugLow, "using %v insertion workers", maxInsertWorkers)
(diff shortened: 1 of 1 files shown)


tfogo:
Thanks for rebasing this and thanks for the info on the performance. We're working on some benchmarks for mongorestore this week. Once those a done, we'll run them on this PR and share the results here. 

We can definitely optimize `result.combineWith(NewResultFromBulkResult(bulk.InsertRaw(rawDoc)))` if we need to. No reason for us to constantly allocate a new Result object.

hariamoor:
@feliixx I've made a [comment](https://github.com/mongodb/mongo-tools/pull/277#issuecomment-660387872) on #277 detailing the nature of my findings under the benchmarking that I did. The performance improvement that you reported definitely exists, but apparently only under sufficiently large workloads.

I also took the liberty to rerun the CI tests. Surprise -- they all work now 😉 

@tfogo I've followed up on `result.combineWith(...)` by filing [TOOLS-2665](https://jira.mongodb.org/browse/TOOLS-2665) -- I checked for a similar ticket from you first, but I didn't see one! Hope that's OK!

hariamoor:
@huan-Mongo @tfogo Thanks, guys!

Would it be alright if I followed up here on GitHub with the pprof traces and `go test -bench=.` output? I'd like to keep this PR updated for future reference, unless there's some reason I shouldn't.

huan-Mongo:
Sure, you can copy these files to this PR.

hariamoor:
The files are on #277 so that they don't get merged along with this

mongorestore/restore.go
```diff
@@ -23,7 +24,7 @@ import (
 	"go.mongodb.org/mongo-driver/mongo"
 )
 
-const insertBufferFactor = 16
+const insertBufferFactor = 3
```
rychipman, line 27: What's the rationale for this change?

mongorestore/restore.go
```diff
@@ -339,6 +340,11 @@ func fixDottedHashedIndex(index IndexDocument) {
 	}
 }
 
+type chunk struct {
```
rychipman, line 343: what's the rationale for using this `chunk` struct instead of just having a pool of `bson.Raw`s

tfogo: If I understand correctly, I think the reason is to insert the correct number of remaining documents in the last batch. We don't want to insert all 1000 array elements if only the first 50 were new elements. 

But, we can just take a new slice of the array instead. That would be a cleaner way of doing it and we could just use a pool of `[]bson.Raw`s.

rychipman: I agree that a pool of `[]bson.Raw` would be preferable to the `chunk` struct. At a higher level though, I'm wondering why we are batching at all. We weren't batching prior to this PR, and it doesn't seem that batching should be required to use a memory pool, as `pool.Get()` could return a `bson.Raw` instead of a `[]bson.Raw`.

I could imagine why batching might be beneficial (to reduce lock contention on `sync.Pool`, perhaps), but if we don't need to batch in order to get the perf improvements, I'd suggest we don't do it

tfogo: Ah I see what you're saying. I think doing batches is the correct thing to do. The reason is the `BufferedBulkInserter` buffers docs in batches before flushing. So we can't return documents to the pool one by one, we can only return batches of documents once the `BufferedBulkInserter` has flushed them. 

This does bring up an important point, the length of the `[]bson.Raw` in the pool should be `restore.OutputOptions.BulkBufferSize` to match the buffer size of the `BufferedBulkInserter`. Otherwise changing `--batchSize` would lead to errors.

rychipman: Makes sense -- SGTM!

feliixx: I think it was in order to reduce the number of object going through the `docChan` channel. It made a small difference for large number of documents (10M+ ? ) but I don't have the numbers anymore :/ I agree that it's a lot of extra complexity, so I switched the code back  `bson.Raw`

feliixx:
Hi @tfogo, @rychipman 

Thanks for the review, I switched the code back to using a `chan bson.Raw`, as I cannot remember for sure why I did it this way 3 years ago, and the hypothetical performance gain isn't worth the extra complexity
Let me know if there's more to be done !

tfogo:
Hi @feliixx, Thanks for the changes. I don't think that's quite right though. Using a `chan bson.Raw` will cause errors. With the current implementation, the following happens:

1. You get a single `bson.Raw` out of the pool. The underlying type of `bson.Raw` is just `[]byte`.  
2. You copy bytes into this `bson.Raw`, which writes to the underlying array of the slice. 
3. You pass it to the docChan.
4. It gets inserted to the `BufferedBulkInserter`. At this point, it has not been inserted into the database. A copy of the slice is in the `BufferedBulkInserter`'s buffer. It will only run the actual insert command when the buffer is full or when the flush method is run.
5. You return the `bson.Raw` back to the pool.

The problem with this is that the `bson.Raw` is returned to the pool before the `BufferedBulkInserter` has flushed its buffer. So the slices that are in the `BufferedBulkInserter`'s buffer are pointing to arrays which will get overwritten before they get inserted if a `bson.Raw` pointing to the same array gets taken out of the pool again. 

So you can only return `bson.Raw`s to the pool once you know they have been flushed by the `BufferedBulkInserter`. This means you need to batch together `bson.Raw`s in batches of the same size as the `BufferedBulkInserter`'s buffer. By default this size is 1000. But in general it is what is set by `restore.OutputOptions.BulkBufferSize`

So I think you were originally right that you needed to use a `[]bson.Raw`. But there are two changes that you need to make to the original. 

Firstly, the `[]bson.Raw` slice must be made with length `restore.OutputOptions.BulkBufferSize` since you can't know for sure that the default of 1000 will be used.

Secondly, you don't need to use the `chunk` type. A `[]bson.Raw` will do. Instead of keeping track of nbToInsert in a struct, just reslice the `[]bson.Raw` slice to the required length before passing it to the docChan if it needs to be less than the cap. If you do this I think you'll also need to check that any `[]bson.Raw` that you get from the pool has a length equal to the cap and reslice if it's not.

Does that make sense?

tfogo:
Hi @feliixx, just checking in here. Do my suggestions above sound okay?

feliixx:
Hi @tfogo, 

sorry for the late answer, I agree with your changes, I'll update the PR next week when I have some time !

tfogo:
Hi @feliixx, just checking in on this again. If you don't have time to make these changes, we're happy to make the changes necessary to your original PR.

feliixx:
Hi @tfogo, 

Sorry for the delay, I've updated the PR and rebased it on master, can you please take a look at it ?

tfogo (approved):
Hi @feliixx, sorry for the delay reviewing it. It looks great! Thanks for all your help with this - this is a fantastic improvement!

PR #277, opened 2020-07-15
TOOLS-1856: Use a memory pool in mongorestore (benchmark harness)

hariamoor:
This is an E2E benchmark harness for mongorestore. TOOLS-1856 is solved in #274 -- this is intended to help explore the effects of that optimization under different workloads.


hariamoor:
Some technical context:

This is currently in draft mode so that discussion from @huan-Mongo, @tfogo, et al. is possible. In its present state, we assume that the standalone BSON files used for benchmarking are in `benchmarkDataRootDir`.

I plan to write further setup code to populate the same with auto-generated test files, but in this stage, it's necessary to manually populate it, i.e. with the scripts included in [sqlproxy/189](https://github.com/10gen/sqlproxy/pull/189).

Right now, in [4c01cbc](https://github.com/mongodb/mongo-tools/pull/277/commits/4c01cbc8523696059d21809a34022c585a324834), there is non-functional benchmark code that needs feedback. The benchmarks should be _idempotent_, i.e. complete with automatic tear-down of the `restore`'d database and collections. Right now, all calls to restore but the first for each dataset fail because the current tear-down code is not working.

hariamoor:
Some possible follow-up tickets (out of scope of this PR):

**Best practices for setup/tear-down of E2E tests:** This [SO thread](https://stackoverflow.com/questions/42310088/setup-and-teardown-for-each-test-using-std-testing-package) specifies `(*testing.T).Cleanup()` is appropriate in Go >=1.14; however, it is not compatible with earlier versions of the toolchain. We might need a refactor ticket to improve these semantics when/if the Tools break compatibility with Go <1.14.

**Generalization of benchmark harness**: This PR, upon completion, should provide automatic generation of datasets of varying size and complexity, which is likely useful for testing and benchmarking across the Tools -- not just in `mongorestore`. It might make sense to include some of this in `mongo-tools-common/testutil/` so that we can minimize code duplication.

hariamoor:
I've completed the benchmarking process, as discussed.

I've provided a [small report](https://docs.google.com/document/d/1zeFzkQaXsIcGKAWak2ZR10hAB9h53o81KUPId-SqlNA/edit?usp=sharing) sharing my findings. In a nutshell, we seem to achieve the expected improvement of runtime performance, but only when the workload is sufficiently large -- in cases of smaller workloads, there is at best a moderate improvement.

I procured the test data from the scripts provided in [sqlproxy/189](https://github.com/10gen/sqlproxy/pull/189). I only used the ones that have workloads generalizable to the Tools -- some of them have BIC-specific context; however, I detail the purpose of each dataset in the supplied report.

@huan-Mongo @mattChiaravalloti @rychipman @tfogo I plan to schedule a meeting with you guys in order to discuss the results -- the biggest question is whether the observed behavior is significant enough to warrant the additional complexity of `sync.Pool`. I'm currently of the opinion that it is, but I could easily be convinced otherwise, if you guys feel that there's insufficient _marginal_ value provided by the change.

Looking forward to discussing this with you guys further!

tfogo:
@hariamoor, great stuff! Could you please share the pprof files on slack?

huan-Mongo:
Great job! Besides the pprof file @tfogo mentioned, can you also attach the go test run traces with the benchmark output? Thx!

hariamoor:
@tfogo I published the pprof traces I got, along with the benchmark output I'm getting from my local set up (as requested by @huan-Mongo), to [https://drive.google.com/drive/folders/1TAuTxtvCjbzqYRXq8bQFNw3dcoX6d-Qz?usp=sharing](a Google Drive directory). I thought this would be better than sharing it on Slack, since I handled pprof on a dataset-specific basis.

You should run something along the lines of the following to reproduce:

```bash
go tool pprof -png
                      -base=${OLD_TRACE}
                      ${NEW_TRACE}
```

If this is particularly confusing for anyone, I also have a lot of scripts I haven't yet pushed to make this convenient for myself -- reading those would perhaps help make my setup for this clearer.

I also pushed the traces I got in the last commit, along with a convenient script to create the benchmark files, in case anyone wants to pull this PR and have a look around.

tfogo:
Thanks @hariamoor! It looks like none of the datasets were over 1000 documents, is that correct? Since the sync.Pool contains 1000 documents, I think we won't be fully measuring the potential performance improvements from the pool unless we use over 1000 documents. I would be interested in seeing the results for datasets with 100,000 documents or more. But we should maybe talk with Huan and Ryan to see if that's even necessary. Based on your research so far, I think using sync.Pool looks good.

mongorestore/mongorestore_test.go
```diff
@@ -63,6 +68,59 @@ func getRestoreWithArgs(additionalArgs ...string) (*MongoRestore, error) {
...
+					collectionName,
+				}
+
+				restore, err := getRestoreWithArgs(args...)
+
+				if err != nil {
+					b.Fatalf("Failed to generate restore context: %v", err)
+				}
+
+				defer func() {
```
mattChiaravalloti, line 101: It doesn't seem necessary to put all of this in a `defer` block. You can simply put this code _after_ the two lines at the bottom of this function, i.e.:
```
...
b.ResetTimer()
restore.Restore()

// moved from deferred function
client, err := restore.SessionProvider.GetSession()
if err != nil {
    b.Fatalf("...")
}

err = client.Database(benchmarkDBName).Drop(context.Background())
if err != nil {
    b.Fatalf("...")
}
```

mongorestore/mongorestore_test.go
```diff
@@ -63,6 +68,59 @@ func getRestoreWithArgs(additionalArgs ...string) (*MongoRestore, error) {
...
+					client, err := restore.SessionProvider.GetSession()
+					defer restore.Close()
+
+					if err != nil {
```
mattChiaravalloti, line 105: [nit] You shouldn't put code between a line that assigns an error and the line that checks that error.

Along with my suggestion above, I advise moving the `defer restore.Close()` line higher up, anyway. Perhaps after the error check on line 97, i.e.:
```
...
restore, err := getRestoreWithArgs(args...)
if err != nil { // notice there is _no_ newline between the line assigning err and the conditional checking it
    b.Fatalf("...")
}

defer restore.Close()
...
```

mongorestore/mongorestore_test.go
```diff
@@ -63,6 +68,59 @@ func getRestoreWithArgs(additionalArgs ...string) (*MongoRestore, error) {
...
+						b.Fatalf("Could not gather context from MongoRestore for %v: %v", testName, err)
+					}
+
+					db := client.Database("benchmark")
```
mattChiaravalloti, line 109:
```suggestion
					db := client.Database(benchmarkDBName)
```
[nit] There is a variable for this; use it! (I implied this in my suggestion above)

mongorestore/testdata/tools-1856/js/d4.js
```diff
@@ -0,0 +1,15 @@
+// Dataset 7: Many (1,000) arrays
```
mattChiaravalloti, line 1:
```suggestion
// Dataset 4: Many (1,000) arrays
```
[nit] the comment doesn't match the filename

mongorestore/testdata/tools-1856/js/d5.js
```diff
@@ -0,0 +1,28 @@
+// Dataset 8: Deeply nested arrays (100 layers of nesting)
```
mattChiaravalloti, line 1:
```suggestion
// Dataset 5: Deeply nested arrays (100 layers of nesting)
```
[nit] the comment doesn't match the filename

mongorestore/testdata/tools-1856/js/d4.js
hariamoor: Yeah, these were an unfortunate accident -- thanks for noticing this!

hariamoor:
@tfogo Thanks a lot for the feedback! I'll respond to these points inline.

> It looks like none of the datasets were over 1000 documents, is that correct? Since the sync.Pool contains 1000 documents, I think we won't be fully measuring the potential performance improvements from the pool unless we use over 1000 documents.

That's true, but I tried a dataset of 1M and saw that it yielded no unique performance characteristics that were suitable for the report. With that said, a dataset of 10-20M would definitely be _interesting_, but at that point, it'd be a question of whether a benchmark over such a dataset is necessary to get #274 merged. If it is indeed necessary, I'd be happy to provide it -- just figured it'd be better to hold off on that. We can definitely discuss it in our meeting on this!

> I also think you should remove the profiles and traces from the PR. We don't need to check those in.

I was under the impression that this PR wouldn't get merged -- it'd just exist as part of the GitHub history for reference purposes. If that's not the case anymore, and we _should_ merge this, then I'd be happy to clean this up and get rid of those files.

mongorestore/mongorestore_test.go
```diff
@@ -63,6 +68,59 @@ func getRestoreWithArgs(additionalArgs ...string) (*MongoRestore, error) {
...
+					err = db.Drop(context.Background())
+
+					if err != nil {
+						b.Fatalf("Could not drop database for %v: %v", testName, err)
+					}
+				}()
+
+				b.ResetTimer()
+				restore.Restore()
```
huan-Mongo, line 118: The code in `b.Run` block above this line could be put out of block, since we are only interested in the benchmark in `Restore` function. We don't care about the session creation etc.

tfogo:
> but I tried a dataset of 1M and saw that it yielded no unique performance characteristics

Oh interesting. Seems counterintuitive to me. But what you have in the report now is fine for making a decision.

> I was under the impression that this PR wouldn't get merged

Oh I thought we were going to merge the benchmark function. I'm fine either way. @huan-Mongo what do you think?

hariamoor:
Closed in favor of #277

PR #279, opened 2020-07-17
Disable tasks tracked by build failure tix

rychipman:


tfogo:
I actually don't think we should be skipping TOOLS-2658 and TOOLS-2662. I think we should immediately put those into scheduled. TOOLS-2658 is related to a recent change so we should make sure that's working correctly. TOOLS-2662 more than just a flakey test - it's a whole version on Windows not getting tests. I think that's too large to ignore.

tfogo:
Or rather, if we do skip them, I think we should only do so if we immediately schedule work on them.

rychipman:
Agreed that those two are both more critical than some of the other ones that were disabled. I've moved them both into "Scheduled" and marked them with the `100.1.0` fixVersion.

PR #280, opened 2020-07-20
TOOLS-2494: Address mongorestore race condition

sphanse99:
**TLDR:** This solution avoids the panic as follows: if two restoring documents simultaneously return errors, they will both be logged. Since the termChan is no longer closed by this code, it cannot be closed twice.

**Detailed explanation:**
To fix the panic, instead of closing termChan when an error is encountered from the resultChan, we send an empty struct notifier across termChan. This behavior aligns with the intended behavior of termChan as on line 81 in `mongorestore.go`, the termChan should `// channel on which to notify if/when a termination signal is received`

This channel is also monitored by the go routine that streams the documents meant to be restored in `restore.go`. When termChan receives a signal, the termination error is logged, termErr is set (useful for other go routines as well), and document streaming is aborted.


mongorestore/restore.go
```diff
@@ -431,7 +431,7 @@ func (restore *MongoRestore) RestoreCollectionToDB(dbName, colName string,
 		totalResult.combineWith(<-resultChan)
 		if finalErr == nil && totalResult.Err != nil {
 			finalErr = totalResult.Err
-			close(restore.termChan)
+			restore.termChan <- struct{}{}
```
tfogo, line 434: Ah! I'm sorry, I suggested you do this but I just realized, this won't work. Multiple goroutines are listening to `restore.termChan`. If we send a single `struct{}{}`, only one of the goroutines will receive it. We do actually need to close the channel to ensure all the goroutines terminate.

So the correct thing to do is to check if the channel is already closed before closing it.

tfogo: > So the correct thing to do is to check if the channel is already closed before closing it.

Although this may end up being more complicated than necessary - there could be an easier way of doing this with some shared state and mutexes.

rychipman: Isn't checking the channel before closing it still racy? The channel could be closed by another goroutine after the check but before the call to close.

Another possibility (albeit a bit of a messy one) here would be to do the following:
- make `termChan` a buffered channel
- after a goroutine receives the signal on `termChan`, it writes that signal back onto the channel so that other goroutines can pick it up

tfogo: Yeah it's still racy. You'd have to use a mutex or a sync.Once to stop it being racy. 

I think using a channel makes it more complicated than it needs to be. We aren't using the channel for anything other than closing it. So we're essentially using it as a bool. I think the simplest thing to do would be to actually use a boolean value instead.

huan-Mongo:
A question is how do we test this change? Is it possible to create a test case for it?

mongorestore/restore.go
```diff
@@ -372,8 +372,9 @@ func (restore *MongoRestore) RestoreCollectionToDB(dbName, colName string,
 			if doc == nil {
 				break
 			}
-			select {
-			case <-restore.termChan:
+
+			switch restore.terminate {
+			case true:
```
mattChiaravalloti, line 377: [nit] might as well change this from a `switch` statement at this point, since we're checking only a single boolean value for `true`:

```
if restore.terminate {
    log.Logvf(...)
    termErr = ...
    close(...)
    return
}

rawBytes := ...
...
```

mongorestore/restore.go
sphanse99: Thanks for pointing this out! I was able to confirm that multiple go routines are listening to `restore.termChan` as the function `RestoreCollectionToDB` is called by `RestoreIntent` which is called by multiple go routines simultaneously. Thus, the most intuitive way to resolve this race condition is as Tim suggested -- change `restore.termChan` to a boolean ('restore.terminate'). Once `restore.terminate` is set, all go routines checking for the state of this variable will stop writing to their individual docChan's, finish reading the remaining data in docChan's buffer, and terminate with an error. `restore.terminate` is set in two cases -- (i) there is an error while attempting to restore a document or (ii) an interrupt signal is received. 

Since only one go routine attempts to close the local variable docChan per `RestoreCollectionToDB` thread, we will not encounter a double close panic here.

sphanse99:
> A question is how do we test this change? Is it possible to create a test case for it?

Unfortunately, there is not a good way to directly test this since we would want to confirm functionality of go routines inside `RestoreCollectionToDB`. Currently, `RestoreCollectionToDB` and other methods in `restore.go` are tested indirectly through the end-to-end tests, so I followed this design pattern. However, I am open to suggestions!

PR #282, opened 2020-07-22
TOOLS-2645: Check for duplicate index keys after converting legacy in…

huan-Mongo:
…dex definitions
This PR depends on https://github.com/mongodb/mongo-tools-common/pull/45/files . I will revendor after the MTC PR gets merged. 
Here I introduced flag fixDuplicatedLegacyIndexes. When enabled, after legacy indexes get converted, the indexes with duplicated index key will be skipped.


mongorestore/options.go
```diff
@@ -67,23 +67,24 @@ func (*InputOptions) Name() string {
...
+	NumParallelCollectionsOption     = "--numParallelCollections"
+	NumInsertionWorkersOption        = "--numInsertionWorkersPerCollection"
+	StopOnErrorOption                = "--stopOnError"
+	BypassDocumentValidationOption   = "--bypassDocumentValidation"
+	PreserveUUIDOption               = "--preserveUUID"
+	TempUsersCollOption              = "--tempUsersColl"
+	TempRolesCollOption              = "--tempRolesColl"
+	BulkBufferSizeOption             = "--batchSize"
+	FixDottedHashedIndexesOption     = "--fixDottedHashIndex"
+	FixDuplicatedLegacyIndexesOption = "--fixDuplicatedLegacyIndexes"
```
rychipman, line 87: I may be recalling an outdated discussion, but I thought when we last discussed we were not going to add a new flag and instead make it part of the default `--convertLegacyIndexes` behavior

mongorestore/restore.go
```diff
@@ -306,17 +306,27 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
 	return result
 }
 
-func (restore *MongoRestore) convertLegacyIndexes(indexes []IndexDocument, ns string) {
+func (restore *MongoRestore) convertLegacyIndexes(indexes []IndexDocument, ns string) []IndexDocument {
+	indexKeys := make(map[string]bool) // A set contains all the unique index's json strings
```
rychipman, line 310: small nit, but the go idiom is to use `map[string]struct{}` for sets

mongorestore/options.go
huan-Mongo: I may remember incorrectly, @tfogo Do you remembre if we would need a flag or not?

mongorestore/restore.go
huan-Mongo: Good idea!

mongorestore/options.go
tfogo: I think we agreed on not having a flag but adding a note to the documentation.

huan-Mongo: ah, cool. I will remove the flag then. Thanks!

mongorestore/mongorestore_test.go
```diff
@@ -770,6 +770,73 @@ func TestDeprecatedIndexOptions(t *testing.T) {
...
+	}
+	Convey("With a test MongoRestore", t, func() {
+		args := []string{
+			ConvertLegacyIndexesOption,
+		}
+
+		restore, err := getRestoreWithArgs(args...)
+		So(err, ShouldBeNil)
+
+		Convey("Creating index with invalid option and --convertLegacyIndexes should succeed", func() {
```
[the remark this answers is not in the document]
dragonlee8: Good catch! fixed.

mongorestore/restore.go
```diff
@@ -306,17 +306,28 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
+func (restore *MongoRestore) convertLegacyIndexes(indexes []IndexDocument, ns string) []IndexDocument {
+	indexKeys := make(map[string]struct{}) // A set contains all the unique index's json strings
+	var indexesConverted []IndexDocument
+	var exists = struct{}{}
 	for _, index := range indexes {
 		bsonutil.ConvertLegacyIndexKeys(index.Key, ns)
+		indexString := bsonutil.CreateExtJSONString(index.Key)
+		if _, ok := indexKeys[indexString]; ok {
+			// skip duplicated indexes
+			continue
```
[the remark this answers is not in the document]
huan-Mongo: Make sense, added the log

mongorestore/mongorestore_test.go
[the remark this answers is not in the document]
huan-Mongo: Good catch. fixed!

PR #289, opened 2020-07-31
TOOLS-2668: Create regex interface for getting files from remote FS in mongofiles

hariamoor:
This PR should enable the user to query GridFS by regex, i.e. PCRE, over filenames. The command should be as follows:

`mongofiles ... get_regex "<regex>" --regexOptions "<options>"`

`<options>` are optional additional arguments to the underlying _$regex_ query -- the possible options are [enumerated in the reference manual](https://docs.mongodb.com/manual/reference/operator/query/regex/#op._S_options).

**Test Plan**: Include two integration tests to check the `get_regex` command -- one without any options specified, and the other with the option for case-insensitivity.

Additionally, include two unittests to similarly check for cases of user input including and not including server options.

### DIFF
--- a/mongofiles/mongofiles.go
+++ b/mongofiles/mongofiles.go
@@ -273,15 +273,15 @@ func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
 	} else {
 		query = bson.M{"filename": mf.FileName}
 	}
-	
-	gridFiles, err = mf.findGFSFiles(query)
-        if err != nil {
-		return nil, err
-        }
 
-        if len(gridFiles) == 0 {
+	gridFiles, err = mf.findGFSFiles(query)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(gridFiles) == 0 {
 		return nil, fmt.Errorf("Could not find any files matching the query specified")
-        }
+	}
 
 	return gridFiles, err
 }
--- a/mongofiles/mongofiles.go
+++ b/mongofiles/mongofiles.go
@@ -33,7 +33,7 @@ const (
 	PutID    = "put_id"
 	Get      = "get"
 	GetID    = "get_id"
-        GetRegex = "get_regex"
+	GetRegex = "get_regex"
 	Delete   = "delete"
 	DeleteID = "delete_id"
 )
@@ -66,9 +66,9 @@ type MongoFiles struct {
 	// arguments in put and get commands
 	FileNameList []string
 
-        // Regular expression as supporting argument
-        // for get_regex
-        FileNameRegex string
+	// Regular expression as supporting argument
+	// for get_regex
+	FileNameRegex string
 
 	// GridFS bucket to operate on
 	bucket *gridfs.Bucket
@@ -130,12 +130,12 @@ func (mf *MongoFiles) ValidateCommand(args []string) error {
 		mf.FileNameList = args[1:]
 	case GetRegex:
 		// mongofiles get_regex ... should work over a PCRE
-                // and a string of options passed to the $regex query
-                if len(args) == 1 {
+		// and a string of options passed to the $regex query
+		if len(args) == 1 {
 			return fmt.Errorf("'%v' argument missing", args[0])
-                }
+		}
 
-                mf.FileNameRegex = args[1]
+		mf.FileNameRegex = args[1]
 	case Search, Delete:
 		if len(args) > 2 {
 			return fmt.Errorf("too many non-URI positional arguments (If you are trying to specify a connection string, it must begin with mongodb:// or mongodb+srv://)")
@@ -262,16 +262,21 @@ func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
 			return nil, fmt.Errorf("requested files not found: %v", mf.FileNameList)
 		}
 	} else if mf.FileNameRegex != "" {
-		query := bson.M{"filename": bson.M{"$regex": mf.FileNameRegex, "$options": mf.StorageOptions.RegexOptions}}
+		query := bson.M{
+			"filename": bson.M{
+				"$regex":   mf.FileNameRegex,
+				"$options": mf.StorageOptions.RegexOptions,
+			},
+		}
 
 		gridFiles, err = mf.findGFSFiles(query)
-                if err != nil {
+		if err != nil {
 			return nil, err
-                }
+		}
 
-                if len(gridFiles) == 0 {
+		if len(gridFiles) == 0 {
 			return nil, fmt.Errorf("files matching the following pattern were not found: %v", mf.FileNameRegex)
-                }
+		}
 	} else {
 		var queryProp string
 		var query string
--- a/mongofiles/mongofiles.go
+++ b/mongofiles/mongofiles.go
@@ -247,14 +247,14 @@ func (mf *MongoFiles) findGFSFiles(query bson.M) (files []*gfsFile, err error) {
 // Gets the GridFS file the options specify. Use this for the get family of commands.
 func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
 	var query bson.M
-	var expectedNumDocs int = 1
-	var expectedNumDocsError error
+	var minimumExpectedDocs int = 1
+	var minimumExpectedDocsError error
 
 	if len(mf.FileNameList) > 0 {
 		// Case supporting queries one or many files specified in mongofiles ... get ...
 		query = bson.M{"filename": bson.M{"$in": mf.FileNameList}}
-		expectedNumDocs = len(mf.FileNameList)
-		expectedNumDocsError = fmt.Errorf("requested files not found: %v", mf.FileNameList)
+		minimumExpectedDocs = len(mf.FileNameList)
+		minimumExpectedDocsError = fmt.Errorf("requested files not found: %v", mf.FileNameList)
 	} else if mf.FileNameRegex != "" {
 		// Case supporting queries by regex specified in mongofiles ... get_regex ...
 		query = bson.M{
@@ -263,10 +263,10 @@ func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
 				"$options": mf.StorageOptions.RegexOptions,
 			},
 		}
-		expectedNumDocsError = fmt.Errorf("files matching the following pattern were not found: %v", mf.FileNameRegex)
+		minimumExpectedDocsError = fmt.Errorf("files matching the following pattern were not found: %v", mf.FileNameRegex)
 	} else if mf.Id != "" {
 		// Case supporting queries by file ID specified in mongofiles ... get_id ...
-		expectedNumDocsError = fmt.Errorf("no such file with _id: %v", mf.Id)
+		minimumExpectedDocsError = fmt.Errorf("no such file with _id: %v", mf.Id)
 
 		id, err := mf.parseOrCreateID()
 		if err != nil {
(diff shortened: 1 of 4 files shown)


mongofiles/mongofiles.go
```diff
@@ -123,6 +128,14 @@ func (mf *MongoFiles) ValidateCommand(args []string) error {
 		}
 
 		mf.FileNameList = args[1:]
+	case GetRegex:
+		// mongofiles get_regex ... should work over a PCRE
+		// and a string of options passed to the $regex query
+		if len(args) == 1 {
```
huan-Mongo, line 134: maybe do checking `if len(args) == 1 || args[1] == "" ` like above?

mongofiles/mongofiles.go
```diff
@@ -248,6 +261,22 @@ func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
 		if len(gridFiles) < len(mf.FileNameList) {
 			return nil, fmt.Errorf("requested files not found: %v", mf.FileNameList)
 		}
+	} else if mf.FileNameRegex != "" {
+		query := bson.M{
+			"filename": bson.M{
+				"$regex":   mf.FileNameRegex,
+				"$options": mf.StorageOptions.RegexOptions,
```
huan-Mongo, line 268: What would the value be if user didn't provide options?

mongofiles/options.go
```diff
@@ -102,6 +103,10 @@ type StorageOptions struct {
 	// By default, mongofiles waits for a majority of members from the replica set to respond before returning.
 	// Cannot be used simultaneously with write concern options in a URI.
 	WriteConcern string `long:"writeConcern" value-name:"<write-concern>" default-mask:"-" description:"write concern options e.g. --writeConcern majority, --writeConcern '{w: 3, wtimeout: 500, fsync: true, j: true}'"`
+
+	// RegexOptions specifies the options passed to "$regex" queries that are used for get_regex
+	// The default is to use no options, i.e. standard PCRE syntax
+	RegexOptions string `long:"options" value-name:"<regex-options>" short:"o" description:"regex options used for get_regex"`
```
huan-Mongo, line 109: if this is a common option, which means it can be put anywhere in the command, such as `mongofiles --options i get_regex. ` Maybe we should rename it to some more unambiguous name such as `get-regex-options`?
Or we should check this options to be positioned after get_regex subcommand.

mongofiles/mongofiles.go
hariamoor: It would be the empty string, which translates on the server side to standard PCRE.

Case in point, a command without a specified pattern would be equivalent to `/<pattern>/`, which is just a regex search for "<pattern>".

This is indicated by MQL reference spec that I linked above.

mongofiles/options.go
hariamoor: I don't think either of those are necessary. The comments indicate that `StorageOptions` contains all the mongofiles-specific command-line options, being that that's also where `--local` is specified.

Like I mentioned, I followed the precedent set by the implementation of `get ... --local` for optional command-line flags specific to mongofiles. Unless I'm mistaken, that precedent indicates that no additional checks on our side are necessary.

mongofiles/mongofiles.go
hariamoor: Sure -- I can do this.

mongofiles/options.go
hariamoor: On second thought, I'd be fine with renaming it to `--regex-options` (even though I'm not sure if it's necessary), as long as it remains consistent with the implementation of `--local`.

mongofiles/mongofiles.go
huan-Mongo: If so, in the option definition, ```	RegexOptions string `long:"options" value-name:"<regex-options>" short:"o" description:"regex options used for get_regex"`
``` 
There should be a `default`empty value defined.

File: mongofiles/options.go (hunk 3 above)
Comment (huan-Mongo) [reply mongofiles/options.go:109]: local and options have a pretty big difference to the users. I am for renaming this field. But I'd like to hear others' thoughts.

Comment (tfogo) [reply mongofiles/options.go:109]: The commands use snake case but the options use camel case (see docs). So it should be something like: 

```
mongofiles get_regex <pattern> --regexOptions <options>
```

I don't know why it's different in commands vs options, but we should maintain consistency with what's already there.

Comment (hariamoor) [reply mongofiles/options.go:109]: This sounds good to me!

File: mongofiles/mongofiles.go (hunk 2 above)
Comment (hariamoor) [reply mongofiles/mongofiles.go:268]: Sounds good!

File: mongofiles/mongofiles.go [hunk 4]
```diff
@@ -248,6 +261,22 @@ func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
 		if len(gridFiles) < len(mf.FileNameList) {
 			return nil, fmt.Errorf("requested files not found: %v", mf.FileNameList)
 		}
+	} else if mf.FileNameRegex != "" {
```
mattChiaravalloti, line 264: This method is now quite large and contains effectively no code shared by the three branches of the conditional. I wonder if refactoring is in order here. I see three options:
1. Leave this as-is
2. Separate the three branches into three separate methods
3. Factor this method so the three branches share common code. It looks like each branch determines what the `query` is to argue to `mf.findGFSFiles` and each branch returns a unique error if `len(gridFiles)` is unexpected. Perhaps the three branches can set `query` and some error format string and then common code can run `mf.findGFSFiles(query)` and return an error if something is wrong.

mongofiles/options.go
```diff
@@ -28,6 +28,7 @@ Possible commands include:
 	put_id    - add a file with filename 'filename' and a given '_id'
 	get       - get files with filenames specified in the supporting arguments
 	get_id    - get a file with the given '_id'
+        get_regex - get files matching the supplied 'regex'
```
mattChiaravalloti, line 31:
```suggestion
	get_regex - get files matching the supplied 'regex'
```
[nit] tabs vs. spaces

mongofiles/mongofiles.go
hariamoor: I don't want to break this off and fragment the code structure, so I'd personally prefer (3).

I'm interested in thoughts from @huan-Mongo and @tfogo, however!

mongofiles/mongofiles.go
```diff
@@ -123,6 +128,14 @@ func (mf *MongoFiles) ValidateCommand(args []string) error {
...
 
 		mf.FileNameList = args[1:]
+	case GetRegex:
+		// mongofiles get_regex ... should work over a PCRE
+		// and a string of options passed to the $regex query
+		if len(args) == 1 || args[1] == "" {
+			return fmt.Errorf("'%v' argument missing", args[0])
+		}
+
+		mf.FileNameRegex = args[1]
```
tfogo, line 138: We also want a `if len(args) > 2` check. You can copy the check from List.

mongofiles/mongofiles.go
tfogo: (2) or (3) sounds good. I imagine we'll be able to get this done this week.

mongofiles/options_test.go
```diff
@@ -467,4 +482,29 @@ func TestPositionalArgumentParsing(t *testing.T) {
 			}
 		}
 	})
+
+	Convey("Testing 'get_regex' with '--options' should parse the regex and the options properly", t, func() {
```
tfogo, line 486: We generally don't test things like `--regexOptions` getting set to `mf.StorageOptions.RegexOptions`. We only test options parsing where we have specific code parsing the options. But `--regexOptions` is parsed entirely by the goflags library. 

I think it's fine to keep the test, but it shouldn't be in `TestPositionalArgumentParsing`. Pull it out into its own test function.

mongofiles/mongofiles.go
hariamoor: I'll move forward on (3) then, as long as Huan doesn't have any objections.

huan-Mongo: 3 Looks good to me!

mongofiles/options.go
hariamoor: Shouldn't this be handled by `gofmt` and # verified by the `format-go` and `lint-go` Evergreen tasks?

rychipman: No, it wouldn't. This indentation is part of a string variable, and so wouldn't be changed (or even inspected) `by gofmt`

mongofiles/options_test.go
hariamoor: Done.

mongofiles/mongofiles.go
hariamoor: Consider it done.

mongofiles/mongofiles.go
```diff
@@ -234,49 +249,38 @@ func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
...
-				return nil, err
-			}
-			gridFiles, err = mf.findGFSFiles(bson.M{"_id": id})
-			if err != nil {
-				return nil, err
-			}
-		} else {
-			queryProp = "name"
-			query = mf.FileName
+		query = bson.M{"filename": mf.FileName}
```
mattChiaravalloti, line 274: Does this case ever actually happen? I can't see any codepath that would get to this branch; `get_id` and `delete_id` will hit the `mf.Id != ""` branch, `get_regex` will hit the regex branch, and `get` will hit the `len(mf.FileNameList) > 0` branch.

mongofiles/mongofiles.go
```diff
@@ -234,49 +249,38 @@ func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
...
+	gridFiles, err = mf.findGFSFiles(query)
+	if err != nil {
+		return nil, err
+	}
 
-		if len(gridFiles) == 0 {
-			return nil, fmt.Errorf("no such file with %v: %v", queryProp, query)
-		}
+	if len(gridFiles) == 0 {
+		return nil, fmt.Errorf("Could not find any files matching the query specified")
```
mattChiaravalloti, line 283: We're now losing out on the detailed error messages from the previous version. Could we possibly assign an `errMsg` in each branch that maintains the old messages? As in,
```
	var gridFiles []*gfsFile
	var errMsg string
	var query bson.M

	if len(mf.FileNameList) > 0 {
		query = ...
		errMsg = fmt.Sprintf("requested files not found: %v", mf.FileNameList)
	} ...
	...
	if len(gridFiles) == 0 {
		return nil, errors.New(errMsg)
	}
	...
```
Something along those lines?

mongofiles/mongofiles.go
```diff
@@ -234,49 +249,38 @@ func (mf *MongoFiles) getTargetGFSFiles() ([]*gfsFile, error) {
...
-		}
+	gridFiles, err = mf.findGFSFiles(query)
+	if err != nil {
+		return nil, err
+	}
 
-		if len(gridFiles) == 0 {
-			return nil, fmt.Errorf("no such file with %v: %v", queryProp, query)
-		}
+	if len(gridFiles) == 0 {
```
mattChiaravalloti, line 282: This wasn't the condition for the first branch (`len(mf.FileNameList) > 0`). In that branch, we returned an error if `len(gridFiles) < len(mf.FileNameList)`. We can capture this in a general way by having this condition be `len(gridFiles) < someThreshold`. Perhaps we can default to `someThreshold := 1` (since that's the value for the latter two branches) and set `someThreshold = len(mf.FileNameList)` in the first branch?

It's up to you how to address this, the main thing is that this condition is different from the one before this change!

mongofiles/mongofiles.go
hariamoor: Yes, we need it for `mongofiles ... get ... --local ...`, it seems.

mongofiles/mongofiles.go
hariamoor: Consider it done.

mongofiles/mongofiles.go
hariamoor: Consider it done.

PR #291, opened 2020-08-07
TOOLS-2693: Fix race detector build failures

rychipman:
There were two distinct problems causing build failures in the race-detector variant.

One was that some of our automation code tries to determine the platform from the build variant name. Since "ubuntu-race" is not itself a platform, I had to add a way to override the variant name.

The other was that a number of our integration tests have expansions that cause them to _always_ run with TLS on, while the race detector buildvariant was configured not to use SSL. The easy fix here was to change the race detector variant to use SSL, though I also filed https://jira.mongodb.org/browse/TOOLS-2694 to address the underlying configuration inconsistency

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -1119,6 +1119,9 @@ pre:
           export EVG_IS_PATCH='${is_patch}'
           export EVG_BUILD_ID='${build_id}'
           export EVG_VARIANT='${build_variant}'
+          if [ '${_platform}' != '' ]; then
+            export EVG_VARIANT='${_platform}'
+          fi
           export EVG_USER='${evg_user}'
           export EVG_KEY='${evg_key}'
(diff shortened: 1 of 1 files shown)


tfogo:
- Why are qa-tests-4.0 failing? We shouldn't even be running qa-tests on this variant right? We're not doing the race detection on the qa-tests.
- I wonder if we should add -race to all tests similarly to how it's done in mirror. I opened TOOLS-2699 to discuss this.

rychipman:
@tfogo looks like we're all green now!

For the qa test failures, I had to add the `resmoke_use_ssl` expansion. After looking into it more thoroughly, I think the qa-tests _are_ actually running correctly with the race detector -- if a race is detected, the tool should exit with code `66`, which should in turn cause a qa-test to fail!

PR #295, opened 2020-08-14
TOOLS-2700: Use tag-triggered versions for releasing

rychipman:
This PR updates the release process in RELEASE.md to use evergreen's tag-triggered versions for releases. `common.yml` has been updated with tags for the tasks that should run on tag-triggered versions.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -47,26 +47,18 @@ Those changes will need to be made before we can do a new major release.
 At the time when the major release process is formalized, this section will be replaced with more specific instructions.
 
 #### Minor/Patch Release
-##### Determine the new version
-First, determine the version number of the new release.
-The new version should be the smallest possible minor or patch increment from the current version.
 
-For example, consider the version `<x>.<y>.<z>`.
-Bumping the patch version would give version `<x>.<y>.<z+1>`.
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -39,54 +39,6 @@ The only uncompleted ticket in the release should be the release ticket.
 If there are any remaining tickets that will not be included in this release, remove the fixVersion and assign them a new one if appropriate.
 Close the release on JIRA, adding the current date.
 
-### Releasing
```
rychipman, line 42: The git diff mangled this a bit, but what really happened here is that I moved the "Update the changelog" section into pre-release tasks instead of post-release tasks

tfogo: One issue with that is the "update the changelog" section says the following:

> - Update the release date to the date the release-json task finished on Evergreen in Eastern Time. You can set your timezone in "User Settings". 

I think it could be useful to have an accurate date for when the release actually became available for download. Do we want to edit this date post-release, or use the date when the tag was cut, or remove the date entirely?

rychipman: good point! moved it back to where it was in the process

rychipman:
@tfogo I've rebased this on the latest master and it's ready for another look!

tfogo (approved):
lgtm!

PR #299, opened 2020-08-19
Tools-2710: fix convertLegacyIndexes

huan-Mongo:
Fix the convertlegacyIndexes logic with the new helper function in MTC. Instead comparing index keys' Json string, the helper function will compare the index key and its value one by one, then skip the duplicated indexes. Also, any value not 0, not empty string or invalid input will be preserved with their original type. 

I will revendor the MTC version after https://github.com/mongodb/mongo-tools-common/pull/50 is merged.


tfogo (approved):
LGTM!

PR #300, opened 2020-08-22
TOOLS-2589: [v4.0]  add deprecation message for sslAllowInvalidHostnames and …

huan-Mongo:
…sslAllowInvalidCertificates

Evergreen patch build is here: https://evergreen.mongodb.com/version/5f57d8ac3627e066526bd172


common/options/options_test.go
```diff
@@ -422,3 +426,37 @@ func TestHiddenOptionsDefaults(t *testing.T) {
 	})
 
 }
+
+func TestDeprecationWarning(t *testing.T) {
+	if !(testtype.HasTestType(testtype.SSLTestType) && testtype.HasTestType(testtype.SSLTestType)) {
```
mattChiaravalloti, line 431: Why is this condition `testtype.HasTestType(testtype.SSLTestType)` duplicated here?

common/testutil/ssl_integration.go
```diff
@@ -15,8 +15,8 @@ func GetSSLOptions() commonOpts.SSL {
 	if testtype.HasTestType(testtype.SSLTestType) {
 		return commonOpts.SSL{
 			UseSSL:        true,
-			SSLCAFile:     "../common/db/openssl/testdata/ca.pem",
-			SSLPEMKeyFile: "../common/db/openssl/testdata/server.pem",
+			SSLCAFile:     "../common/db/openssl/testdata/ca-ia.pem",
+			SSLPEMKeyFile: "../common/db/openssl/testdata/test-client.pem",
```
mattChiaravalloti, line 19: I noticed that two `.pem` files were added: `test_server.pem` and `test_client.pem`. Why are two new files needed? Why did this line change from `.../server.pem` to `.../client.pem`? Does it matter?

tfogo:
FYI you can run the patch on the [mongo-tools-4.0](https://evergreen.mongodb.com/waterfall/mongo-tools-v4.0) project to compare against the 4.0-master branch. 

Looks like there are some qa-tests failing - are those flakes or something more serious?

common/options/options_test.go
huan-Mongo: ah, no idea why I have them twice. Thanks for pointing out!

common/testutil/ssl_integration.go
huan-Mongo: The old certificate expired when I did the test. So I copied the new one from master branch. 
Yes, the change here is needed. Since the client certificate used in the test was wrong, it should not be the same as the server certificate. But this problem was not shown since we had sslAllowInvalidCertificates in server startup cmd.

huan-Mongo:
@tfogo ah, OK. I run the patch against the current master instead. But I think it also works. I noticed the qa-test failure. Some are all caused by the expired ssl certificate as I mentioned above. Some are failed because of the extra warning messages printed out. There are a bunch of these tests. Do you think we need to fix them here?

tfogo:
Yeah I do think we need to fix them. We need to be fully green before merging to the 4.0 branch. We can cause server BFs if we're not fully green before adding tools to the server.

huan-Mongo:
@tfogo OK. I will try fix these.

huan-Mongo:
@tfogo I have fixed all the qa-tests and tests are run here https://evergreen.mongodb.com/version/5f57d8ac3627e066526bd172
Please take another look. thx!

tfogo (approved):
LGTM!

PR #301, opened 2020-08-25
TOOLS-2588: [v4.2] sslAllowInvalidHostnames bypass ssl/tls server certification validation entirely

huan-Mongo:
The patch build is https://evergreen.mongodb.com/version/5f4d10af0ae60606823c7bf3. Almost all tests passed with two failures, which is not related to the change.


mongoexport/options.go
```diff
@@ -33,7 +33,7 @@ type OutputFormatOptions struct {
 	Type string `long:"type" value-name:"<type>" default:"json" default-mask:"-" description:"the output format, either json or csv (defaults to 'json')"`
 
 	// Deprecated: allow legacy --csv option in place of --type=csv
-	CSVOutputType bool `long:"csv" default:"false" hidden:"true"`
```
huan-Mongo, line 36: With the newer version go-flags, the default for boolean doesn't work. Showing errors 
```
boolean flag `csv' may not have default values, they always default to `false' and can only be turned on
```
So I have to remove all the false boolean default value for flags.

tfogo (approved):
LGTM

rychipman:
LGTM

rychipman (approved):

PR #304, opened 2020-08-31
TOOLS-2716 Improve package dependency resolution with older versions

deriamis:
Upgrades from MongoDB v4.2 to v4.4 were somewhat problematic due to
incorrect dependency resolution resulting in both the older
mongodb-*-tools package and this package being installed at the same
time, causing file conflicts. The dependency resolution is improved so
that the package manager has an easier time doing an automated upgrade.


rychipman:
Approved previously in https://mongodbcr.appspot.com/655130001/

PR #305, opened 2020-09-02
TOOLS-2711: Tools fail if mongodb+srv URI or a legacy uri containing multiple mongos is specified

varsha804:


tfogo (approved):
It looks like this wasn't just a problem with mongoses, this was a problem with any seedlist. Updating the go driver in 100.1.0 caused this change in behavior. 

Previously we would force direct connections for seedlists without replica set names. I'm not sure this was the most sane default. I think we should research why this decision was made to see if it was needed. At the very least we may want to update the documentation. I opened TOOLS-2732 for the investigation.

This fix LGTM, though.

PR #308, opened 2020-09-10
TOOLS-2589: [v4.0]  add deprecation message for sslAllowInvalidHostna…

huan-Mongo:
…mes and … (#300

This is cherry-picked from commit https://github.com/mongodb/mongo-tools/commit/3bbfaa049a85f6e6721006f2c8d97722caf9e598


tfogo (approved):
lgtm!

tfogo:
Wait, actually I think there's a problem. v4.0 should be exactly in sync with v4.0-master. So we want to do a fast-forward merge, not a cherry-pick. Is there a reason you're doing a cherry-pick?

tfogo:
The cherry-picking mentioned in the vendoring docs is only for vX.Y-master.

huan-Mongo:
@tfogo I checked v4.2-master and v4.2, seems the last 3 commits have different hashes, so fast forward doesn't work. https://github.com/mongodb/mongo-tools/compare/v4.2...v4.2-master
If you think these two branch should be exact in sync, should I do a force push and correct it?

tfogo:
Ah yes, I remember we had this problem in February. Weird, I thought we had fixed this issue, but apparently not. 

Yes, we need to make them in sync. Ideally we want to keep the commit hashes from the v4.2 branch because we reference these hashes in commits that exist in the server. So in v4.2-master we want to get rid of `a99b0291a39` to `fd53ea4bf5`, and replace them with `11f19a0e1` to `1ca852c2d74` from v4.2. You might need to talk to @rychipman about getting permissions to force push.

huan-Mongo:
@rychipman Is it OK for me to force push to v4.2 from v4.2-master in order to keep them in sync as the discussion above?

rychipman:
We should keep them in sync, but you should only force push if you've already verified that the contents are the same (i.e. the output of `git diff` between the two is empty)

huan-Mongo:
Yes, I did a git diff and the code between these two are identical before my recent commit. So I will do a force push. Closing this  PR and reopen a new one with fast forward.

PR #312, opened 2020-09-11
TOOLS-2588: [v4.2] sslAllowInvalidHostnames bypass ssl/tls server cer…

huan-Mongo:
…tification validation entirely (#301)


tfogo (approved):
lgtm!

PR #314, opened 2020-09-15
TOOLS-2708: Atlas recommended connection string for mongostat doesn't work

varsha804:
The connection string failure seems to be a result of [this](https://github.com/mongodb/mongo-tools/commit/2aa1d7079b20817d17649ff7b43b2d79b69f1c8e) code change. Since `opts.Port` and `opts.Host` are not set when a user only uses `--uri`, `fullhost` gets set to `localhost:27017`, thereby incorrectly rewriting the hostname to `localhost`.

Manual testing:
1. Atlas cluster
2. mlaunch cluster
3. `--discover` without a connection string
4. Connection string as a positional argument without `--uri`
Screen Shot 2020-09-15 at 2 41 38 PM

### DIFF
--- a/mongostat/mongostat.go
+++ b/mongostat/mongostat.go
@@ -237,18 +237,15 @@ func NewNodeMonitor(opts options.ToolOptions, fullHost string) (*NodeMonitor, er
 	optsCopy.Connection.Host = host
 	optsCopy.Connection.Port = port
 	uriCopy := *opts.URI
-	optsCopy.URI = &uriCopy
-
-	if len(optsCopy.ConnString.Hosts) > 1 {
-		optsCopy.Direct = false
-	} else {
-		optsCopy.Direct = true
-		newCS, err := rewriteURI(uriCopy.ConnectionString, fullHost)
-		if err != nil {
-			return nil, err
-		}
-		uriCopy.ConnectionString = newCS
+	newCS, err := rewriteURI(uriCopy.ConnectionString, fullHost)
+	if err != nil {
+		return nil, err
 	}
+	uriCopy.ConnectionString = newCS
+	optsCopy.URI = &uriCopy
+	optsCopy.Direct = true
+	optsCopy.ConnString.Hosts = []string{fullHost}
+
 	sessionProvider, err := db.NewSessionProvider(optsCopy)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


mongostat/mongostat.go
```diff
@@ -243,8 +243,12 @@ func NewNodeMonitor(opts options.ToolOptions, fullHost string) (*NodeMonitor, er
 	}
 	uriCopy.ConnectionString = newCS
 	optsCopy.URI = &uriCopy
-	optsCopy.Direct = true
-	optsCopy.ConnString.Hosts = []string{fullHost}
```
huan-Mongo, line 247: Since we are not replacing the `Hosts` in ConnString with single node `fullHost`, the above code 
```	newCS, err := rewriteURI(uriCopy.ConnectionString, fullHost)
	if err != nil {
		return nil, err
	}
```
doesn't make sense in a replica-set connection anymore. I guess it didn't affect the tests, but it might have some side impact. 

I suggest we put the `rewriteURI` in the below if condition as well.

huan-Mongo (changes requested):

mongostat/mongostat.go
varsha804: Yeah, that's a good point!

tfogo (changes requested):
> Since `opts.Port` and `opts.Host` are not set when a user only uses `--uri`

I think this is the underlying bug though. What you have works, but I think it's more of a band-aid. I think we should fix the problem in `ToolOptions.setOptionsFromURI()` instead.

varsha804:
Updated manual testing:
Screen Shot 2020-09-23 at 2 46 08 PM

tfogo (changes requested):

vendor/github.com/mongodb/mongo-tools-common/options/options.go
```diff
@@ -596,6 +596,20 @@ func (opts *ToolOptions) setOptionsFromURI(cs connstring.ConnString) error {
 					return ConflictingArgsErrorFormat("host", strings.Join(cs.Hosts, ","), opts.Host, "--host")
 				}
 			}
+		} else if len(cs.Hosts) > 0 {
+			hostPort := strings.Split(cs.Hosts[0], ":")
+			opts.Host = hostPort[0]
```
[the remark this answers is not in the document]
tfogo: Oh I didn't see your question about this on the mtc PR - I'll respond there.

tfogo (approved):
lgtm once the mtc PR is merged.

huan-Mongo (approved):
LGTM after revendor

PR #322, opened 2020-09-30
TOOLS-2747 Get task lists from current build/version only

tfogo:
We needed to retrieve a single sign or dist task to push the releases. But GetTasksForRevision would get the tasks for both the commit and the git tag. This will PR ensure that we only get the tasks for the git tag.


tfogo:
I haven't been able to test this. My thought it we can push to master then run a test release of 100.2.0-<hash>. That will push to development repos so shouldn't dirty up our actual releases.

pmeredit (approved):
LGTM!

varsha804 (approved):
lgtm!

PR #323, opened 2020-10-09
TOOLS-2735: Move server vendoring instructions to a README in the repo

huan-Mongo:
I did many research and found Github doesn't support fast-forward merge which doesn't create new SHA. So I added a section of merging the vX.Y-master branch to vX.Y, which involves creating a PR for review and then manually merge.


tfogo (approved):
Nitpick: The blockquote (`>`) formatting is a bit unusual - I'd remove that styling. But lgtm.

PR #327, opened 2020-10-21
TOOLS-2707: Build mongo-tools with go 1.15

huan-Mongo:


tfogo (changes requested):
Looks like SUSE 15 is failing in the test because it can't find the 1.15 binaries. Has the build team fixed that yet? I restarted the test to check.

Also, can you include an update to the supported version in the readme?

tfogo:
Yeah SUSE 15 is still failing.

huan-Mongo:
Sure, I will add the supported version and added SUSE 15 build ticket to the dependency. I will first close this PR and reopen after the build ticket is finished.

huan-Mongo:
@tfogo I fixed SUSE15 and now all the tests passed except for those always been failing. Please take another look, thx!

tfogo (approved):
Nice! LGTM!

PR #332, opened 2020-11-10
TOOLS-2767: Windows 64 dist task fails

varsha804:
The macos 10.14 sign task is currently failing because Patrick changed his Apple password and forgot to change the corresponding evergreen variable -- should be fixed soon!


rychipman (approved):
LGTM

PR #334, opened 2020-11-17
TOOLS-2774: Raise default import/restore buffer size from 1000 to 100,000

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2774

This bumps up the `BulkBufferSize` to match what the server supports for versions 3.6+

### DIFF
--- a/mongoimport/mongoimport.go
+++ b/mongoimport/mongoimport.go
@@ -255,9 +255,8 @@ func (imp *MongoImport) validateSettings(args []string) error {
 	log.Logvf(log.DebugLow, "using %v decoding workers", imp.IngestOptions.NumDecodingWorkers)
 	log.Logvf(log.DebugLow, "using %v insert workers", imp.IngestOptions.NumInsertionWorkers)
 
-	// get the number of documents per batch
-	if imp.IngestOptions.BulkBufferSize <= 0 || imp.IngestOptions.BulkBufferSize > 1000 {
-		imp.IngestOptions.BulkBufferSize = 1000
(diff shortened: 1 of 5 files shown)


tfogo (changes requested):
We don't want to break support for versions <=3.4. Although not mentioned in docs, we do still internally support 3.4. So we should check the version to determine the value. Looks like we don't have any tests that hit the max buffer size (3.4 tests are still passing), so it would be good to add a test for this too.

edobranov:
Hmm, so I briefly thought about that and don't think we need to make any adjustments for versions <=3.4. I saw this under [the docs entry for `isMaster.maxWriteBatchSize`:](https://docs.mongodb.com/manual/reference/command/isMaster/#isMaster.maxWriteBatchSize)

> If a batch exceeds this limit, the client driver divides the batch into smaller groups each with counts less than or equal to the value of this field.

And it looks like that's indeed what happens in the Go driver [where they execute a batch op](https://github.com/mongodb/mongo-tools/blob/175e24a7371c944aa1880e3ada1d0151de244cc0/vendor/go.mongodb.org/mongo-driver/x/mongo/driver/operation.go#L321). So it should be handled for us and I don't think there's much (if any) risk for versions <=3.4, short of some overhead to perform the split, but that seems negligible.

Does that sound right to you? If you think it's still worthwhile to add a test for it though, I can toss one in.

tfogo:
Oh nice! Good research. LGTM then!

tfogo (approved):

tfogo:
Actually, hold on a second - I've got another concern - I'll slack you.

PR #336, opened 2020-11-24
TOOLS-2760: Remove obsoletes directive from rpm spec file

rychipman:
just a really quick fix that I picked up to help pare down the scheduled queue


tfogo (approved):
LGTM!

pmeredit (approved):
+1, but are we sure we want to remove the line entirely rather than just change <= to <?

rychipman:
@pmeredit yes, if you read the ticket comments, you'll see the rationale.

PR #340, opened 2021-01-27
TOOLS-2798: Add amazon2-arm64 platform

rychipman:


edobranov:
It looks like this variant went purple with this output:

```
running: rpmbuild -bb mongodb-database-tools.spec
rpmbuild
failed: ExitError: exit status 1. Stderr: "error: No compatible architectures found for build\n"
```

I poked around in some forums / the rpmbuild docs and it seems like a possible reason is that `mongodb-database-tools.spec` doesn't account for this architecture. Some people suggest omitting the `BuildArch` line which specifies the architecture(s), but I actually don't see that line in the spec file. We actually have `BuildArchitectures` which isn't in the rpmbuild docs afaict, so I'm not positive what the exact solution is.

Maybe some trial-and-error with adjusting that line?

edobranov (approved):
LGTM! Looks like it's passing now

PR #341, opened 2021-01-28
TOOLS-2795: Use system gcc instead of mongodbtoolchain gcc

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2795

Looks like all the variants are green, with a couple still scheduled: https://evergreen.mongodb.com/version/601307420305b9243b77846f

There's a system failure on Amazon Linux 64, though it's an unrelated permissions issue so that may be transient.

### DIFF
--- a/set_goenv.sh
+++ b/set_goenv.sh
@@ -38,7 +38,7 @@ set_goenv() {
         UNAME_M=$(PATH="/usr/bin:/bin" uname -m)
         case $UNAME_M in
             s390x)
-                export CC=/opt/mongodbtoolchain/v3/bin/s390x-mongodb-linux-gcc
+                export CC=/usr/bin/gcc
             ;;
             *)
                 # Not needed for other architectures
--- a/set_goenv.sh
+++ b/set_goenv.sh
@@ -16,8 +16,7 @@ set_goenv() {
         ;;
         *)
             PREF_GOROOT="/opt/golang/go1.15"
-            # XXX might not need mongodbtoolchain anymore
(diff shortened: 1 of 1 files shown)


set_goenv.sh
```diff
@@ -33,19 +32,11 @@ set_goenv() {
...
         UNAME_M=$(PATH="/usr/bin:/bin" uname -m)
         case $UNAME_M in
-            aarch64)
-                export CC=/opt/mongodbtoolchain/v3/bin/aarch64-mongodb-linux-gcc
-            ;;
-            ppc64le)
-                export CC=/opt/mongodbtoolchain/v3/bin/ppc64le-mongodb-linux-gcc
-            ;;
             s390x)
                 export CC=/opt/mongodbtoolchain/v3/bin/s390x-mongodb-linux-gcc
```
rychipman, line 41: out of curiosity, does it fix the s390x issue if you just change this to `gcc`?

rychipman (changes requested):

set_goenv.sh
edobranov: I think so: https://evergreen.mongodb.com/version/601323753e8e86520af94e4a

Hard to technically say for all the versions since it'll take a very long time for them to run, but all s390x's should be the same in this regard. Better to use `/usr/bin/gcc` then?

rychipman: If we can get rid of all the toolchain uses in one fell swoop, might as well!

edobranov: Sounds good! I added a new commit below if everything looks good to you.

rychipman (approved):
LGTM

PR #343, opened 2021-02-05
TOOLS-2790: Add 100.3.0 changelog

rychipman:


tfogo (approved):
Nit: We should probably have changed the TOOLS-2013 (Support PKCS8 encrypted client private keys) ticket to be a new feature. But it doesn't matter really.

LGTM!

PR #344, opened 2021-02-05
TOOLS-2801: Migrate from dep to Go modules and update README

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2801

This replaces `Gopkg.toml` / `Gopkg.lock` with `go.mod` / `go.sum`. I did this by running `go mod init` followed by `go mod tidy` and manually removing the Gopkg files. I also updated the README with some relevant info.

I'll revendor with `go mod vendor` later and push those changes separately, since the size of those changes exceed Evergreen's limits for patch builds.


README.md
```diff
@@ -15,9 +15,9 @@ Report any bugs, improvements, or new feature requests at https://jira.mongodb.o
 Building Tools
 ---------------
 
-We currently build the tools with Go version 1.15, other Go versions may work but they are untested. `go get` will not
-work; you need to clone the repository to build it. Be sure to clone the repository into your Go workspace inside your 
-$GOPATH.
+We currently build the tools with Go version 1.15. Other Go versions may work but they are untested.
+
+Using `go get` will not work. To build the tools, you need to first clone this repository into your Go workspace inside your $GOPATH.
```
edobranov, line 20: I originally thought that switching to Go modules will allow others to download and build the tools with `go get`, but that's not the case.

If you do `go get github.com/mongodb/mongo-tools`, that'll download the tools just fine, but it'll also run `go install` which won't work since we have multiple binaries to build. So users still have to use the make script or run `build.go` themselves.

mattChiaravalloti: That's too bad. Do you know what specific `go install` command it runs? When you say it "won't work" do you mean it does nothing or does it encounter an error?

Also, on a different note: since we've updated `mongo-tools` to modules, do we need to strictly clone the repo inside the `$GOPATH`?

README.md
```diff
@@ -30,14 +30,30 @@ git clone https://github.com/mongodb/mongo-tools
...
+In order to make changes to dependencies, you first need to change the `go.mod` file. You can manually edit that file to add/update/remove entries, or you can run the following in the repository directory:
+
+```
+go mod edit -require=<package>@<version>  # for adding or updating a dependency
+go mod edit -droprequire=<package>        # for removing a dependency
+```
+
+Then run `go mod vendor -v` to reconstruct the `vendor` directory to match the changed `go.mod` file.
+
+Optionally, run `go mod tidy -v` to ensure that the `go.mod` file matches the vendored code.
```
mattChiaravalloti, line 56: Is that really what `tidy` does? I thought `go mod tidy` removed unused dependencies. Saying it ensures the `go.mod` file matches the vendored code seems misleading.

README.md
edobranov: Specifically for building the tools — we haven't made the tools "go gettable" yet, but [judging from the docs](https://golang.org/cmd/go/#hdr-Add_dependencies_to_current_module_and_install_them), I believe `go get github.com/mongodb/mongo-tools` will download the repo's contents to $GOPATH/src and then run a plain `go install` from the repo's directory:
> With no package arguments and also without -u, 'go get' is not much more than 'go install'

I tried doing a plain `go install` / `go get` from the mongo-tools root directory and found that both basically make the build file into an executable named `mongo-tools`, and put it in $GOPATH/bin like so:

```
$ $GOPATH/bin/mongo-tools
USAGE: [tasks ...] [options ...]

TASKS:
  build(tools)
        build the tools
  ---------------------------------------------------------------------
  test -> [test:unit test:integration test:kerberos]
  test:integration(tools, ssl, auth, kerberos, topology)
        runs integration tests
...

$ $GOPATH/bin/mongo-tools build
START  | build
       | rm: bin/bsondump
       | exec: '/usr/local/bin/go run release/release.go get-version'
...
```

To summarize our options, I think these will all be equivalent ways to build the tools:
- `git clone github.com/mongodb/mongo-tools` + `./make build` from the cloned directory
- `go get github.com/mongodb/mongo-tools` + `$GOPATH/bin/mongo-tools build` from the $GOPATH/src/github.com/mongodb/mongo-tools directory
- `go get -d github.com/mongodb/mongo-tools` + `./make build` from the $GOPATH/src/github.com/mongodb/mongo-tools directory

So `go get` will technically work, but it seems kind of impractical to me to use it as a cloning mechanism for building the tools. It might be better suited to use in projects that need `mongo-tools` as a dependency.

If this all sounds right to you, do you think it's worth mentioning in the README?

edobranov: And to your second q — you're right that we won't need to clone inside the $GOPATH anymore. One of the many upsides of Go modules :)

I'll remove that phrase from the README.

README.md
edobranov: Ah, I pulled the phrasing from [the Go docs entry for go mod](https://golang.org/ref/mod#go-mod-tidy) (first sentence) but I misinterpreted it. Thanks for catching this.

It does remove unused dependencies, but it could also add new ones. Would phrasing it like this seem alright? Feel free to suggest something else too.
> Optionally, run `go mod tidy -v` to ensure that the `go.mod` file matches the `mongo-tools` source code.

go.mod
```diff
@@ -0,0 +1,29 @@
+module github.com/mongodb/mongo-tools
+
+go 1.14
```
rychipman, line 3: The README.md says `go 1.15`, but this says `go 1.14`. Which one is correct?

go.mod
```diff
@@ -0,0 +1,29 @@
...
+	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d // indirect
+	github.com/mongodb/mongo-tools-common v4.0.18+incompatible
+	github.com/nsf/termbox-go v0.0.0-20160718140619-0723e7c3d0a3
+	github.com/smartystreets/assertions v0.0.0-20160205033931-287b4346dc4e // indirect
+	github.com/smartystreets/goconvey v1.6.1-0.20160205033552-bf58a9a12912
+	github.com/xdg/stringprep v1.0.1-0.20180714160509-73f8eece6fdc // indirect
+	github.com/youmark/pkcs8 v0.0.0-20181117223130-1be2e3e5546d // indirect
+	go.mongodb.org/mongo-driver v1.4.2
+	golang.org/x/sys v0.0.0-20200302150141-5c8b2ff67527 // indirect
+	gopkg.in/mgo.v2 v2.0.0-00010101000000-000000000000
```
rychipman, line 24: it looks like it might be an accident that we're still depending on mgo. can you file a ticket to get rid of it?

rychipman (changes requested):
what did you end up deciding re: how you wanted to handle the vendoring for this PR?

go.mod
edobranov: It's `go 1.15`, thanks for catching that. I thought I was using that version locally but I have `go 1.14`, which is why the go.mod file defaulted to that.

go.mod
edobranov: It looks like there was only one mgo import in `mongostat/mongostat_test.go`. Removing that and running `go mod tidy` also removes the entry for mgo.

I think we can include the change here if that sounds ok to you.

linter.go
```diff
@@ -0,0 +1,13 @@
+// +build linter
+
+package main
+
+import _ "github.com/3rf/mongo-lint/golint"
```
edobranov, line 5: I think this file is necessary to retain a vendored third-party tool that we use in Evergreen. Without this file, `go mod tidy` will remove it.

Some more explanation [here](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) on why it seems to be the best way to do it.

common.yml
```diff
@@ -815,6 +815,8 @@ functions:
           set -v
           set -e
           mkdir -p bin
+          ${_set_shell_env}
+          go mod vendor -v
```
edobranov, line 819: > what did you end up deciding re: how you wanted to handle the vendoring for this PR?

@rychipman, I took @huan-Mongo's suggestion to temporarily revendor on Evergreen to ensure that static checks and compilation pass.

I'll remove these two lines before merging and subsequently push the vendored changes directly to master.

README.md
mattChiaravalloti: > If this all sounds right to you, do you think it's worth mentioning in the README?

Nah, not worth it! Thanks for the detailed explanation.


> And to your second q — you're right that we won't need to clone inside the $GOPATH anymore. One of the many upsides of Go modules :)
>
> I'll remove that phrase from the README.

Great

README.md
mattChiaravalloti: Yes that is better phrasing.

linter.go
mattChiaravalloti: Seems reasonable since it's following the golang convention.

edobranov (commented):
I made some iterative changes, though there's one issue we can't really fix at the moment.

Since `go mod` prunes all files that aren't directly relevant to the mongo-tools Go code, it'll also prune any PEM files in mtc. This makes it impossible for mtc's `testutil` package to function properly, so most (all?) of the mongo-tools tests will fail for the moment.

We can fix this by making an adjusted branch in mtc, or by migrating mtc as a subpackage to tools. @rychipman suggested directly doing the latter and letting some of the tools tests fail temporarily.

mattChiaravalloti (approved):
The core changes here look good.

Letting some tests fail for now sounds good to me since you talked it out with Ryan 👍

rychipman (approved):
LGTM

PR #345, opened 2021-02-08
TOOLS-704: Remove system.indexes collection dumping from mongodump

varsha804:
All of the tests passed after removing `system.indexes` dumping.


tfogo (approved):
LGTM!

PR #347, opened 2021-02-11
TOOLS-2803: [v3.6] Add --config option for password values

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2803

This backports the `--config` option and unit tests to v3.6. Just one thing to point out below.


common/options/options.go
```diff
@@ -455,6 +466,91 @@ func (o *ToolOptions) ParseArgs(args []string) ([]string, error) {
...
+	}
+	err = yaml.UnmarshalStrict(configBytes, &config)
+	if err != nil {
+		return errors.Wrapf(err, "error parsing config file %s", opts.General.ConfigPath)
+	}
+
+	// Assign each parsed value to its respective ToolOptions field.
+	opts.Auth.Password = config.Password
+	opts.URI.ConnectionString = config.ConnectionString
+	opts.SSL.SSLPEMKeyPassword = config.SSLPEMKeyPassword
```
edobranov, line 549: Even though v3.6 normally requires you to build with the `ssl` tag to enable the `--sslPEMKeyPassword` option, the config file would actually let you specify the password regardless.

That shouldn't break anything since the code checks for a `sslPEMKeyFile` first, so I didn't think it's necessary to check whether the tools are built with the `ssl` tag here.

rychipman (approved):
LGTM! A couple process suggestions for the remaining steps in the backport:
- If you haven't already, maybe just do a quick manual test of both ssl and non-ssl builds to make sure everything works as expected, especially with the sslPEMKeyPassword option
- When you merge into v3.6, let's make sure that the tests pass on the waterfall for that commit before we submit the mongo revendor commit to the commit queue

PR #348, opened 2021-02-16
TOOLS-2805: Add mod tidy static analysis check for Go modules

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2805

This adds mongohouse's `go mod tidy` check to the tools. Building should now look like:
```
$ ./make build

START  | sa:modtidy
       | exec: '/usr/local/bin/go mod tidy'
FINISH | sa:modtidy in 123.32094ms
START  | build
       | rm: bin/bsondump
       | exec: '/usr/local/bin/go run release/release.go get-version'
       | exec: '/usr/bin/git rev-parse HEAD'
       ...
```

### DIFF
--- a/build.go
+++ b/build.go
@@ -12,8 +12,13 @@ import (
 var taskRegistry = task.NewRegistry(task.WithAutoNamespaces(true))
 
 func init() {
+	// Build
+	taskRegistry.Declare("build").Description("build the tools").OptionalArgs("tools").Do(buildscript.BuildTools)
+
+	// Static Analysis
 	taskRegistry.Declare("sa:modtidy").Description("runs go mod tidy").Do(buildscript.SAModTidy)
-	taskRegistry.Declare("build").Description("build the tools").DependsOn("sa:modtidy").OptionalArgs("tools").Do(buildscript.BuildTools)
+
+	// Testing
 	taskRegistry.Declare("test:unit").Description("runs unit tests").OptionalArgs("tools").Do(buildscript.TestUnit)
 	taskRegistry.Declare("test:integration").Description("runs integration tests").OptionalArgs("tools", "ssl", "auth", "kerberos", "topology").Do(buildscript.TestIntegration)
 	taskRegistry.Declare("test:kerberos").Description("runs kerberos tests").Do(buildscript.TestKerberos)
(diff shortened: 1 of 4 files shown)


build.go
```diff
@@ -12,7 +12,8 @@ import (
 var taskRegistry = task.NewRegistry(task.WithAutoNamespaces(true))
 
 func init() {
-	taskRegistry.Declare("build").Description("build the tools").OptionalArgs("tools").Do(buildscript.BuildTools)
+	taskRegistry.Declare("sa:modtidy").Description("runs go mod tidy").Do(buildscript.SAModTidy)
+	taskRegistry.Declare("build").Description("build the tools").DependsOn("sa:modtidy").OptionalArgs("tools").Do(buildscript.BuildTools)
```
rychipman, line 16: instead of making the `build` task depend on this one, let's instead run it in a new `modtidy` task in the Static Analysis buildvariant

buildscript/build.go
```diff
@@ -25,6 +27,45 @@ var toolNames = []string{
 	"mongofiles",
 }
 
+// SAModTidy runs go mod tidy and ensure no changes were made.
+// Copied from mongohouse: https://github.com/10gen/mongohouse/blob/333308814f96a0909c8125f71af7748b263e3263/buildscript/sa.go#L72
+func SAModTidy(ctx *task.Context) error {
```
rychipman, line 32: since we're aiming to eventually move more of our static analysis stuff into go code, perhaps we can put this into a separate `sa.go` file instead of just bundling it here?

rychipman (changes requested):

common.yml
```diff
@@ -1319,6 +1319,13 @@ tasks:
       vars:
         test_path: "test/legacy42"
 
+- name: mod-tidy
+  commands:
+    - func: "fetch source"
+    - func: "run make target"
+      vars:
+        target: sa:modtidy
```
edobranov, line 1327: Do you think it's worth renaming "run make target" to something else like "run target", since it's now the entry point for building, testing, and static analysis?

edobranov (commented):
Cool, should be working by the looks of it: https://evergreen.mongodb.com/version/602c450b2fbabe1305c2c479

common.yml
rychipman: I don't think so; I don't think there's anything about having a static analysis task that makes it more or less of a "make" target.

rychipman (approved):
LGTM

PR #349, opened 2021-02-17
TOOLS-2783: Revert "TOOLS-1856: Add a memory pool in mongorestore"

rychipman:
This reverts commit 83e6dd5115e82ed8bb51428a0d78b12b5b66dac0.

It appears that the memory pool has cause signifcantly increased memory usage for a number of users.
I'm proposing that we revert it ASAP, and then re-open TOOLS-1856 with a note that we will need a solution for the memory usage issue if we want to add the memory pool back in.

### DIFF
--- a/mongorestore/restore.go
+++ b/mongorestore/restore.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"strings"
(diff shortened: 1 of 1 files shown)


tfogo (approved):
LGTM

PR #350, opened 2021-02-22
TOOLS-2796 Fix mongotop_sharded.js test

tfogo:
It looks like the test was checking that only the error message was printed. When we added some warning messages, they got printed too and the test started failing. The test now doesn't care if anything else is printed.


varsha804 (approved):
nice, lgtm!

PR #351, opened 2021-02-22
TOOLS-2791: Update release process

rychipman:
This ticket updates `RELEASE.md` in advance of the next release to account for a couple things:
- We should use the docs changes needed field on the ticket instead of manually filing downstream docsp tix
- Downstream CLOUDP ticket creation will have been automated (see PROG-4705).


tfogo (approved):
LGTM!

PR #352, opened 2021-02-23
TOOLS-2584: Restoring single BSON file should use db set in URI

varsha804:
I changed all usages of `NSOptions.{DB, Collection}` to `ToolOptions.Namespace.{DB, Collection}`, and added a test to `mongorestore/options_test.go`

### DIFF
--- a/mongorestore/options.go
+++ b/mongorestore/options.go
@@ -128,8 +128,6 @@ const (
 
 // NSOptions defines the set of options for configuring involved namespaces
 type NSOptions struct {
-	DB                         string   `short:"d" long:"db" value-name:"<database-name>" description:"database to use when restoring from a BSON file"`
-	Collection                 string   `short:"c" long:"collection" value-name:"<collection-name>" description:"collection to use when restoring from a BSON file"`
 	ExcludedCollections        []string `long:"excludeCollection" value-name:"<collection-name>" description:"DEPRECATED; collection to skip over during restore (may be specified multiple times to exclude additional collections)"`
 	ExcludedCollectionPrefixes []string `long:"excludeCollectionsWithPrefix" value-name:"<collection-prefix>" description:"DEPRECATED; collections to skip over during restore that have the given prefix (may be specified multiple times to exclude additional prefixes)"`
 	NSExclude                  []string `long:"nsExclude" value-name:"<namespace-pattern>" description:"exclude matching namespaces"`
@@ -166,12 +164,6 @@ func ParseOptions(rawArgs []string, versionStr, gitCommit string) (Options, erro
 			"provide only one polling interval in seconds and only one MongoDB connection string. " +
 			"Connection strings must begin with mongodb:// or mongodb+srv:// schemes",
 		)
-	}
-
-	// Allow the db connector to fall back onto the current database when no
-	// auth database is given.
-	if opts.DB == "" {
-		opts.Namespace = &options.Namespace{DB: nsOpts.DB, Collection: nsOpts.Collection}
 	}
 
 	log.SetVerbosity(opts.Verbosity)
--- a/mongorestore/options.go
+++ b/mongorestore/options.go
@@ -146,7 +146,7 @@ func (*NSOptions) Name() string {
 // ParseOptions reads the command line arguments and converts them into options used to configure a MongoRestore instance
 func ParseOptions(rawArgs []string, versionStr, gitCommit string) (Options, error) {
 	opts := options.New("mongorestore", versionStr, gitCommit, Usage, true,
-		options.EnabledOptions{Auth: true, Connection: true, URI: true})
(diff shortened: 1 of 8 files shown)


tfogo (changes requested):
In order to make this work, we have to remove `DB` and `Collection` from the `NSOptions` struct. We would also need to add `Namespace: true` to the `EnabledOptions` struct we pass to `options.New()`. The reason is currently the `NSOptions` are added as a group to the `ToolOptions.parser` but the `ToolOptions.Namespace` struct is not. So `-d` and `-c` do not set any values in `ToolOptions.Namespace` (though the URI database will).

Unfortunately, if we do this we won't get the custom help text for `-d` and `-c` in mongorestore. It would be nice if you could find a way to keep the custom help text too. It might be easier to achieve this by keeping mongorestore using `NSOptions` and do some sort of normalization of the two options. For example adding more logic to where we currently do this in `ParseOptions()`:

```
opts.Namespace = &options.Namespace{DB: nsOpts.DB}
```

mongorestore/options.go
```diff
@@ -146,7 +146,7 @@ func (*NSOptions) Name() string {
 // ParseOptions reads the command line arguments and converts them into options used to configure a MongoRestore instance
 func ParseOptions(rawArgs []string, versionStr, gitCommit string) (Options, error) {
 	opts := options.New("mongorestore", versionStr, gitCommit, Usage, true,
-		options.EnabledOptions{Auth: true, Connection: true, URI: true})
+		options.EnabledOptions{Auth: true, Connection: true, Namespace: true, URI: true})
```
varsha804, line 149: ^

mongorestore/options.go
```diff
@@ -169,8 +169,10 @@ func ParseOptions(rawArgs []string, versionStr, gitCommit string) (Options, erro
 	}
 
 	// Allow the db connector to fall back onto the current database when no
-	// auth database is given; the standard -d/-c options go into nsOpts now
-	opts.Namespace = &options.Namespace{DB: nsOpts.DB}
+	// auth database is given.
+	if opts.DB == "" {
+		opts.Namespace = &options.Namespace{DB: nsOpts.DB, Collection: nsOpts.Collection}
+	}
```
varsha804, line 175: ^

varsha804:
@tfogo Oh, I actually did add `Namespace: true` to `EnabledOptions, and I modified `ParseOptions()` like so:
```
if opts.DB == "" {
		opts.Namespace = &options.Namespace{DB: nsOpts.DB, Collection: nsOpts.Collection}
	}
```
I kept the `DB` and `Collection` fields in `NSOptions` so I could make that change.

Sorry if that wasn't clear; I probably should've pointed out those changes with some comments! Will do so now.

mongorestore/options.go
varsha804, line 175: ^

mongorestore/options.go
varsha804, line 149: ^

tfogo:
@varsha804, I'm so sorry I totally messed up here. I 100% should have seen those changes - it's absolutely my fault. I think I confused myself when I was looking at my local repo and I thought I had your changes checked out but I did not.

However there's still a problem that we have duplicates of the `--db` and `--collection` options. From the output of `--help`:

```
namespace options:
  -d, --db=<database-name>                                  database to use
  -c, --collection=<collection-name>                        collection to use

uri options:
      --uri=mongodb-uri                                     mongodb uri connection string

namespace options:
  -d, --db=<database-name>                                  database to use when restoring from a BSON file
  -c, --collection=<collection-name>                        collection to use when restoring from a BSON file
      --excludeCollection=<collection-name>                 DEPRECATED; collection to skip over during restore (may be specified multiple times to exclude additional collections)
      --excludeCollectionsWithPrefix=<collection-prefix>    DEPRECATED; collections to skip over during restore that have the given prefix (may be specified multiple times to exclude additional prefixes)
      --nsExclude=<namespace-pattern>                       exclude matching namespaces
      --nsInclude=<namespace-pattern>                       include matching namespaces
      --nsFrom=<namespace-pattern>                          rename matching namespaces, must have matching nsTo
      --nsTo=<namespace-pattern>                            rename matched namespaces, must have matching nsFrom
```

And I'm not entirely sure how the `--db` and `--collection` options will get parsed into the data structures. I think the simplest solution here is to remove `DB` and `Collection` from the `NSOptions` struct. That will mean we'll lose the custom help text for `--db` and `--collection` for mongorestore. Could you also open a ticket in the CLI options rationalization epic about adding those custom help descriptions back in? I expect we'll need to restructure the options logic to make it possible.

tfogo (changes requested):
(see comment above)

varsha804:
@tfogo Oh, no worries! Easy mistake to make 😊

tfogo (approved):
LGTM!

PR #353, opened 2021-02-25
TOOLS-2806: Add mongo-tools/common unit tests

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2806

This adds coverage for mtc's unit tests to the existing `build.go` files. It'll also make it easier to add mtc's integration and auth tests.

Evergreen patch of all variants that run `unit`: https://evergreen.mongodb.com/version/60380a1ad6d80a572f2d02e0

A few comments below.


common/failpoint/failpoint_test.go
```diff
@@ -12,37 +12,37 @@ import (
...
-		So(Enabled("foo"), ShouldBeTrue)
-		So(Enabled("baz"), ShouldBeTrue)
-		So(Enabled("biz"), ShouldBeTrue)
-		So(Enabled(""), ShouldBeTrue)
-		So(Enabled("bar"), ShouldBeFalse)
+		c.So(Enabled("foo"), c.ShouldBeTrue)
+		c.So(Enabled("baz"), c.ShouldBeTrue)
+		c.So(Enabled("biz"), c.ShouldBeTrue)
+		c.So(Enabled(""), c.ShouldBeTrue)
+		c.So(Enabled("bar"), c.ShouldBeFalse)
```
edobranov, line 29: There's a collision between the `Reset()` func declared in our `failpoint` package here and the `convey` package that we use for tests.

I'm not sure how that wasn't an issue before (maybe it had to do with the `// +build failpoints` build constraint at the top), but we should specify the `convey` funcs with an alias for this test file anyway.

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d // indirect
 	github.com/nsf/termbox-go v0.0.0-20160718140619-0723e7c3d0a3
 	github.com/pkg/errors v0.9.1
-	github.com/smartystreets/assertions v0.0.0-20160205033931-287b4346dc4e // indirect
+	github.com/smartystreets/assertions v0.0.0-20190401211740-f487f9de1cd3 // indirect
```
edobranov, line 19: Had to update this for some assertions related to `time.Time` to work.

common/db/testdata/test-client-pkcs8-encrypted-demo.pem
```diff
@@ -1,91 +0,0 @@
------BEGIN CERTIFICATE-----
-MIIGYTCCBEmgAwIBAgIJAL+8WDUncZFZMA0GCSqGSIb3DQEBCwUAMIGaMQswCQYD
```
edobranov, line 2: Accidentally added this file a while back for some personal testing; removing it now.

build.go
```diff
@@ -18,10 +18,13 @@ func init() {
...
-	taskRegistry.Declare("test:unit").Description("runs unit tests").OptionalArgs("tools").Do(buildscript.TestUnit)
-	taskRegistry.Declare("test:integration").Description("runs integration tests").OptionalArgs("tools", "ssl", "auth", "kerberos", "topology").Do(buildscript.TestIntegration)
-	taskRegistry.Declare("test:kerberos").Description("runs kerberos tests").Do(buildscript.TestKerberos)
+	// Tools Testing
+	taskRegistry.Declare("test:tools.unit").Description("runs tools unit tests").OptionalArgs("tools").Do(buildscript.TestToolsUnit)
+	taskRegistry.Declare("test:tools.integration").Description("runs tools integration tests").OptionalArgs("tools", "ssl", "auth", "kerberos", "topology").Do(buildscript.TestToolsIntegration)
+	taskRegistry.Declare("test:tools.kerberos").Description("runs tools kerberos tests").Do(buildscript.TestToolsKerberos)
+
+	// Tools Common Testing
+	taskRegistry.Declare("test:common.unit").Description("runs common unit tests").OptionalArgs("tools").Do(buildscript.TestCommonUnit)
```
edobranov, line 27: I thought it'd be good to separate the tests into two camps of `test:tools.<type>` and `test:common.<type>` since there are so many `common` tests. Open to suggestions for other naming schemes though.

rychipman: Instead of creating a whole separate task tree and task functions for common, I think we should just keep using the existing `test:unit`, `test:integration`, etc, and just change the `-tools` flag to something like `-pkg` that can be any of the top-level packages (i.e. one of the tool names or `common`)

go.mod
edobranov: Sure, that should work. The way I just had it (with `assertions` being 3 years newer than `goconvey`) is the way mtc has it set up.

I'll update to the most recent version of `goconvey` (1.6.4), which looks to actually get rid of the pinned indirect dependencies in `go.mod`.

buildscript/build.go
```diff
@@ -123,12 +131,20 @@ func runToolTests(ctx *task.Context, tool string, testType string) error {
...
 	}
 
-	args := []string{"test"}
+	// If we're running the ./common tests, use the recursive
+	// wildcard (...) to run the tests for all subpackages there.
+	// Otherwise, run the tests for the specified tool.
+	args := []string{"test", "./" + dir}
+	if dir == "common" {
+		args[1] = args[1] + "/..."
+	}
```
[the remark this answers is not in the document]
edobranov: I originally thought it'd try to test things in the `vendor` directory, but I realized that tests don't get vendored by `go mod`. I'll just use `./...` in conjunction with adding the `-pkg` flag.

common/txn/testdata/oplog_entries.json
```diff
@@ -0,0 +1,372 @@
+{
```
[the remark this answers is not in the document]
edobranov: Ah right, this is data that the `common/txn/main_test.go` file relies on. We gitignore `*.json` files by default, so I force-added this one for the txn tests to work.

edobranov:
Evergreen w/ unit tests: https://evergreen.mongodb.com/version/60394cf656234339d89e2ea0

It looks like there might be an inconsistent test (race condition maybe?) with the progress bar: https://evergreen.mongodb.com/test_log/60394e5da4cf4710352c316d#L1553
I'll look into it a bit now, but might just file a ticket if I can't find the issue.

rychipman (changes requested):
code updates look good, but seems like there are quite a few tests failing!

edobranov:
> code updates look good, but seems like there are quite a few tests failing!

Whoops, I had the integration tests for `common` running by accident. We can skip them for now and re-enable them in TOOLS-2807.

rychipman (approved):
LGTM

PR #354, opened 2021-03-01
TOOLS-2807: Add common integration and srv tests

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2807

This enables integration tests for `common`, including one options test with test type `SRVConnectionStringTestType`.

I think there's already sufficient platform coverage in mongo-tools for integration tests and that we don't need to sift through mtc's setup, but I can take a closer look if you think otherwise.

### DIFF
--- a/buildscript/build.go
+++ b/buildscript/build.go
@@ -51,11 +51,11 @@ func TestKerberos(ctx *task.Context) error {
 
 // TestIntegration is an Executor that runs all integration tests for the provided packages.
 func TestIntegration(ctx *task.Context) error {
-	err := runTests(ctx, selectedPkgs(ctx), testtype.IntegrationTestType)
-	if err != nil {
-		return err
-	}
+	return runTests(ctx, selectedPkgs(ctx), testtype.IntegrationTestType)
+}
 
+// TestSRV is an Executor that runs the SRV connection string test for package `common`.
+func TestSRV(ctx *task.Context) error {
 	return runTests(ctx, []string{"common"}, testtype.SRVConnectionStringTestType)
 }
 
--- a/buildscript/build.go
+++ b/buildscript/build.go
@@ -54,9 +54,9 @@ func TestIntegration(ctx *task.Context) error {
 	return runTests(ctx, selectedPkgs(ctx), testtype.IntegrationTestType)
 }
 
-// TestSRV is an Executor that runs the SRV connection string test for package `common`.
+// TestSRV is an Executor that runs all SRV tests for the provided packages.
 func TestSRV(ctx *task.Context) error {
-	return runTests(ctx, []string{"common"}, testtype.SRVConnectionStringTestType)
+	return runTests(ctx, selectedPkgs(ctx), testtype.SRVConnectionStringTestType)
 }
 
 // buildToolBinary builds the tool with the specified name, putting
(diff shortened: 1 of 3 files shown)


buildscript/build.go
```diff
@@ -51,18 +51,12 @@ func TestKerberos(ctx *task.Context) error {
...
-			pkgs = append(pkgs[:i], pkgs[i+1:]...)
-			break
-		}
+	err := runTests(ctx, selectedPkgs(ctx), testtype.IntegrationTestType)
+	if err != nil {
+		return err
 	}
-	return runTests(ctx, pkgs, testtype.IntegrationTestType)
+
+	return runTests(ctx, []string{"common"}, testtype.SRVConnectionStringTestType)
```
rychipman, line 59: I'm not sure these should be grouped in with the rest of the integration tests. I think these are similar to the Kerberos or AWS tests, in that they have different MongoDB setup/connection requirements. As such, I think that we should probably run them in a separate target instead of bundling them with the integration tests

rychipman (changes requested):

buildscript/build.go
edobranov: Yea, I was thinking about this a bit. So this testtype only applies to one test (`TestOptionsParsingForSRV`) and this is the relevant setup code from mtc's Evergreen config:
```
if [ "${INT_TEST}" = "true" ]; then
    export TOOLS_TESTING_INTEGRATION="true"
    export TOOLS_TESTING_SRV="true"
    export ATLAS_URI=${atlas_srv_uri}
fi;
```
Based on that, the srv test should have identical req's as any integration test on Evergreen. And now that we always export `ATLAS_URI`, we could theoretically just change the testtype of `TestOptionsParsingForSRV` to an integration test.

Side note though -- this test only works locally if you're connected to the vpn (probably stemming from some dns lookup that the driver's connstring package does). That may be why there's two testtypes in the first place.

Still think it's worth it to separate it, or should we just make it an integration test? I'm okay either way.

rychipman: I think the local testing use-case is the main reason why we'd want to keep them as separate targets in `build.go` -- not only would you have to be on the VPN, but you'd also have to have the `ATLAS_URI` set, which I think is a secret expansion from the evg config.

buildscript/build.go
```diff
@@ -51,18 +51,12 @@ func TestKerberos(ctx *task.Context) error {
...
-			break
-		}
-	}
-	return runTests(ctx, pkgs, testtype.IntegrationTestType)
+	return runTests(ctx, selectedPkgs(ctx), testtype.IntegrationTestType)
+}
+
+// TestSRV is an Executor that runs the SRV connection string test for package `common`.
+func TestSRV(ctx *task.Context) error {
+	return runTests(ctx, []string{"common"}, testtype.SRVConnectionStringTestType)
```
rychipman, line 59:
```suggestion
	return runTests(ctx, selectedPkgs(ctx), testtype.SRVConnectionStringTestType)
```
I don't think we should hard-code that common is the only package with srv tests. This is the case right now, but a future refactoring could _theoretically_ change this. I think that maintaining consistency with the other test targets will minimize surprise for folks working on this in the future

rychipman (changes requested):

rychipman (approved):
LGTM

PR #355, opened 2021-03-03
TOOLS-2717: Add suse15-s390x platform

rychipman:


edobranov (approved):
Lgtm! The dist task might take a while in the queue, so I'm okay with pushing the changes ahead of the task.

rychipman:
closing this, since it turns out that this platform is _not_ actually going to be supported by the server
Sorry for the confusion!

PR #357, opened 2021-03-09
TOOLS-2808: Add `common` AWS IAM auth tests

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2808

Patch with `aws-auth` tasks: https://evergreen.mongodb.com/version/6047a0c757e85a11db4b27c0

This enables the existing `TestAuthConnection` test and allows for easy addition of future tests with the `AWSAuthTestType`.

A couple comments below on some changes I had to make when migrating this from mtc's script-based testing approach.


common.yml
```diff
@@ -904,6 +912,91 @@ functions:
...
+            "iam_auth_assume_aws_secret_access_key": "${iam_auth_assume_aws_secret_access_key}",
+            "iam_auth_assume_role_name": "${iam_auth_assume_role_name}"
+          }
+          EOF
+
+  "setup-aws-auth-test-with-assume-role-credentials":
+    - command: shell.exec
+      type: test
+      params:
+        silent: true
```
edobranov, line 938: I'm keeping these functions as silent like in mtc since they contain private vars, but in general I don't think it matters that much.

common.yml
```diff
@@ -904,6 +912,91 @@ functions:
...
+        script: |
+          set -x
+          set -v
+          set -e
+
+          # The aws_e2e_assume_role script requires python3 with boto3.
+          # Set up or use an existing python virtualenv for boto3.
+          venv='venv'
+          venvNonWindows () { . "$venv"/bin/activate; python3 -m pip install boto3; }
+          venvWindows () { . "$venv"/Scripts/activate; pip install boto3; }
```
edobranov, line 949: I copied this section from sqlproxy's approach to python venvs for Windows vs. Non-Windows. We were actually installing `boto3` globally on Windows machines in the mtc script.

common/db/db_test.go
```diff
@@ -407,18 +408,25 @@ func TestAuthConnection(t *testing.T) {
...
 		enabled := options.EnabledOptions{URI: true}
+
 		var uri string
 		if testtype.HasTestType(testtype.AWSAuthTestType) {
-			uri = os.Getenv("MONGOD")
+			uriBytes, err := ioutil.ReadFile("../lib/MONGOD_URI")
+			if err != nil {
+				panic("Could not read MONGOD_URI file")
+			}
+			uri = string(uriBytes)
```
edobranov, line 418: I had some trouble with this part since I didn't immediately realize that env vars aren't retained across Evergreen `shell.exec` blocks.

Reading a file produced by the Evergreen script seems like the easiest way to get the connection string here.

common/lib/aws_assume_role.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
```
rychipman, line 1: can we move the `common/lib/` dir to `common/testdata/lib/`, so that it's guaranteed to be ignored by the go tooling?

common.yml
```diff
@@ -92,6 +92,7 @@ mongo_tools_variables:
       - name: integration-4.4-auth
       - name: integration-4.4-cluster
       - name: kerberos
+      - name: aws-auth
```
rychipman, line 95: when looking at [this task in the patchbuild|https://spruce.mongodb.com/task/mongo_tools_ubuntu1604_aws_auth_patch_db970fef5a038c6f47e8d765066f7a342858cecc_6047a313850e615d98d694bc_21_03_09_16_32_21/tests?execution=0], I noticed that there were a bunch of tests running in it that aren't actually AWS IAM tests.

It looks like this is because they don't call `SkipUnlessTestType` -- could we add the appropriate `SkipUnlessTestType` call to those tests so they don't clutter up the results for all the test tasks?

rychipman (changes requested):

common.yml
edobranov: Sure thing, they all look like unit tests to me.

rychipman (approved):
LGTM

PR #358, opened 2021-03-10
TOOLS-2041 Build indexes after oplogReplay

tfogo:
TOOLS-2041 and TOOLS-176 raise issues with oplog replay and unique indexes. There are multiple related situations where oplog replay can fail due to a duplicate key error. 

The most common situations can be solved by switching the order of building indexes and replaying oplog entries. The one thing this does not solve is a class of errors associated with restoring into a pre-existing collection. For this class of errors I believe that failing when a duplicate index key error occurs during oplog replay is the correct behavior. In fact, I think in a future major version update we should consider disallowing the use of `--oplogReplay` without `--drop`.

Before this PR, the order of operations was:

1. Copy data
2. Build indexes
3. Replay oplog

This PR changes the order to:

1. Copy data
2. Replay oplog
3. Build indexes

There are three main scenarios which can cause a duplicate key error during oplog replay. Imagine two documents that contain a duplicate index key. The scenarios are:

1. The collection dump contains both documents.
2. The collection dump contains one document and the oplog inserts the other.
3. The oplog inserts both documents.

## Scenario 1

**The dump documents:**
```
{ _id: 1, a: 1 }
{ _id: 2, a: 1 }
```

**The dump oplog:**
```
DELETE { _id: 1 }
INSERT { _id: 2, a: 1 }
```

**The dump indexes:**
```
{ key: {a: 1}, unique: true }
```

**What happens:**
We copy the collection, then the index build will fail due to duplicate index keys.

## Scenario 2

**The dump documents:**
```
{ _id: 2, a: 1 }
```

**The dump oplog:**
```
INSERT { _id: 1, a: 1 }
DELETE { _id: 1 }
INSERT { _id: 2, a: 1 }
```

**The dump indexes:**
```
{ key: {a: 1}, unique: true }
```

**What happens:**
We copy the collection and build the index. Then the oplog replay will fail due to duplicate key error.

## Scenario 3

**The dump documents:**
```
{ _id: 3, a: 2 }
```

**The dump oplog:**
```
INSERT { _id: 1, a: 1 }
INSERT { _id: 2, a: 1 }
DROP collection
CREATE collection
CREATE INDEX { key: {a: 1}, unique: true }
INSERT { _id: 3, a: 2 }
```

**The dump indexes:**
```
{ key: {a: 1}, unique: true }
```

**What happens:**
The collection is copied and the index is built. Then a duplicate key error will occur during the oplog replay when we try to insert `{ _id: 2, a: 1 }`.

## _id index

Scenarios 1 and 2 can occur with the _id index only if `--forceTableScan` is used with MMAPv1. But this will not cause an error and the collection will be consistent after the restore. Scenario 1 could error with the _id index if `--stopOnError` is set, but I think this should be expected.

Scenario 3 can occur with the _id index if there is a pathological situation involving a collection where `autoIndexId=false`. In this case also, there will be no error and the data will be consistent after restore. This is due to oplog entry inserts replacing documents that have the same _id. 

## Solution

All three scenarios are solved by switching the order of index builds and the oplog replay.

## Restoring into a pre-existing collection

Switching the index builds and the oplog replay will not solve the problem of restoring into a collection that already exists. In this case, unique index constraints could exist on the collection before the restore even starts. There could also be pre-existing data in the target collection. 

I think this use case is fairly dubious. The point of `--oplogReplay` is to provide a point-in-time restore. If there's pre-existing data, you cannot guarantee the state after the restore is consistent. You would have a mix of documents from different times. This is a problem even if there aren't any duplicate key errors. 

If we were to ignore duplicate key errors during oplog application and skip those ops, it would get even worse - it would be possible to have documents which aren't even internally consistent. Different fields will be from different times. 

So I think it's correct that duplicate key errors during `--oplogReplay` do cause mongomirror to fail. If you want to restore into a pre-existing collection, it's best not to use `--oplogReplay`. In fact, I think in a future major version update we should consider disallowing the use of `--oplogReplay` without `--drop`.

It was suggested to use a different `oplogApplicationMode` for `applyOps` in order to ignore duplicate index key errors. But there is not an `oplogApplicationMode` that does this. During initial sync the server can ignore a duplicate index key error of indexes other than _id, but the only reason that's okay is because a) we know the collection will eventually be made consistent, and b) we know the syncing node cannot accept reads or writes. There are no such guarantees when using mongorestore to restore to a pre-existing collection. So even if we had an `oplogApplicationMode` that could ignore a duplicate index key errors, we would not want to use it.


mongorestore/mongorestore.go
```diff
@@ -327,6 +327,7 @@ func (restore *MongoRestore) ParseAndValidateOptions() error {
 
 // Restore runs the mongorestore program.
 func (restore *MongoRestore) Restore() Result {
+	defer restore.SessionProvider.Close()
```
rychipman, line 330: What's the rationale for closing the SessionProvider when we call `Restore()`? Given that it's not opened/created in `Restore()`, this seems like surprising behavior to me

mongorestore/metadata.go
```diff
@@ -36,17 +36,16 @@ type authVersionPair struct {
...
 type Metadata struct {
-	Options        bson.D          `bson:"options,omitempty"`
-	Indexes        []IndexDocument `bson:"indexes"`
-	UUID           string          `bson:"uuid"`
-	CollectionName string          `bson:"collectionName"`
+	Options        bson.D                  `bson:"options,omitempty"`
+	Indexes        []intents.IndexDocument `bson:"indexes"`
+	UUID           string                  `bson:"uuid"`
+	CollectionName string                  `bson:"collectionName"`
+	DatabaseName   string                  `bson:",omitempty"`
```
rychipman, line 43: where does this `DatabaseName` field get used? C-f isn't finding it for me in this patch

common/intents/intent.go
```diff
@@ -41,6 +41,13 @@ type FileNeedsIOBuffer interface {
 	ReleaseIOBuffer()
 }
 
+// IndexDocument holds information about a collection's index.
+type IndexDocument struct {
```
rychipman, line 45: it feels a little odd to have this in a package called `intents`, especially when it's imported from elsewhere. Perhaps we could put this into `common/db` or into a new package like `common/metadata` or `common/indexes`?

mongorestore/restore.go
```diff
@@ -169,90 +313,50 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
...
-		if metadata != nil {
-			options = metadata.Options
-			indexes = metadata.Indexes
-			if restore.OutputOptions.PreserveUUID {
-				if metadata.UUID == "" {
-					log.Logvf(log.Always, "--preserveUUID used but no UUID found in %v, generating new UUID for %v", intent.MetadataLocation, intent.Namespace())
-				}
-				uuid = metadata.UUID
-			}
+	log.Logvf(log.Always, "options D: %#v\n", options)
```
rychipman, line 327: is this intended to be a debug statement?

mongorestore/restore.go
```diff
@@ -157,7 +301,7 @@ func (restore *MongoRestore) RestoreIntent(intent *intents.Intent) Result {
 			if strings.HasPrefix(intent.C, "system.") {
 				log.Logvf(log.Always, "cannot drop system collection %v, skipping", intent.Namespace())
 			} else {
-				log.Logvf(log.Info, "dropping collection %v before restoring", intent.Namespace())
+				log.Logvf(log.Always, "dropping collection %v before restoring", intent.Namespace())
```
rychipman, line 304: any particular reason this was bumped up to `log.Always`?

mongorestore/restore.go
```diff
@@ -69,6 +69,150 @@ func NewResultFromBulkResult(result *mongo.BulkWriteResult, err error) Result {
 	return Result{nSuccess, nFailure, err}
 }
 
+func (restore *MongoRestore) RestoreIndexesForIntents() error {
```
rychipman, line 72: are there any key things to review in this file? It looks like a lot of it is refactoring, but it's hard to tell from the diff

rychipman (changes requested):
some initial questions. also looks like there are a couple evg tasks failing

mongorestore/mongorestore.go
tfogo: When we run the tests, we only run `restore.Restore()` so the sessions don't get closed like they usually would if `main()` is the entry point. This was causing issues for me with testing locally since the mongod was flooded with connections and crashed. I suppose the more correct approach would be to close the connections properly in each test.

mongorestore/metadata.go
tfogo: Ah yeah, at some point during development I was using it but not anymore.

mongorestore/mongorestore_test.go
```diff
@@ -736,7 +732,7 @@ func TestDeprecatedIndexOptions(t *testing.T) {
 			restore.TargetDirectory = "testdata/indextestdump"
 			result := restore.Restore()
 			So(result.Err, ShouldNotBeNil)
-			So(result.Err.Error(), ShouldStartWith, `indextest.test_collection: error creating indexes for indextest.test_collection: createIndex error:`)
+			So(result.Err.Error(), ShouldStartWith, `restore error: error creating indexes for indextest.test_collection: createIndex error:`)
```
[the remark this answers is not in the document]
tfogo: Yes it does, good point!

mongorestore/restore.go
tfogo: Yeah, I thought it made sense for this message in particular to always be logged. I think it gives good information about what restore is doing, and it's good to be explicit when taking a destructive action.

There's a larger question here about how we make the choice about what verbosity level messages should use.

mongorestore/oplog.go
```diff
@@ -250,7 +250,8 @@ Loop:
 // ApplyOps is a wrapper for the applyOps database command, we pass in
 // a session to avoid opening a new connection for a few inserts at a time.
 func (restore *MongoRestore) ApplyOps(session *mongo.Client, entries []interface{}) error {
-	singleRes := session.Database("admin").RunCommand(nil, bson.D{{"applyOps", entries}})
+	log.Logvf(log.DebugHigh, "Applying entries: %+v\n", entries)
```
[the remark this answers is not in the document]
tfogo: Oh yeah, this was just from testing. Removed it.

common/intents/intent.go
tfogo: Yeah good point. I moved it to `common/db`.

mongorestore/restore.go
tfogo: Not sure I can narrow it down. I'd love some eyes on the refactoring itself to double check it's correct. Previously reading metadata files, copying data, and index builds all happened in `RestoreIntent()`. Now we cycle through the intents 3 times and do each task separately. Copying data and index builds are parallelized but reading metadata files is single threaded.

huan-Mongo (changes requested):
This change looks good to me in general. I just have a little concern with moving the build index to the end of the restoration when there is drop index command in the oplog replay. Because in Mongodump, the metadata was created prior to the oplog dump. 

For Example, if we have the following scenaior: 
Assume at the beginning of the dump, the collection contains an index `{ key: {a: 1}, unique: true }`. So in the metadata, the index exists. 

During the dump, user dropped the index so there is a `dropIndexes` oplog record in the oplog file. 

After this change, the oplog will be replayed before the index creation, so I believe the `dropIndexes` oplog will be silently dropped. After the whole mongorestore is done, the index will still be created since it exists in the metadata file. 

Please let me know if my concern is valid. Thx.

tfogo:
@huan-Mongo Yes! Great catch! 

In mongomirror we intercept any oplog entries that change indexes and update the IndexCatalog instead of applying them. We might be able to move IndexCatalog into common and use that instead of intents.Manager for indexes. But it might also be a bit too tied into mongomirror. I'll see what I can do.

tfogo:
This is ready for review again. There's now a `common/idx` package that contains the index catalog. This is based on the index catalog in mirror except:

- there's no need to resync indexes 
- collMod supports `hidden` and specifying indexes by key
- there's some extra logic to maintain collation information for collections
- there's a `NamespaceQueue` which is used in `restore.RestoreIndexes()` to get the namespaces that need indexes built

The new namespace queue means I got rid of the index prioritizer on the intents manager. There are also changes in `oplog.go` to update the index catalog and ignore index builds/deletions.

mongorestore/metadata.go
```diff
@@ -180,12 +174,6 @@ func (restore *MongoRestore) CreateIndexes(dbName string, collectionName string,
 		if !restore.OutputOptions.KeepIndexVersion {
 			delete(index.Options, "v")
 		}
-
-		// for non-simple default collation on the collection, indexes without
-		// a collation option need to add "collation:{locale:"simple"}}
-		if _, ok := index.Options["collation"]; hasNonSimpleCollation && !ok {
```
tfogo, line 186: This is now handled in the IndexCatalog

mongorestore/mongorestore_test.go
```diff
@@ -808,38 +820,59 @@ func TestFixDuplicatedLegacyIndexes(t *testing.T) {
...
-			So(res.Key[0].Key, ShouldEqual, "foo")
-			So(res.Key[0].Value, ShouldEqual, 2)
+			var indexKey bson.D
+			// Check that only one of foo_, foo_1, or foo_1.0 was created
+			indexKeyFoo, ok := indexKeys["foo_"]
+			indexKeyFoo1, ok1 := indexKeys["foo_1"]
+			indexKeyFoo10, ok10 := indexKeys["foo_1.0"]
 
-			So(c.TryNext(context.Background()), ShouldBeFalse)
+			So(ok || ok1 || ok10, ShouldBeTrue)
```
tfogo, line 850: I had to redo this test since the index catalog uses a map and the order in which indexes are built is now random.

mongorestore/oplog.go
```diff
@@ -159,6 +147,54 @@ func (restore *MongoRestore) RestoreOplog() error {
 
 }
 
+func (restore *MongoRestore) HandleOp(oplogCtx *oplogContext, op db.Oplog) error {
```
tfogo, line 150: Pulled this out so we can recursively handle applyOps oplog entries.

mongorestore/oplog.go
```diff
@@ -167,32 +203,136 @@ func (restore *MongoRestore) HandleNonTxnOp(oplogCtx *oplogContext, op db.Oplog)
...
+				err = bson.Unmarshal(bytesOp, &nestedOp)
+				if err != nil {
+					return fmt.Errorf("could not unmarshal applyOps command: %v: %v", rawOp, err)
+				}
+
+				restore.HandleOp(oplogCtx, nestedOp)
+				if err != nil {
+					return err
+				}
+			}
```
huan-Mongo, line 291: Should we `return nil` here? I guess we don't want to run `restore.ApplyOps` on this op again, right?

huan-Mongo (changes requested):
Nice job! This is a pretty big PR. I think the change would work. Just some minor comments after the initial review.

mongorestore/oplog.go
tfogo: Yes, great catch!

rychipman (changes requested):
no new questions/requests, aside from the fact that there are some failing tests in the evg patch

tfogo:
Ah yeah, I put the `return nil` in the wrong place. Fixed now.

huan-Mongo (approved):
LGTM! Nice job!

edobranov (approved):
Found some time to go through everything and lgtm! It's a sizable refactor so nice work :)

rychipman (approved):
LGTM! nice work!

A40in:
here is a copy-paste mistake in IsEqual func.
```
func IsEqual(left, right bson.D) (bool, error) {
	leftBytes, err := bson.Marshal(left)
	if err != nil {
		return false, err
	}

	rightBytes, err := bson.Marshal(left) //must be right
```

tfogo:
@A40in thank you very much for letting us know about this. That is a very unfortunate bug. I have opened [TOOLS-2958](https://jira.mongodb.org/browse/TOOLS-2958) to fix this.

Did you encounter issues due to this bug? If you need a workaround, versions of mongorestore prior to v100.4.0 do not have this bug.

A40in:
I had no problem. I just tried using this function in tests. Good luck

PR #363, opened 2021-04-08
TOOLS-2824: Add static analysis task that runs `evergreen validate`

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2824

Evergreen patch with task: https://evergreen.mongodb.com/version/606f6d6cd6d80a5fbdd77f56

This adds a task for `evergreen validate` in the buildscript and updates our Static Analysis buildvariant with it. Also adding some sample output of different scenarios below.

Everything succeeds:
```
$ go run build.go -v sa:evgvalidate
START  | sa:evgvalidate
       | exec: '.../bin/evergreen validate --file common.yml'
FINISH | sa:evgvalidate in 971.621181ms
---------------
Completed in 971.889045ms
```

Breaking the syntax:
```
$ go run build.go -v sa:evgvalidate
START  | sa:evgvalidate
       | exec: '.../bin/evergreen validate --file common.yml'
FAIL   | sa:evgvalidate in 807.488006ms
       | error from `evergreen validate`: ERROR: load project error(s): error unmarshalling into parser project: yaml: line 282: did not find expected key
       | common.yml is an invalid configuration: exit status 1
task(s) [sa:evgvalidate] failed
exit status 2
```

Misspelling a distro name:
```
$ go run build.go -v sa:evgvalidate
START  | sa:evgvalidate
       | exec: '.../bin/evergreen validate --file common.yml'
FAIL   | sa:evgvalidate in 863.701372ms
       | error from `evergreen validate`: WARNING: yaml: unmarshal errors:
       |   line 302: key "mongod_port" already set in map
       |   line 294: key "mongod_port" already set in map
       |   line 305: key "mongod_port" already set in map
       | WARNING: buildvariant 'suse12' in project '' references a nonexistent distro 'susie12-sp5-small'.
       | common.yml is valid with warnings
task(s) [sa:evgvalidate] failed
exit status 2
```
^ for the last one, I'll make a comment below about `key "mongod_port" already set in map`

### DIFF
--- a/buildscript/sa.go
+++ b/buildscript/sa.go
@@ -57,7 +57,7 @@ func SAEvergreenValidate(ctx *task.Context) error {
 		return fmt.Errorf("error from `evergreen validate`: %s: %w", output, err)
 	}
 
-	// TODO: change this if-block in TOOLS-...
+	// TODO: change this if-block in TOOLS-2840.
 	// This check ignores any YAML warnings related to duplicate keys in YAML maps.
 	// See ticket for more details.
 	if strings.HasSuffix(output, "is valid with warnings") {
--- a/buildscript/sa.go
+++ b/buildscript/sa.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"strings"
 
 	"github.com/craiggwilson/goke/pkg/sh"
 	"github.com/craiggwilson/goke/task"
@@ -44,6 +45,29 @@ func SAModTidy(ctx *task.Context) error {
 		_ = ioutil.WriteFile("go.mod", origGoMod, 0600)
 		_ = ioutil.WriteFile("go.sum", origGoSum, 0600)
 		return errors.New("go.mod and/or go.sum needs changes: run `go mod tidy` and commit the changes")
+	}
+
+	return nil
+}
+
+// SAEvergreenValidate runs `evergreen validate` on common.yml and ensures the file is valid.
+func SAEvergreenValidate(ctx *task.Context) error {
+	output, err := sh.RunOutput(ctx, "evergreen", "validate", "--file", "common.yml")
+	if err != nil {
+		return fmt.Errorf("error from `evergreen validate`: %s: %w", output, err)
+	}
+
+	// TODO: change this if-block in TOOLS-...
+	// This check ignores any YAML warnings related to duplicate keys in YAML maps.
+	// See ticket for more details.
+	if strings.HasSuffix(output, "is valid with warnings") {
+		for _, line := range strings.Split(output, "\n") {
+			if !(strings.HasSuffix(line, "unmarshal errors:") ||
+				strings.HasSuffix(line, "already set in map") ||
+				strings.HasSuffix(line, "is valid with warnings")) {
+				return fmt.Errorf("error from `evergreen validate`: %s", output)
+			}
+		}
 	}
 
 	return nil
--- a/common.yml
+++ b/common.yml
@@ -7,9 +7,9 @@ command_type: system
 # run the same task in the previous revision if the current task fails
 stepback: true
 
-mongo_tools_variables:
+variables:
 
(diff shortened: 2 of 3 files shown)


buildscript/sa.go
```diff
@@ -48,3 +49,26 @@ func SAModTidy(ctx *task.Context) error {
...
 }
+
+// SAEvergreenValidate runs `evergreen validate` on common.yml and ensures the file is valid.
+func SAEvergreenValidate(ctx *task.Context) error {
+	output, err := sh.RunOutput(ctx, "evergreen", "validate", "--file", "common.yml")
+	if err != nil {
+		return fmt.Errorf("error from `evergreen validate`: %s: %w", output, err)
+	}
+
+	// TODO: change this if-block in TOOLS-...
```
edobranov, line 60: Since there's some shortcomings with the parser detailed in [EVG-14227](https://jira.mongodb.org/browse/EVG-14227), I'm thinking of making a TOOLS ticket blocked on that EVG ticket for removing the duplicate key special-casing below (I'll add it once everything is lgtm'd).

The special-casing is just looping through the `evergreen validate` output and only returns an error if it sees a warning that's not associated with duplicate keys. It shouldn't break when the Evergreen team updates to `yaml.v3`, since these types of warnings will just disappear.

common.yml
```diff
@@ -7,9 +7,9 @@ command_type: system
 # run the same task in the previous revision if the current task fails
 stepback: true
 
-mongo_tools_variables:
+variables:
```
edobranov, line 10: This should've been `variables` since `mongo_tools_variables` isn't a valid keyword to define on the root.

See [this line](https://github.com/evergreen-ci/evergreen/blob/a4d568801804ccc151e65fcaaaa00e3886be9fbd/self-tests.yml#L40) of Evergreen's config file.

common.yml
```diff
@@ -343,6 +343,7 @@ functions:
       working_dir: src/github.com/mongodb/mongo-tools
       script: |
         ${_set_shell_env}
+        PATH=$PATH:$HOME
```
edobranov, line 346: We need this for the `SAEvergreenValidate` func to run the `evergreen` binary, which is found in the `$HOME` directory on distros.

rychipman: perhaps we can do this in `_set_shell_env` instead?

buildscript/sa.go
rychipman: > I'm thinking of making a TOOLS ticket blocked on that EVG ticket for removing the duplicate key special-casing below

I think this is a good idea. I'd say just go ahead and file it now; no need to wait for this to get LGTM'ed

rychipman (changes requested):

common.yml
edobranov: Hmm, I tried simply adding `export PATH=$PATH:$HOME` at the bottom of `_set_shell_env` in [this patch build](https://evergreen.mongodb.com/task/mongo_tools_static_evergreen_validate_patch_cbdf89967b93b03a915e87d1c7694f77671c4769_6070a3f89ccd4e38a8c71257_21_04_09_18_59_11) (with the [common.yml diff](https://evergreen.mongodb.com/filediff/6070a3f89ccd4e38a8c71257/?file_name=common.yml&patch_number=0)), but it's somehow giving me a Go import cycle error.

I also tried it without `export` but it's the [same result](https://evergreen.mongodb.com/task/mongo_tools_static_evergreen_validate_patch_cbdf89967b93b03a915e87d1c7694f77671c4769_6070ad85c9ec440a16248859_21_04_09_19_39_54) (with the [common.yml diff](https://evergreen.mongodb.com/filediff/6070ad85c9ec440a16248859/?file_name=common.yml&patch_number=0)). Not totally sure why that's happening...

I was originally thinking that `_set_shell_env` was more for Evergreen expansions, but I'm fine exporting the PATH there if you have a sense of what could be causing that error.

rychipman (approved):
LGTM

PR #368, opened 2021-04-28
TOOLS-2832: Update to non-EOL windows distro

rychipman:
Evergreen patch: https://spruce.mongodb.com/version/6088cb1c3e8e8678e0e8fed2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


tfogo:
It looks like that patch has some failing tests that are passing on the base. I had a quick look at the logs and it looks like it's mostly connectivity issues so maybe they're just flakes. Could we re-run the failing tasks to check?

tfogo:
This is the same on all three PRs btw - they all have failing tasks.

PR #370, opened 2021-05-19
TOOLS-2866: Drop support for zSeries platforms

edobranov:
Jira: https://jira.mongodb.org/browse/TOOLS-2866

This removes all but the latest (`rhel72`) zSeries platforms from our Evergreen config.

I thought about renaming the `rhel67_s390x_tasks` to `rhel72_s390x_tasks` and using those tests in `rhel72`, but:
1) It doesn't work right off the bat ([Evg patch](https://spruce.mongodb.com/version/60a53dc32fbabe20ec7ed168)).
2) `rhel72` is going to be removed soon per [BUILD-12567](https://jira.mongodb.org/browse/BUILD-12567), so if anything, we can consider running tests on RHEL 8 directly when the time comes.


rychipman (changes requested):
> This removes all but the latest (rhel72) zSeries platforms from our Evergreen config.

RHEL 7.2 is being removed from evergreen as well, so it should also be removed as part of this patch.

These platforms also need to be removed from the release automation code

rychipman (approved):
LGTM

PR #371, opened 2021-05-20
TOOLS-2873 Run full test suite on all supported distros

tfogo:
- Add the full test suite for all supported distros.
- Regenerate test TLS certs since they had expired. New expiry is 10 years. Includes config and a script for regenerating them in the future.
- Removes the task list in common.yml and uses task tags instead.
- Fixes the legacy42 tests which were failing on 4.4 (needed to fix an index spec in the test data).

There are still quite a few test and system failures for the added tests. I suggest we merge this then open tickets to fix those tests one at a time.


common.yml
```diff
@@ -439,7 +168,7 @@ functions:
         if [ "${USE_TLS}" = "true" ]; then
           MONGOD_ARGS="${mongod_args_tls}"
         elif [ "${AWS_AUTH}" = "true" ]; then
-          MONGOD_ARGS="--auth --setParameter authenticationMechanisms=MONGODB-AWS,SCRAM-SHA-256"
```
rychipman, line 442: what's the reason for this removal?

common.yml
```diff
@@ -2098,87 +2196,108 @@ buildvariants:
...
+  - name: "push"
+    run_on: rhel80-small
 
 #######################################
 #     Z (s390x) Buildvariants         #
 #######################################
 
-- name: rhel67-s390x
-  display_name: ZAP s390x RHEL 6.7
+- name: rhel72-s390x
```
rychipman, line 2224: you'll probably see rebase conflicts when you rebase on the lastest master, but as of TOOLS-2866 we don't support any zseries platforms (except for rhel8-zseries, which we haven't added yet)

common.yml
```diff
@@ -2203,4 +2322,11 @@ buildvariants:
...
     USE_SSL: "true"
-  tasks: *ubuntu_x86_64_race_tasks
+  tasks: 
+  - name: "unit"
+  - name: ".3.4"
+  - name: ".3.6"
+  - name: ".4.0"
+  - name: ".4.2"
+  - name: ".4.4"
+  - name: "dist"
```
rychipman, line 2332: since we don't release on the race-detector buildvariant, is there any particular reason we need to run the `dist` task?

rychipman (changes requested):
a few small questions/comments. we should probably also reconfigure the evg patch to include all buildvariants, so that we know the newly added tasks will all pass before we merge

common.yml
tfogo: As far as I can tell it isn't a valid option for `authenticationMechanisms`. It was causing system failures for the `aws-auth` test.

tfogo:
I rebased so now the EOL zseries distros are no longer there. Also removed dist from the race variant.

rychipman (approved):
LGTM

PR #373, opened 2021-06-07
TOOLS-2857 Dump timeseries collections

tfogo:
This PR includes the following:

- Adds 5.0 tests to evergreen (this is based off the `TOOLS-2873` branch).
- Adds a spec for the archive format (I made this to help with implementation so I included it here).
- Adds support for dumping timeseries collections. Does not include `--query` and `--queryFile` support which will be completed in TOOLS-2858.

This takes a different approach to the original technical design. The original plan was to dump the timeseries collection as a view (i.e. metadata file only) and the system.buckets collection as a regular collection (with both bson and metadata files). Thus we would create two intents while dumping and three files will be produced.

This caused a couple of problems on the mongorestore side. The strategy of creating the timeseries collection and the system.buckets collection independently made usage with `--drop` complicated. Instead of attempting some sort of synchronization between the two intents, I thought it would make more sense to bundle them in one intent. This maintains the abstraction of a timeseries collection as a single entity. 

Also, I think it is probably going to be more reliable to create the system.buckets collection as part of creating the timeseries collection. This provides less exposure to the implementation details of how the system.buckets collection is created.

So instead, the timeseries and system.buckets collection get bundled into one intent. An intent with a `Type` of `"timeseries"` will dump a metadata file for the timeseries collection and a bson file for the system.buckets collection. On the restore side, this intent will cause the timeseries collection to be created and then the system.buckets collection is restored.

This is also implemented in archives by adding a `Type` field to `CollectionMetadata` (which is backwards compatible), and by making the demux/mux read/write data to the system.buckets collection for timeseries collections.


common.yml
```diff
@@ -8,253 +8,6 @@ command_type: system
 stepback: true
 
 variables:
-<<<<<<< HEAD
```
tfogo, line 11: This is a merge conflict on the TOOLS-2873 base branch that I need to fix

tfogo:
Here's an evg patch showing the integration tests passing on 5.0 and the Ubuntu 16.04 tests passing for previous versions: https://spruce.mongodb.com/version/60be8408d6d80a710258f6d8/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

There is a unit test failing which I need to fix - I think it is related to this PR.

The jstests that are failing are due to a setup failure. This is present in the TOOLS-2873 base branch.

mongodump/mongodump_test.go
```diff
@@ -1415,3 +1490,249 @@ func TestCount(t *testing.T) {
...
+					}
+				}
+
+				So(os.RemoveAll(archiveFilePath), ShouldBeNil)
+			})
+
+			Convey("to a directory", func() {
+
+				Convey("when dumping the whole database", func() {
+					// This is the default test of "to an archive".
```
tfogo, line 1597: I need to update this comment to "to a directory"

huan-Mongo (changes requested):
Looks good! Only get some questions on tests and comments.

common.yml
```diff
@@ -8,250 +8,6 @@ command_type: system
 stepback: true
 
 variables:
-
-  # List of tests to run on each buildvariant
-  mongo_tools_task_lists:
```
rychipman, line 13: can you rebase this PR on master now that TOOLS-2873 is merged?

common/archive/archive.go
```diff
@@ -25,6 +25,7 @@ type CollectionMetadata struct {
 	Collection string `bson:"collection"`
 	Metadata   string `bson:"metadata"`
 	Size       int    `bson:"size"`
+	Type       string `bson:"type"`
```
rychipman, line 28: Is the docstring's assertion that `There is one CollectionMetadata per collection that will be in the archive` still true? Or will these be handled like the intents, where there is one `CollectionMetadata` that covers both?

mongodump/mongodump_test.go
```diff
@@ -1415,3 +1490,249 @@ func TestCount(t *testing.T) {
...
+					md.ToolOptions.DB = dbName
+					md.ToolOptions.Collection = colName
+				})
+
+				Convey("even when the system.buckets collection is excluded", func() {
+					Convey("by --excludeCollections", func() {
+						md.OutputOptions.ExcludedCollections = []string{"system.buckets." + colName}
+					})
+
+					Convey("by --excludeCollectionsWithPrefix", func() {
```
rychipman, line 1540: Whoa, I didn't know `goconvey` allowed this! I know you could nest `Convey` calls to have a bunch of setup code that is rerun for each inner scope, but you can also have inner `Convey` calls do multiple branches of setup for assertions in an outer scope?

rychipman (changes requested):
looks good overall! just a few questions

mongodump/mongodump_test.go
huan-Mongo: Yea, surprising to me at first as well! Seems it reruns the outer scope code for each inner Convey calls!

common/archive/demultiplexer.go
```diff
@@ -60,9 +60,10 @@ func CreateDemux(namespaceMetadatas []*CollectionMetadata, in io.Reader) *Demult
 		In:              in,
 	}
 	for _, cm := range namespaceMetadatas {
-		ns := cm.Database + "." + cm.Collection
-		demux.NamespaceStatus[ns] = NamespaceUnopened
-
+		if cm.Type != "timeseries" {
```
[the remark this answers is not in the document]
tfogo: `NamespaceStatus` tracks the status of the demultiplexer reading the data blocks of a specific namespace. There are three options: 

- The demux has not encountered a data block for the namespace (`NamespaceUnopened`)
- The demux has encountered a data block for the namespace (`NamespaceOpened`)
- The demux has encountered the EOF for the namespace (`NamespaceClosed`)

There are no data blocks for the timeseries view, so this means there will be a `NamespaceUnopened` in `demux.NamespaceStatus` at the end of reading the archive. This causes an error in `demux.End()` which expects to see all namespaces in the closed state.

The `NamespaceStatus` for the `system.buckets` collection gets set in `demux.HeaderBSON()` as long as `demux.outs` includes a `DemuxOut` for the `system.buckets` collection. This is always the case since the `DemuxOut` gets added during `demux.Open()`. Therefore it's not necessary to add the `NamespaceStatus` for the `system.buckets` in `CreateDemux()`.

However, this is used as a double check against archive corruption in `demux.End()`. It ensures that we do see the data blocks expected for each metadata prelude. So I think to maintain this check, I'll add a `NamespaceStatus` for the `system.buckets` collection in `CreateDemux()`.

common/archive/spec.md
```diff
@@ -0,0 +1,92 @@
...
+        string metadata, 
+        int32 size,
+        string type
+    }
+    ```
+    - `db` - databse name.
+    - `collection` - collection name.
+    - `metadata` - the collection metadata (including options, index definitions, and collection type) encoded in canonical [Extended JSON v2](https://docs.mongodb.com/manual/reference/mongodb-extended-json/). This is the same data written to `metadata.json` files.
+    - `size` - the total uncompressed size of the collection in bytes.
+    - `type` - set to `"timeseries"` for timeseries collections, `""` otherwise.
```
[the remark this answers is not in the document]
tfogo: Yes, I forgot to add that!

mongodump/mongodump_test.go
```diff
@@ -265,6 +267,66 @@ func setUpMongoDumpTestData() error {
...
+		}
+		_, err = coll.InsertOne(nil, bson.M{"ts": primitive.NewDateTimeFromTime(time.Now()), "meta": metadata, "measurement": i})
+		if err != nil {
+			return err
+		}
+
+		idx := mongo.IndexModel{
+			Keys: bson.M{"meta.device": 1},
+		}
+		_, err = coll.Indexes().CreateOne(context.Background(), idx)
```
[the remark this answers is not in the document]
tfogo: Yes, good catch!

common/archive/archive.go
tfogo: Depends how you interpret "collection that will be in the archive". There is still one `CollectionMetadata` per collection that will appear in the data blocks. For a timeseries collection T, there will be no data blocks for T, instead there will be system.buckets.T. So there is a one-to-one mapping in that sense (although the namespace names are different).

However if you're thinking in terms of collections that will be created on the destination, then yes timeseries `CollectionMetadata` are now one-to-two. 

I think we should add a note to the comment explaining this.

huan-Mongo (approved):
LGTM!

rychipman (approved):
LGTM

PR #375, opened 2021-06-15
TOOLS-2849 Error during resharding when dumping with oplog

tfogo:


huan-Mongo (changes requested):
LGTM for most changes. Just a minor comment on the oplog check.

rychipman (changes requested):
One high-level thing that I just noticed now, and totally missed when reviewing the tech design -- in the scope, we said the following:

>Resharding oplog entries must result in an error in mongomirror and in mongorestore with --oplogReplay.

I think we have a couple options here:
- Implement these checks in `mongorestore` instead of `mongodump`
- Implement these checks in `mongorestore` in addition to `mongodump`
- Implement these checks only in `mongodump`, and go through the process for modifying the scope doc

any thoughts on how you'd like to proceed @tfogo?

tfogo:
@rychipman I think that was a mistake in the scope doc. I wanted us to implement this in the same way we implement collection renames, which is catching then in `mongodump`, not `mongorestore`. So I think we should alter the scope doc.

I think when I was writing the scope I mixed up which tool the rename collection check was in.

mongodump/oplog_dump.go
```diff
@@ -125,16 +126,35 @@ func (dump *MongoDump) checkOplogTimestampExists(ts primitive.Timestamp) (bool,
...
+
+	if ok && nsStr == "admin.system.version" {
+		return fmt.Errorf("cannot dump with oplog if admin.system.version is modified")
 	}
 
 	if _, err := raw.LookupErr("o", "renameCollection"); err == nil {
 		return fmt.Errorf("cannot dump with oplog while renames occur")
 	}
 
+	dbName := strings.SplitN(nsStr, ".", 2)[0]
```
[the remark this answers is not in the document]
tfogo: Ah good catch! Just pushed a fix.

huan-Mongo (approved):
LGTM!

mongodump/oplog_dump.go
```diff
@@ -125,16 +126,37 @@ func (dump *MongoDump) checkOplogTimestampExists(ts primitive.Timestamp) (bool,
...
 
 	if _, err := raw.LookupErr("o", "renameCollection"); err == nil {
 		return fmt.Errorf("cannot dump with oplog while renames occur")
 	}
 
+	if ok {
+		dbName := strings.SplitN(nsStr, ".", 2)[0]
+
+		if dbName == "config" {
+			if collNameRaw, err := raw.LookupErr("o", "create"); err == nil {
```
rychipman, line 151: should we also be checking `insert` commands?

rychipman (changes requested):
only one other question, which also came up on the corresponding mirror PR. Otherwise, LGTM!

mongodump/oplog_dump.go
tfogo: Oh yeah, I put that in the design to cover all the bases. But I think just checking for existence at the start and create oplog entries is sufficient, right? I don't see how we would miss anything with the checks we currently have.

rychipman: Resolving per discussion in standup this morning about how this should be sufficient, because collections created by an insert are still replicated as create+insert

rychipman (approved):
LGTM

tfogo:
rebased and fixed conflicts

PR #380, opened 2021-07-03
TOOLS-2893: fix legacy-js-tests 4.4 and 5.0

huan-Mongo:
It looks like the mongo shell removed `MongoRunner.runMongoTool` function since 4.4. Besides, many failures in js-4.2 4.4 tests are because dataPath is defaulting to '/data/db/' which is read-only on many distros.

### DIFF
--- a/test/legacy42/jstests/tool/command_line_quotes.js
+++ b/test/legacy42/jstests/tool/command_line_quotes.js
@@ -1,5 +1,43 @@
 jsTest.log("Testing spaces in mongodump command-line options...");
 
+/**
+ * Starts an instance of the specified mongo tool
+ *
+ * @param {String} binaryName The name of the tool to run
+ * @param {Object} opts options to pass to the tool
+ *    {
+ *      binVersion {string}: version of tool to run
+ *    }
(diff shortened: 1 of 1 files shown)


test/legacy42/jstests/tool/command_line_quotes.js
```diff
@@ -1,5 +1,43 @@
...
+ * @param {string} [opts.binVersion] - The version of the tool to run.
+ *
+ * @param {...string} positionalArgs - Positional arguments to pass to the tool after all
+ * options have been specified. For example,
+ * MongoRunner.runMongoTool("executable", {key: value}, arg1, arg2) would invoke
+ * ./executable --key value arg1 arg2.
+ *
+ * @see MongoRunner.arrOptions
+ */
+MongoRunner.runMongoTool = function(binaryName, opts, ...positionalArgs) {
```
huan-Mongo, line 22: This function is copied from mongo repo v4.2 branch.

tfogo (approved):
LGTM! Since Ryan is OOO until next week, I think it's probably okay to just merge this.

PR #383, opened 2021-07-06
TOOLS-2860 add timeseries rename tests

tfogo:
Thanks to the way timeseries collections are supported in TOOLS-2859, all that needs to be done for this ticket is adding tests for renaming timeseries collections.

### DIFF
--- a/mongorestore/mongorestore_test.go
+++ b/mongorestore/mongorestore_test.go
@@ -1687,6 +1687,7 @@ func TestRestoreTimeseriesCollections(t *testing.T) {
 			So(err, ShouldBeNil)
 
 			result := restore.Restore()
+			defer restore.Close()
 			So(result.Err, ShouldNotBeNil)
 		})
 
@@ -1697,6 +1698,7 @@ func TestRestoreTimeseriesCollections(t *testing.T) {
 			So(err, ShouldBeNil)
 
 			result := restore.Restore()
+			defer restore.Close()
 			So(result.Err, ShouldNotBeNil)
 		})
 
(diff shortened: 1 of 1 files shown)


tfogo:
Evergreen: https://spruce.mongodb.com/version/60e58d4b9ccd4e42f9c293fb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

mongorestore/mongorestore_test.go
```diff
@@ -1936,5 +1950,61 @@ func TestRestoreTimeseriesCollections(t *testing.T) {
...
+			count, err = testdb.Collection("foo_rename_ts").CountDocuments(nil, bson.M{})
+			So(err, ShouldBeNil)
+			So(count, ShouldEqual, 1000)
+
+			count, err = testdb.Collection("system.buckets.foo_rename_ts").CountDocuments(nil, bson.M{})
+			So(err, ShouldBeNil)
+			So(count, ShouldEqual, 10)
+		})
+
+		Convey("system.buckets should not be renamed if the timeseries collection is not renamed", func() {
```
huan-Mongo, line 1982: Is it more accurate to say if the system bucket is rename?

huan-Mongo (approved):
Only one suggestion on the test name, still LGTM though!

mongorestore/mongorestore_test.go
tfogo: @huan-Mongo do you mean something like:

```
"system.buckets collection should not be renamed if the timeseries collection is not renamed, even if the user tries to rename the system.buckets collection"
```

dragonlee8: Yes, that's what I meant。

huan-Mongo: Yes, that's what I meant.

PR #385, opened 2021-07-06
TOOLS-2849 fix resharding tests

tfogo:
The resharding tests were failing on standalones because they require a replica set to test the oplog. These tests now skip standalones.

The test was also flaking on replica sets due to the timing between creating and dropping the collections in the test. I have modified the timing and it looks like it no longer flakes.

Evergreen patch: https://spruce.mongodb.com/version/60e4d55fe3c3312642105c74/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


huan-Mongo (approved):
LGTM!

PR #387, opened 2021-07-09
TOOLS-2892: fix all aws-auth tests

huan-Mongo:


release/platform/platform.go
```diff
@@ -353,4 +353,12 @@ var platforms = []Platform{
 		Repos:     []string{RepoOrg, RepoEnterprise},
 		BuildTags: []string{"ssl", "sasl", "gssapi", "failpoints"},
 	},
+	{
+		Name:      "ubuntu1604",
```
huan-Mongo, line 357: I know zseries distro would be deprecated for ubuntu systems according to TOOLS-2866. However, this platform is still used for `z Race Detector test`. All the tests in that variant failed after TOOLS-2866. I tried to change the distro to RHEL 8 zseries, but there are other dependencies not working there. Maybe we should create another ticket to fix this variant?

common.yml
```diff
@@ -168,7 +168,7 @@ functions:
         if [ "${USE_TLS}" = "true" ]; then
           MONGOD_ARGS="${mongod_args_tls}"
         elif [ "${AWS_AUTH}" = "true" ]; then
-          MONGOD_ARGS="--auth --setParameter authenticationMechanisms=SCRAM-SHA-256"
+          MONGOD_ARGS="--auth --setParameter authenticationMechanisms=MONGODB-AWS,SCRAM-SHA-256"
```
huan-Mongo, line 171: @tfogo Seems `MONGODB-AWS` is a valid value according to here: https://docs.mongodb.com/drivers/node/fundamentals/authentication/mechanisms/#mongodb-aws

binaryurl.py
```diff
@@ -8,7 +8,10 @@
 import argparse
 import json
 import sys
-import urllib2
+try:
```
huan-Mongo, line 11: the python binary is python 2 on some machine and python3 on others. This way it could work for both.

common.yml
```diff
@@ -218,8 +218,8 @@ functions:
                 exit 0
             else
                 SECS=`expr $SECS + 1`
-                if [ $SECS -gt 20 ]; then
-                    echo "mongod not ready after 20 seconds"
+                if [ $SECS -gt 100 ]; then
+                    echo "mongod not ready after 100 seconds"
```
huan-Mongo, line 222: This gives more time for mongodb to start up.

common.yml
```diff
@@ -2116,7 +2120,7 @@ buildvariants:
     <<: *mongod_tls_startup_args
     <<: *mongo_tls_startup_args
     mongo_os: "ubuntu1604"
-    mongo_edition: "targeted"
+    mongo_edition: "enterprise"
```
huan-Mongo, line 2123: the binary only exists for enterprise Mongodb for some versions.

huan-Mongo:
This fixes most of the tests except some are failed with missing mongodb binary.

release/platform/platform.go
tfogo: Yeah, sounds good!

common.yml
tfogo: Ah okay! It's strange, it looks like it isn't valid on some distros. Here's MacOS 10.14 throwing an error because of it:

https://evergreen.mongodb.com/task_log_raw/mongo_tools_macos_aws_auth_4.4_patch_57f7f7eafdd68fd8d570dfa1848f7d1312724ed1_60e8dd620ae6067ce0cd2b06_21_07_09_23_36_03/0?type=T#L162

tfogo (approved):
LGTM! Looks like not all aws-auth tests are fixed, but most of them are. I'm happy to merge this and fix the remaining tests later.

binaryurl.py
rychipman: it seems like it would be a better fix to just ensure we're using python 3 on all machines

huan-Mongo: I agree. I was able to fix to python3 binary on all the machines after a few tweaks.

common.yml
```diff
@@ -2125,9 +2134,9 @@ buildvariants:
     USE_SSL: "true"
   tasks:
   - name: "unit"
-  - name: ".3.4"
-  - name: ".3.6"
-  - name: ".4.0"
+  - name: ".3.4 !.mmap"
+  - name: ".3.6 !.mmap"
+  - name: ".4.0 !.mmap"
```
rychipman, line 2139: what's the reason for this change?

release/platform/platform.go
rychipman: Sounds good to me as well. @huan-Mongo can you link the ticket here once you've filed it?

rychipman (changes requested):
> I'm happy to merge this and fix the remaining tests later.

This sounds fine to me as well. @huan-Mongo if you're going to take this approach, can you file the ticket and link here?

common.yml
huan-Mongo: Because when I check the tests like this one
https://evergreen.mongodb.com/task_log_raw/mongo_tools_ubuntu1604_arm64_integration_4.0_mmapv1_patch_5fb51b831d623729dbc2b68ad97575d8464e1903_60ec5b252fbabe24ffe508ad_21_07_12_15_09_51/0?type=T
The server complains 
```
Cannot start server with an unknown storage engine: mmapv1, terminating
```
 I suspect the mmapv1 is not available on those server versions on that platform.

huan-Mongo:
Yes @rychipman I have created TOOLS-2922 to address the remaining platforms.

release/platform/platform.go
huan-Mongo: TOOLS-2922 is created.

rychipman (changes requested):
> Yes @rychipman I have created TOOLS-2924 to fix the z Race detector test

Do we also need another ticket to track the remaining aws-auth-4.4 failures? Or are we just going to move this ticket back to Scheduled instead of closing?

huan-Mongo:
@rychipman Yes, another ticket https://jira.mongodb.org/browse/TOOLS-2922 was created for that purpose.

rychipman (approved):
LGTM

PR #388, opened 2021-07-12
TOOLS-2921 Update to Go Driver 1.6

tfogo:
In [GODRIVER-1881](https://jira.mongodb.org/browse/GODRIVER-1881), `EstimatedDocumentCount()` uses `$collStats` which will not work for views. So `DeferredQuery.Count()` now uses `CountDocuments()` for views.


huan-Mongo (approved):
LGTM!

dragonlee8 (approved):
LGTM!

rychipman (approved):
revendor LGTM, but if we could fix the `modtidy` failure before merging, that would be great!

PR #392, opened 2021-07-14
TOOLS-2925: fix release goroutine loop and log curatorArgs

huan-Mongo:


tfogo (approved):
LGTM!

release/release.go
```diff
@@ -1196,6 +1196,7 @@ func linuxRelease(v version.Version) {
 							"--username", os.Getenv("BARQUE_USERNAME"),
 							"--api_key", os.Getenv("BARQUE_API_KEY"),
 						}
+						log.Printf("curatorArgs %v\n", curatorArgs)
```
rychipman, line 1199: printing the args directly is problematic, because we will print the barque API key -- we probably need to do something like replace the API key with `XXXXX` if we're going to log this

rychipman (changes requested):
one comment on the code that's here, otherwise LGTM. as discussed on slack, let's also do the `NOTARY_*` env variable fixes in this patch

release/release.go
```diff
@@ -146,9 +146,14 @@ func streamOutput(prefix string, reader io.Reader) {
 	}
 }
 
-func runAndStreamStderr(logPrefix string, name string, args ...string) error {
+func runAndStreamStderr(logPrefix string, name string, envOverrides map[string]string, args ...string) error {
 	cmd := exec.Command(name, args...)
 
+	cmd.Env = os.Environ()
+	for k, v := range envOverrides {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
```
huan-Mongo, line 154: When there are duplicated env names, the last entry will be taken

release/release.go
```diff
@@ -1108,17 +1113,42 @@ func uploadRelease(v version.Version) {
...
+type VersionWithSigningInfo struct {
+	mongoVersionNumber string
+	notaryKeyName string
+	notaryToken string
 }
 
-var linuxRepoVersionsUnstable = map[string]string{
-	"development": "4.0.0-15-gabcde123", // any non-rc pre-release version will send the package to the "development" repo
+var linuxRepoVersionsStable = map[string]VersionWithSigningInfo{
+	"development": {"4.0.0-15-gabcde123", "", ""}, // any non-rc pre-release version will send the package to the "development" repo
```
huan-Mongo, line 1123: I assume dev and testing version doesn't need notary?

huan-Mongo:
@rychipman @tfogo I hardcoded the notary key name and token inside `linuxRepoVersionsStable` and removed sensitive info from curatorArgs before logging. Please take a look!

release/release.go
```diff
@@ -1202,15 +1232,35 @@ func linuxRelease(v version.Version) {
...
+						envOverridesLog := make(map[string]string)
+						for k, v := range envOverrides {
+							envOverridesLog[k] = v
+						}
+						apiKeyIdex := findArgIndex(curatorArgsLog, "--api_key")
+						if apiKeyIdex >= 0 {
+							curatorArgsLog[apiKeyIdex] = "[REDACTED]"
+						}
+						envOverridesLog["NOTARY_TOKEN"] = "[REDACTED]"
+						log.Printf("curatorArgs: %v, envOverrides: %v\n", curatorArgsLog, envOverridesLog)
```
tfogo, line 1251: Small thing, can we add the prefix to the front of this so we can easily associate these args with the curator command output?

release/release.go
```diff
@@ -1108,17 +1113,42 @@ func uploadRelease(v version.Version) {
...
 }
 
-var linuxRepoVersionsUnstable = map[string]string{
-	"development": "4.0.0-15-gabcde123", // any non-rc pre-release version will send the package to the "development" repo
+var linuxRepoVersionsStable = map[string]VersionWithSigningInfo{
+	"development": {"4.0.0-15-gabcde123", "", ""}, // any non-rc pre-release version will send the package to the "development" repo
+	"testing":     {"4.0.0-rc0", "", ""},          // any rc version will send the package to the "testing" repo
+	"4.3":         {"4.3.0", "server-4.4", os.Getenv("SIGNING_AUTH_TOKEN_SERVER_4_4")},              // any 4.3 stable release version will send the package to the "4.3" repo
+	"4.4":         {"4.4.0", "server-4.4", os.Getenv("SIGNING_AUTH_TOKEN_SERVER_4_4")},              // any 4.4 stable release version will send the package to the "4.4" repo
+	"4.9":         {"4.9.0", "server-5.0", os.Getenv("SIGNING_AUTH_TOKEN_SERVER_5_0")},              // any 4.9 stable release version will send the package to the "4.9" repo
```
tfogo, line 1127: We need to set this env variable in common.yml still, right?

tfogo (commented):
Just want to double check if we need to make an edit to common.yml too.

release/release.go
dragonlee8: We don't since I just override them during `exec.Command`

huan-Mongo: We don't since I just override them during exec.Command

release/release.go
huan-Mongo: Good idea! will add.

release/release.go
huan-Mongo: So it looks like the NOTARY_TOKEN and name env var only matter when we call `curator` script. Since I changed the cmd.Env before running it, the curator process will get an updated env value for these two variables, I don't think we need to change the common.yml anymore.

tfogo: But the `SIGNING_AUTH_TOKEN_SERVER_5_0` and `SIGNING_AUTH_TOKEN_SERVER_4_4` variables are from the environment of release.go. So we still need to set these so `release.go` can get these vars, right?

tfogo: Curently we do `export NOTARY_TOKEN='${signing_auth_token_server_4_4}'` in common.yml, but we need to export vars for 5.0 and 4.0.

huan-Mongo: These are already set on the host machine I think. So if you look at `common.yml`, 
```
          echo "${signing_auth_token_server_4_4}" > ./signing_auth_token
```
it's just retrieving this value and write to file. There is no place in our code to set these values.

huan-Mongo: I kinda get what you said, do you mean we need to have something like
```
          export NOTARY_TOKEN_4_4='${signing_auth_token_server_4_4}'
          export NOTARY_TOKEN_5_0='${signing_auth_token_server_5_0}'
```
inside common.yml and getEnv from these exported variables? Is that necessary?

tfogo: Yes, I think so. `os.Getenv()` needs to get the env variable from somewhere. So we need to set the environment variables there.

huan-Mongo: The thing is currently we call Getenv on "signing_auth_token_server_4_4" and ""signing_auth_token_server_5_0" which should already be set on the server.  `export NOTARY_TOKEN` in `common.yml` is only for curator and is already taken care of by this piece of code 
```
						envOverrides["NOTARY_KEY_NAME"] = versionWithSigningInfo.notaryKeyName
						envOverrides["NOTARY_TOKEN"] = versionWithSigningInfo.notaryToken
....
	cmd.Env = os.Environ()
	for k, v := range envOverrides {
		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
	

```

tfogo: How are they set on the server? Maybe I'm wrong here, but we have them set as Evergreen variables. But those variables are not set as environment variables on the server until we set them in the `pre` script in `common.yml`.

huan-Mongo: It's changed to NOTARY_TOKEN_4_4 and NOTARY_TOKEN_5_0 which are exported in common.yml

common.yml
```diff
@@ -767,6 +767,8 @@ pre:
           export AWS_SECRET_ACCESS_KEY='${release_aws_secret}'
           export NOTARY_KEY_NAME='server-4.4'
           export NOTARY_TOKEN='${signing_auth_token_server_4_4}'
```
rychipman, line 769: since these are now set on a per-curator-call basis in the go code, it seems like we _shouldn't_ have these variables set by default in the evg yml
```suggestion
```

tfogo (approved):
LGTM!

release/release.go
```diff
@@ -1108,17 +1113,42 @@ func uploadRelease(v version.Version) {
...
 }
 
-var linuxRepoVersionsStable = map[string]string{
-	"development": "4.0.0-15-gabcde123", // any non-rc pre-release version will send the package to the "development" repo
-	"testing":     "4.0.0-rc0",          // any rc version will send the package to the "testing" repo
-	"4.3":         "4.3.0",              // any 4.3 stable release version will send the package to the "4.3" repo
-	"4.4":         "4.4.0",              // any 4.4 stable release version will send the package to the "4.4" repo
-	"4.9":         "4.9.0",              // any 4.9 stable release version will send the package to the "4.9" repo
-	"5.0":         "5.0.0",              // any 5.0 stable release version will send the package to the "5.0" repo
+type VersionWithSigningInfo struct {
```
rychipman, line 1116:
```suggestion
type LinuxRepo struct {
    name string
```
a couple suggestions here:
- this struct really represents a linux repo, not a version, so let's call it `linuxRepo`
- since we have a struct now, I'd suggest that we just include the repo name as another field. Then `linuxRepoVersions{Stable,Unstable}` can just be `[]linuxRepo` instead of maps

release/release.go
```diff
@@ -1108,17 +1113,42 @@ func uploadRelease(v version.Version) {
...
 }
 
-var linuxRepoVersionsUnstable = map[string]string{
-	"development": "4.0.0-15-gabcde123", // any non-rc pre-release version will send the package to the "development" repo
+var linuxRepoVersionsStable = map[string]VersionWithSigningInfo{
+	"development": {"4.0.0-15-gabcde123", "", ""}, // any non-rc pre-release version will send the package to the "development" repo
+	"testing":     {"4.0.0-rc0", "", ""},          // any rc version will send the package to the "testing" repo
+	"4.3":         {"4.3.0", "server-4.4", os.Getenv("NOTARY_TOKEN_4_4")},              // any 4.3 stable release version will send the package to the "4.3" repo
+	"4.4":         {"4.4.0", "server-4.4", os.Getenv("NOTARY_TOKEN_4_4")},              // any 4.4 stable release version will send the package to the "4.4" repo
+	"4.9":         {"4.9.0", "server-5.0", os.Getenv("NOTARY_TOKEN_5_0")},              // any 4.9 stable release version will send the package to the "4.9" repo
```
rychipman, line 1127: based on what I'm seeing [here](https://www.mongodb.org/static/pgp/), it looks like there is actually a separate key for 4.9. Interestingly, there _doesn't_ appear to be a separate key for 4.3.

We might want to confirm the following with @briananthonymccarthy:
- what key to use for the 4.3 repo
- no key used for development and testing repos

release/release.go
```diff
@@ -1108,17 +1113,42 @@ func uploadRelease(v version.Version) {
...
+	"5.0":         {"5.0.0", "server-5.0", os.Getenv("NOTARY_TOKEN_5_0")},              // any 5.0 stable release version will send the package to the "5.0" repo
+}
+
+var linuxRepoVersionsUnstable = map[string]VersionWithSigningInfo{
+	"development": {"4.0.0-15-gabcde123", "", ""}, // any non-rc pre-release version will send the package to the "development" repo
+}
+
+// findArgIndex is the helper function to locate index of provided arg value from an array of arg list
+// The arg list array is assumed to be in such format: ["arg1_name", "arg1_value", "arg2_name", "arg2_value"...]
```
rychipman, line 1136: I don't think this is a safe assumption. In fact, this assumption is _already_ violated, and the only reason this function works is because our curator invocation happens to have exactly two positional args and no flags that don't take args. 

I'm concerned about this function silently failing in the future if the positional arg count changes or a no-arg flag is added. One possible mitigation is for the caller to panic if this returns `-1`, since we know that the `--api-key` flag should always exist

rychipman (changes requested):

release/release.go
huan-Mongo: Make sense. made the change.

release/release.go
huan-Mongo: I am checking with Brian on these two questions.

release/release.go
huan-Mongo: Make sense. I added the panic if `--api-key` not found

release/release.go
huan-Mongo: After talking with @zakhark , we figured 
1. There is no need to publish to 4.3 repo. It's a dev release repo and the Build team never advertised it. There are only unstable server versions inside.  I guess we were initially used it to test the separate release of Tools from the Server.  
So I removed 4.3 from versionsToRelease.
2. development and testing repos are not advertised either. No one is expected to use them with package manager. So it doesn't matter which key was used to sign them. However, since the notary key and token cannot be absent when calling curator, I set `server-4.4`key for these two repos.

huan-Mongo: Confirmed with Zahkar and Kelsey, 4.6-4.9 repos are not needed, so I removed 4.9 from here as well.

rychipman (approved):
LGTM, assuming evg patch passes!

PR #395, opened 2021-07-23
TOOLS-2926 remove dev/test repos from stable releases

tfogo:
I'm a bit confused about when we should be pushing to the testing repos. The comment seems to imply that we should only be publishing rc versions... but we aren't doing any filtering of versions. Does barque filter out requests based on version strings?


rychipman (approved):
LGTM

huan-Mongo:
LGTM! I don't think barque would filter out based on version strings.

PR #400, opened 2021-07-29
TOOLS-2863: cs.AuthMechanismProperties is not initialized when mechan…

huan-Mongo:
…ism set by --authenticationMechanism


tfogo (approved):
LGTM!

PR #402, opened 2021-08-05
TOOLS-2937 Set loadbalanced option in db.configureClient()

tfogo:
When `loadBalanced` is set, we need to make sure we don't set `direct` or `replicaSet`. We were doing that by default in some situations.

We also now error if a user sets `loadBalanced` with `direct` or `replicaSet` in the URI.

I tested this manually on serverless and it looks good!

### DIFF
--- a/common/db/db.go
+++ b/common/db/db.go
@@ -355,6 +355,10 @@ func configureClient(opts options.ToolOptions) (*mongo.Client, error) {
 		clientopt.SetMinPoolSize(cs.MinPoolSize)
 	}
 
+	if cs.LoadBalancedSet {
+		clientopt.SetLoadBalanced(cs.LoadBalanced)
+	}
+
 	if cs.ReadConcernLevel != "" {
 		rc := readconcern.New(readconcern.Level(cs.ReadConcernLevel))
 		clientopt.SetReadConcern(rc)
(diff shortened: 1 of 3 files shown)


common/db/db.go
```diff
@@ -355,6 +357,10 @@ func configureClient(opts options.ToolOptions) (*mongo.Client, error) {
 		clientopt.SetMinPoolSize(cs.MinPoolSize)
 	}
 
+	if cs.LoadBalancedSet {
+		clientopt.SetLoadBalanced(cs.LoadBalanced)
+	}
```
edobranov, line 362: I was looking at the documentation for `SetLoadBalanced` in the driver code in `clientoptions.go`:

```
// SetLoadBalanced specifies whether or not the MongoDB deployment is hosted behind a load balancer. This can also be
// set through the "loadBalanced" URI option. The driver will error during Client configuration if this option is set
// to true and one of the following conditions are met:
//
// 1. Multiple hosts are specified, either via the ApplyURI or SetHosts methods. This includes the case where an SRV
// URI is used and the SRV record resolves to multiple hostnames.
// 2. A replica set name is specified, either via the URI or the SetReplicaSet method.
// 3. The options specify whether or not a direct connection should be made, either via the URI or the SetDirect method.
```

Looks like for this PR, we'll now handle (2) and (3) in `common/options/options.go`, but not (1) afaict. Do we have a reason for not also handling (1) ourselves? Or maybe have all conflicts handled by the driver directly?

edobranov (approved):
LGTM! Just one general comment below, more for my understanding than anything else.

common/db/db.go
tfogo: Excellent catch! This was just an oversight on my part - I was only testing against serverless so the multiple hosts thing didn't come up. I added handling for this.

edobranov: Thanks! Looks all good now :)

huan-Mongo (approved):
LGTM!

PR #404, opened 2021-08-10
Tools 2725

huan-Mongo:


common.yml
```diff
@@ -2410,6 +2410,21 @@ buildvariants:
...
+#######################################
+- name: ubi8-container
+  display_name: UBI 8
+  run_on:
+    - ubi8
+  tasks:
+    - name: "dist"
+    - name: "sign"
+      run_on: rhel80-small
+    - name: "push"
```
tfogo, line 2423: We don't want to run a release for UBI 8. For container distros, we only want to run the tests.

tfogo (changes requested):

tfogo:
We recently decided to close TOOLS-2725 and TOOLS-2723 as wontfix so I'm going to close this PR.

PR #407, opened 2021-10-05
TOOLS-2952 Filter config collections in dump/restore

tfogo:
I ended up not running `shouldSkipSystemNamespace()` in `oplogDocumentValidator()` since that would skip all system collections (e.g. `db.system.views`, `admin.system.users`). That would be a behavior change requiring a major version bump. So the config collections are just filtered on the mongorestore side. Likewise I did not merge `shouldSkipSystemNamespace()` and `shouldIgnoreNamespace()` into one function in common.

This fixes a problem with restoring config collections that definitely should not be restored. We should open a ticket to investigate options for filtering the entire config and admin collections.

### DIFF
--- a/mongodump/prepare.go
+++ b/mongodump/prepare.go
@@ -151,7 +151,13 @@ func shouldSkipSystemNamespace(dbName, collName string) bool {
 			return true
 		}
 	case "config":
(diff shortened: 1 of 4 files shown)


edobranov (approved):
LGTM!

PR #409, opened 2021-11-10
TOOLS-2983 - Swap arguments for error messages

bgunics-talend:
I tried to change the order of the parameters to have the error messages more meaningful.


tfogo:
Thank you @bgunics-talend!

PR #410, opened 2021-11-15
MGOMIRROR-469: Fix verbosity bug when calling ParseArgs multiple times

huan-Mongo:
Currently we have a bug in the verbose flag parsing. The `SetVerbosity` function adds up the verbosity level every time when it's called. The reason for such logic is we wanna support command line flag such as 
`-v -v -v`(verbose level =3) or `-v --verbose=3`(verbose level =4)

However, this imposes a bug when `parseArgs`get called multiple times, we will multiple the verbose levels in the case. Currently `parseArgs` is get called both in `func (opts *ToolOptions) ParseArgs` and `ParseConfigFile`. 

There is no perfect solution for this bug right now because we cannot differentiate when `SetVerbosity` was called by a single `parseArgs` for a second time or by different `parseArgs` callers. So I made a little hacky fix.

I created a function `CallArgParser` to wrap `parseArgs` around. This function makes sure we reset the verbosity level every time when calling `parseArgs`, and the reset will only happen if the current  `parseArgs` detects verbosity flags that exist. We should not call `opts.parser.ParseArgs` directly as well.

### DIFF
--- a/common/options/options.go
+++ b/common/options/options.go
@@ -446,7 +446,7 @@ func (opts *ToolOptions) AddToExtraOptionsRegistry(extraOpts ExtraOptions) {
 	opts.URI.extraOptionsRegistry = append(opts.URI.extraOptionsRegistry, extraOpts)
 }
 
-func (opts *ToolOptions) CallParser(args []string) ([]string, error) {
+func (opts *ToolOptions) CallArgParser(args []string) ([]string, error) {
 	args, err := opts.parser.ParseArgs(args)
 	if err != nil {
 		return []string{}, err
@@ -470,7 +470,7 @@ func (opts *ToolOptions) ParseArgs(args []string) ([]string, error) {
 		return []string{}, err
 	}
 
-	args, err := opts.CallParser(args)
+	args, err := opts.CallArgParser(args)
 	if err != nil {
 		return []string{}, err
 	}
@@ -516,7 +516,7 @@ func LogSensitiveOptionWarnings(args []string) {
 
 	// Create temporary options for parsing command line args.
 	tempOpts := New("", "", "", "", true, EnabledOptions{Auth: true, Connection: true, URI: true})
-	extraArgs, err := tempOpts.CallParser(args)
+	extraArgs, err := tempOpts.CallArgParser(args)
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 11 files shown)


common/options/options.go
```diff
@@ -439,6 +446,20 @@ func (opts *ToolOptions) AddToExtraOptionsRegistry(extraOpts ExtraOptions) {
...
+func (opts *ToolOptions) CallArgParser(args []string) ([]string, error) {
+	args, err := opts.parser.ParseArgs(args)
+	if err != nil {
+		return []string{}, err
+	}
+
+	// Set VerbosityParsed flag to make sure we reset verbosity level when we call ParseArgs again
+	if opts.Verbosity.VLevel != 0 && !opts.VerbosityParsed {
+		opts.VerbosityParsed = true
+	}
```
edobranov, line 458:
```suggestion
	if opts.VLevel != 0 && !opts.VerbosityParsed {
		opts.VerbosityParsed = true
	}
```
[nit] Since we reference `VLevel` this way everywhere else.

common/options/options.go
```diff
@@ -439,6 +446,20 @@ func (opts *ToolOptions) AddToExtraOptionsRegistry(extraOpts ExtraOptions) {
...
+
+	return args, err
```
edobranov, line 460:
```suggestion
	return args, nil
```
[nit] Just to be explicit that we won't have an actual error here.

common/options/options.go
```diff
@@ -495,7 +516,7 @@ func LogSensitiveOptionWarnings(args []string) {
 
 	// Create temporary options for parsing command line args.
 	tempOpts := New("", "", "", "", true, EnabledOptions{Auth: true, Connection: true, URI: true})
-	extraArgs, err := tempOpts.parser.ParseArgs(args)
+	extraArgs, err := tempOpts.CallArgParser(args)
```
edobranov, line 519: We technically don't need to use `CallArgParser` here since we call it on the `tempOpts` that we create here, but I'm okay to leave it as-is since it doesn't matter much in the end.

edobranov, line 519: We technically don't need to use `CallArgParser` here since we call it once on the `tempOpts` that we create here, but I'm okay to leave it as-is since it doesn't matter much in the end.

edobranov (approved):
LGTM! I agree this is a bit hacky, but I really don't see a better way to handle this either. The fact that `-v -v ... -v` is valid and will cause the anonymous func we assign to `SetVerbosity` to get called multiple times makes it impossible to have this all be self-contained. The `CallArgsParser` wrapper seems like the next best thing to me.

common/options/options.go
huan-Mongo: Yea, I agree we don't need it here. But just to be consistent and emphasize we shouldn't be calling `parser.ParseArgs` directly even it's not a bug somewhere.

tfogo (approved):
LGTM!

PR #413, opened 2022-01-20
TOOLS-2745 correctly set retryWrites URI option

tfogo:


common/options/options.go
```diff
@@ -906,6 +906,10 @@ func (opts *ToolOptions) setOptionsFromURI(cs connstring.ConnString) error {
 		return fmt.Errorf("cannot use ssl: tool not built with SSL support")
 	}
 
+	if cs.RetryWritesSet {
+		opts.RetryWrites = &cs.RetryWrites
+	}
```
edobranov, line 911: Just to make sure I'm understanding this right -- `RetryWrites` can only be set through the conn string, which is why we _don't_ need to check for conflicting options / return a `ConflictingArgsErrorFormat` error, right?

tfogo: Yes, exactly.

edobranov (approved):

PR #414, opened 2022-01-25
TOOLS-2875 Add batch byte limit to BufferedBulkInserter

tfogo:
Add the batch byte limit back. The byte limit does not take selectors into account so it won't be perfect for delete, update, and replace. But it should be good enough to keep memory usage in check in most cases.

### DIFF
--- a/common/db/buffered_bulk.go
+++ b/common/db/buffered_bulk.go
@@ -16,6 +16,7 @@ import (
 )
 
 // The default value of maxMessageSizeBytes
+// See: https://docs.mongodb.com/manual/reference/command/hello/#mongodb-data-hello.maxMessageSizeBytes
 const MAX_MESSAGE_SIZE_BYTES = 48000000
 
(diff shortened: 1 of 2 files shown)


common/db/buffered_bulk.go
```diff
@@ -15,6 +15,9 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
+// The default value of maxMessageSizeBytes
+const MAX_MESSAGE_SIZE_BYTES = 48000000
```
edobranov, line 19:
```suggestion
// The default value of maxMessageSizeBytes
// See: https://docs.mongodb.com/manual/reference/command/hello/#mongodb-data-hello.maxMessageSizeBytes
const MAX_MESSAGE_SIZE_BYTES = 48000000
```
Could we just add this link below the comment? I always forget where this number comes from 😅

edobranov (approved):
LGTM! One nit below for consideration.

PR #423, opened 2022-05-19
TOOLS-3106 Remove 5.1 and 5.2 tasks from all variants that support 5.3+

autarch:
For RHEL7.2 on s390x, the most recent supported server version is 5.2, so we will keep the 5.1 and 5.2 tasks for that platform.


autarch:
CI shows no difference from master.

tfogo (changes requested):
Let's also remove all the task definitions that are tagged 5.1 or 5.2. We won't need those anymore.

tfogo (approved):
Ah yes! That's okay then. I think there's an interesting question here around what the tools support for rapid releases should look like. I think it might make sense to shrink the support matrix for rapid releases somewhat, but that's a discussion we should have later.

LGTM!

PR #428, opened 2022-06-01
Make bsondump accept documents up to the internal max bson size (16mb + 16kb)

autarch:


autarch:
There are no new CI failures which are not already accounted for in a ticket.

bsondump/bsondump_test.go
```diff
@@ -8,172 +8,284 @@ package bsondump
...
+		// subtract 1 byte for the trailing NULL in each of our two keys
+		2 -
+		// subtract another 4 bytes for the int32 specifying the length of the
+		// each of each of our two fields
+		8 -
+		// subtract 1 for the string's trailing NULL
+		1
+
+	t.Run("bsondump with file at exactly max size of 16mb + 16kb", func(t *testing.T) {
+		_, err := runBsondumpWithLargeFile(t, size, true)
```
tfogo, line 243: The fact that `runBsondumpWithLargeFile(t, size, true)` tests a document of size `size` + 16kb seems a bit unintuitive and not to easy to follow. I think it would be better if it tests a doc of size `size` and we pass in (16 MB + 16 KB) or (16 MB + 16 KB + 1 B) instead.

Without looking at the implementation of `runBsondumpWithLargeFile()` it looked like the limit was still 16 MB.

common/testutil/testutil.go
```diff
@@ -171,3 +174,19 @@ func MergeOplogStreams(input [][]db.Oplog) []db.Oplog {
...
 	return ops
 }
+
+func MakeTempDir(t *testing.T) (string, func()) {
+	require := require.New(t)
+
+	dir, err := ioutil.TempDir("", "mongo-tools-test")
+	require.NoError(err, "can create temp directory")
+	cleanup := func() {
+		if os.Getenv("TOOLS_TESTING_NO_CLEANUP") == "" {
```
tfogo, line 184: Let's document this env variable in CONTRIBUTING.md

tfogo, line 184: Let's document this env variable in CONTRIBUTING.md

bsondump/bsondump_test.go
```diff
@@ -8,172 +8,284 @@ package bsondump
...
+}
+
+func TestBsondumpMaxBSONSize(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	// 16mb
+	size := int(16*math.Pow(1024, 2)) -
+		// subtract 5 for the 4 bytes for the int32 at the head of the document that
+		// specifies its size and 1 for its trailing NULL.
+		5 -
```
tfogo, line 225: Hmm I'm not a huge fan of this formatting. I think it would be much tidier to add together all the subtractions and have one big block comment explaining how to construct the number. 

Or at least group together on one line. 

The BSON spec is very stable so it's not like this is something we expect to need to edit.

tfogo, line 225: Hmm I'm not a huge fan of this formatting. I think it would be much cleaner to add together all the subtractions and have one big block comment explaining how to construct the number. Or at least group together on one line. The BSON spec is very stable so it's not like this is something we expect to need to edit.

tfogo (changes requested):
Looking great in general. I think we should alter `runBsondumpWithLargeFile()` a bit to make it clearer what this function is doing.

bsondump/bsondump_test.go
autarch: I moved this around a bit and changed how it's documented. I would note that while the BSON spec won't change, the document contents could, and a lot of what this documents is tied to the specifics of the document, like how many fields it has, etc.

bsondump/bsondump_test.go
autarch: Yes, this is super confusing. I fixed it.

common/testutil/testutil.go
autarch: Done. I also added go docs for the `MakeTempDir` func.

tfogo (approved):
Fantastic! LGTM!

PR #431, opened 2022-06-03
TOOLS-2963 Move prompting for password to earlier in options processing

autarch:
Previously, we didn't prompt for a password until we attempted to establish a
database connection. But before we did that we would call
`connstring.ConnString.Validate`, which would return an error if the user for
the connection was set but no password was provided.

This commit moves the prompt for the password to just before the call to
`Validate`, so we should never end up complaining about a missing password
before having a chance to prompt the user.

I tested this by hand with a few different patterns:

$> go run ./mongotop/main/ -u user --authenticationMechanism 'PLAIN'
$> go run ./mongotop/main/ -u user --password="" --authenticationMechanism 'PLAIN'
$> go run ./mongotop/main/ -u user --password=""

I tried this with a few different tools, including mongodump, mongorestore,
and mongotop.

### DIFF
--- a/common/db/db.go
+++ b/common/db/db.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/mongodb/mongo-tools/common/log"
 	"github.com/mongodb/mongo-tools/common/options"
-	"github.com/mongodb/mongo-tools/common/password"
 	"github.com/youmark/pkcs8"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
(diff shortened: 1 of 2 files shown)


autarch:
I did some manual testing and I also ran a full set of integration & qa tests on one platform in CI to do some more checking. All tests in that patch passed or failed identically to master - https://spruce.mongodb.com/version/629a55ac57e85a36f9917d66/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

autarch:
BTW, this didn't seem very amenable to unit tests, but if you have an idea on how to do this, I'd be happy to implement it!

tfogo (approved):
LGTM!

PR #432, opened 2022-06-06
TOOLS-3039 Add a RHEL 6.2 x86-64 buildvariant that builds w/o SASL or Kerberos

autarch:
This variant will be signed and uploaded to S3, but we do not add it to the release feed or the RPM repos.


autarch:
I ran the `dist` task for this new variant. The resulting RPM is at https://mciuploads.s3.amazonaws.com/mongo-tools/pkgs/mongo_tools_rhel62_no_sasl_or_kerberos_patch_48ad24e08972af3330309bede669a8b4ca940303_629a6bca3627e03a4693b279_22_06_03_20_15_17/mongodb-database-tools-100.5.2.latest.x86_64.rpm

I downloaded the binary and extracted the `mongoexport` binary. When running `ldd` I get the following output:

```
$> ldd ./mongoexport 
	linux-vdso.so.1 (0x00007ffe8ff9f000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f98e3d85000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f98e3b5d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f98e3d9e000)
```

Whereas the version I have installed locally reports the following:

```
$> ldd $(which mongoexport )
	linux-vdso.so.1 (0x00007fff2d7e0000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f472fe45000)
	libgssapi_krb5.so.2 => /lib/x86_64-linux-gnu/libgssapi_krb5.so.2 (0x00007f472fdf1000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f472ebd8000)
	libkrb5.so.3 => /lib/x86_64-linux-gnu/libkrb5.so.3 (0x00007f472fd26000)
	libk5crypto.so.3 => /lib/x86_64-linux-gnu/libk5crypto.so.3 (0x00007f472fcf7000)
	libcom_err.so.2 => /lib/x86_64-linux-gnu/libcom_err.so.2 (0x00007f472fcef000)
	libkrb5support.so.0 => /lib/x86_64-linux-gnu/libkrb5support.so.0 (0x00007f472fce1000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f472fe5e000)
	libkeyutils.so.1 => /lib/x86_64-linux-gnu/libkeyutils.so.1 (0x00007f472fcda000)
	libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f472fcc6000)
```

So I _think_ this is working as it should.

autarch:
I won't merge this until after the 100.5.3 release, just to keep that on track.

tfogo (changes requested):
This is looking good, but I think `uploadReleaseJSON()` will still update the JSON feed with this release, which we don't want. So we probably want to filter out this sign task in `uploadReleaseJSON()` (or exclude it some other way).

autarch:
> This is looking good, but I think `uploadReleaseJSON()` will still update the JSON feed with this release, which we don't want. So we probably want to filter out this sign task in `uploadReleaseJSON()` (or exclude it some other way).

Good point. I updated the code and added some tests to the release code to check for this case. Now the test will ensure that platforms where `platform.SkipForJSONFeed` is true do not include the `push` task in CI.

release/release.go
```diff
@@ -927,7 +927,7 @@ func uploadReleaseJSON(v version.Version) {
 			continue
 		}
 
-		if _, ok := platform.GetByVariant(task.Variant); !ok {
+		if p, ok := platform.GetByVariant(task.Variant); p.SkipForJSONFeed || !ok {
```
tfogo, line 930: Now we'll have one less sign task, but `platform.Count()` will still be returning the number all the platforms right? So I think `len(signTasks) != pfCount` will return true.

tfogo (changes requested):

release/release.go
autarch: Oh, I think I misunderstood the question. I think you're right that this breaks the count check.

autarch: Fixed.

tfogo (changes requested):
Hey this looks good but it looks like we have some merge conflicts now. Can you resolve the merge conflicts then let me know?

release/platform/platform.go
```diff
@@ -336,22 +356,6 @@ var platforms = []Platform{
...
-		BuildTags: defaultBuildTags,
-	},
-	{
-		Name:      "debian92",
-		Arch:      ArchX86_64,
-		OS:        OSLinux,
-		Pkg:       PkgDeb,
-		Repos:     []Repo{RepoEnterprise, RepoOrg},
-		BuildTags: defaultBuildTags,
-	},
```
tfogo, line 354: Why are these debian platforms removed?

tfogo (commented):
Just one question...

release/platform/platform.go
autarch: They were in the list twice. I removed all dupes and added a test to make sure we don't add more in the future.

tfogo (approved):
LGTM!

PR #438, opened 2022-06-10
When submitting RPMs to barque use the RPMArch

autarch:
This fixes an issue with RHEL release on arm64. We need use "aarch64" as the arch name, not "arm64".


tfogo (approved):
LGTM!

PR #440, opened 2022-06-14
Remove Ubuntu 14.04 from the list of release platforms

autarch:
This also adds a test to check that our list of release platforms is in sync with what's in CI.


tfogo (approved):
LGTM

PR #441, opened 2022-07-06
TOOLS-3062: Add Amazon Linux 2022 x86 to Tools

mattChiaravalloti:
This PR adds Amazon Linux 2022 x86 to Tools. I followed [these instructions](https://github.com/mongodb/mongo-tools/blob/master/PLATFORMSUPPORT.md) (and the examples of existing amazon linux platforms) to add this one.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -2303,12 +2303,6 @@ buildvariants:
     resmoke_args: --jobs 4
   tasks:
   - name: "unit"
-  - name: ".3.6"
-  - name: ".4.0"
-  - name: ".4.2"
-  - name: ".4.4"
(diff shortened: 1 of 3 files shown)


mattChiaravalloti:
Removed tests for server versions before 6.0 since server does not support mongodb on amazon 2022 for earlier versions.

I also see that Go 1.17 does not seem to be installed on the evergreen image, so I may need to contact the BUILD team about that.

mattChiaravalloti:
Closing for now since this is now blocked on BUILD-15298.

tfogo (approved):
LGTM

PR #442, opened 2022-07-06
TOOLS-3058 + TOOLS-3060: Add Ubuntu 22.04 x86 and ARM to Tools

mattChiaravalloti:
This PR adds Ubuntu 22.04 x86 _and_ ARM to tools. I combined TOOLS-3058 and 3060 into one PR since they are similar. I followed [these instructions](https://github.com/mongodb/mongo-tools/blob/master/PLATFORMSUPPORT.md) (and the examples of existing ubuntu platforms) to add these two.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -2716,9 +2716,6 @@ buildvariants:
     resmoke_args: --jobs 4
   tasks:
   - name: "unit"
-  - name: ".4.4"
-  - name: ".5.0"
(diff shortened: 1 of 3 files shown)


autarch (approved):
LGTM!

mattChiaravalloti:
Updated the platforms to not test against any server versions before 6.0 since that is the first platform to have support for it.

PR #443, opened 2022-07-06
TOOLS-3056: Add RHEL9 x86 to Tools

mattChiaravalloti:
This PR adds RHEL 9.0 x86 ~_and_ ARM~ to tools. ~Similar to the ubuntu 22.04 support I did earlier, I combined TOOLS-3054 and 3056 into one PR since they are similar.~

Update: the evergreen RHEL9 ARM distro does not have Go installed, so that ticket (3054) is blocked. I am keeping this PR open since it can still cover the work for 3056.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -2526,6 +2526,32 @@ buildvariants:
   - name: "push"
     run_on: rhel80-small
 
+- name: rhel90
+  display_name: RHEL 9.0
+  run_on:
+  - rhel90-small
+  expansions:
+    <<: [ *mongod_ssl_startup_args, *mongo_ssl_startup_args, *mongod_tls_startup_args, *mongo_tls_startup_args ]
+    mongo_os: "rhel90"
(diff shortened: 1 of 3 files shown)


mattChiaravalloti:
There is still some debugging to do for the RHEL9 x86 platform. Looks like there is some compilation issue that needs to be addressed.

autarch (approved):
LGTM!

mattChiaravalloti:
The BUILD ticket blocking RHEL 9.0 x86 has been addressed, so I just started the task again to see if it works. I also updated this to include RHEL 9.0 ARM again since _that_ BUILD ticket has also been addressed. If those tasks go green, this will be good to merge 👍

mattChiaravalloti:
Ok, RHEL 9 ARM is still broken, but x86 is green. Will update this to only cover x86 and merge.

PR #445, opened 2022-07-14
TOOLS-3113 Add a mongorestore test of timeseries collections

autarch:


tfogo (changes requested):
In 5.0 we can create indexes on the metadata fields and the timestamp field. In 5.3+ we can create indexes on the measurement fields too. 

`{"metadata.sensorId", 1}, {"timestamp", 1}` is testing the 5.0 behavior, I think we do also want to test the 5.3 behavior for adding an index on `temp`. Can we add another index if we're 5.3+?

Otherwise, looks good.

tfogo (approved):
LGTM!

PR #451, opened 2022-07-18
TOOLS-3155 Fix repo config for RHEL 8.3

autarch:
The edition should be `enterprise`, not `org`.

### DIFF
--- a/etc/repo-config.yml
+++ b/etc/repo-config.yml
@@ -122,14 +122,6 @@ repos:
       - yum/redhat/8/mongodb-org
       - yum/redhat/8Server/mongodb-org
 
-  - name: rhel83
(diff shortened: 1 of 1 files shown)


etc/repo-config.yml
```diff
@@ -369,6 +361,14 @@ repos:
       - yum/redhat/8/mongodb-enterprise
       - yum/redhat/8Server/mongodb-enterprise
 
+  - name: rhel83
+    type: rpm
+    edition: enterprise
+    bucket: repo.mongodb.org
+    repos:
+      - yum/redhat/8/mongodb-org
```
tfogo, line 369: s/mongodb-org/mongodb-enterprise/

PR #456, opened 2022-07-29
TOOLS-2913 Prompt for SSL key password when key is encrypted

autarch:
I tested this by hand by running the following command:

    go run ./mongodump/main/ --ssl --sslPEMKeyFile=./common/db/testdata/test-client-pkcs8-encrypted.pem

This prompted me for the password.

### DIFF
--- a/common/db/db.go
+++ b/common/db/db.go
@@ -450,7 +450,7 @@ func configureClient(opts options.ToolOptions) (*mongo.Client, error) {
 		}
 
 		var x509Subject string
-		var keyPasswd string
+		keyPasswd := opts.SSL.SSLPEMKeyPassword
 		var err error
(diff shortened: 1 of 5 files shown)


tfogo (changes requested):
I think one point of confusion here is that if you also use `--password=""` that will prompt for a password. But the prompt text is the same in each case so it's not clear what password should be entered when. Can we have different prompts?

Otherwise, this is looking good.

autarch:
> I think one point of confusion here is that if you also use `--password=""` that will prompt for a password. But the prompt text is the same in each case so it's not clear what password should be entered when. Can we have different prompts?

Yes, we can. It now looks like this:

```
$> go run ./mongodump/main/ --ssl --sslPEMKeyFile=./common/db/testdata/test-client-pkcs8-encrypted.pem
Enter password for client certificate:

> go run ./mongodump/main/ -u foo
Enter password for mongo user:
```

PR #457, opened 2022-08-15
TOOLS-3150 Define default tasks to run in Evergreen config

autarch:
This PR adds a new Go tool to generate the `github_pr_aliases` portion of the
Evergreen config. Those "aliases" define which tasks are run by default
whenever a new PR is created in the repo. This replaces the config in
Evergreen, which I've removed.


autarch:
Notes for reviewers;

You'll want to view the diff with whitespace changes hidden. Otherwise the Evergreen YAML change is ridiculously huge. That's cause I ran `prettier` on the YAML to tidy it before making changes to it.

The default PR tasks now include all supports platforms and server versions, though not every possible combination. It ends up being about 40 more tasks than we used to run by default, which I think is reasonable.

I also found some new test failures, presumably caused by server changes:

* https://jira.mongodb.org/browse/TOOLS-3185
* https://jira.mongodb.org/browse/TOOLS-3186

These are almost certainly not related to the changes in this PR, so I don't think they should block merging.

evergreen/evergreen.go
```diff
@@ -0,0 +1,208 @@
...
+	aliases := []alias{
+		{
+			comment: "Run all static analysis tasks.",
+			variant: `^static$`,
+			tasks:   `.*`,
+		},
+		{
+			comment: "Run unit, sign, and push for every platform.",
+			variant: `.*`,
+			tasks:   `^(unit|sign|push)$`,
```
tfogo, line 110: I think we should not do sign or push for this. I don't think we get much out of it. Push gets skipped on patches anyway.

common.yml
```diff
@@ -823,2402 +822,2612 @@ timeout:
...
+      resmoke_use_ssl: _ssl
+      USE_SSL: "true"
+    tasks:
+      - name: "unit"
+      - name: ".3.6"
+      - name: ".4.0"
+      - name: ".4.2"
+      - name: ".4.4"
+
+github_pr_aliases:
```
tfogo, line 3397: Could we add a comment saying that this is generated and how it was generated? That way people won't accidentally manually edit it (I hope).

tfogo (changes requested):
Just a couple of small comments. I'm a bit worried that these will be a lot of tests to run - but we can try it out and see if we feel the need to cut it down.

edobranov (approved):
LGTM!

common.yml
autarch: Yes, good call! I added this.

tfogo (approved):
LGTM!

PR #483, opened 2022-10-28
TOOLS-3204 Fix Amazon 2 releases on ARM to be packaged as "aarch64"

autarch:


autarch:
I ran the `dist` task for the `amazon2-aarch64` variant. This produced release artifacts at https://mciuploads.s3.amazonaws.com/mongo-tools/task/dist/mongo_tools_amazon2_aarch64_patch_99b582e5bc618140cd8a329619bd99c6a8cd2ea7_635c00d57742ae2a9bebfe27_22_10_28_16_18_30/all-release-artifacts.tgz

I took a look inside the tarball and it contains a file named `mongodb-database-tools-100.6.0.latest.aarch64.rpm`. So it looks like this is the correct name per the ticket.

release/platform/platform.go
```diff
@@ -34,7 +34,11 @@ const (
 type Arch string
 
 const (
-	ArchArm64   Arch = "arm64"
+	ArchArm64 Arch = "arm64"
+	// While arm64 and aarch64 are the same architecture, some Linux distros
```
deriamis, line 38: I did much more testing, and `aarch64` packages will install on an `arm64` platform. I am not sure what I was seeing anymore in previous tests, but the conclusion is definitive - `aarch64` and `arm64` are aliases on all platforms. What this boils down to is consistency in where to find packages, not what will install.

huan-Mongo (commented):
LGTM!

release/platform/platform.go
autarch: Thanks, I updated this comment.

deriamis (approved):

tfogo (approved):
LGTM! Thanks @deriamis for digging into this to find the root cause!

PR #500, opened 2023-01-23
TOOLS-3229 Skip operations on admin DB if connected to atlas proxy for dump and restore

rohshar:
There is a longer writeup on the ticket about what the current behavior is, and what the proposed changes to listCollections would do to dump and restore.

Essentially, this PR changes both mongodump and mongorestore **to not try to dump or restore from/to the admin database if connected to an atlas proxy**. This means that these tools **don't write to the admin DB, and don't run any read commands on the admin DB (other than isMaster)**

**For restore**, these changes will skip trying to restore roles/users from the admin DB, and error if you explicitly try to restore users/roles or to the admin DB if you're connected to an atlas proxy. It also goes a bit farther with an extra safeguard of not ever creating intents for admin collections if a restore is running connected to an atlas proxy (because the command will fail with the following error if any collection isn't specially handled):

```
2023-01-23T19:13:16.780-0500    Failed: admin.system.testtest: error creating collection admin.system.testtest: error running create command: (AtlasError) (Unauthorized) not authorized on admin to execute command { create: "system.testtest", idIndex: { key: { _id: 1 }, ns: "admin.system.testtest", name: "_id_" }, lsid: { id: {4 [55 231 33 172 61 34 78 101 159 54 51 76 118 51 228 18]} }, $clusterTime: { clusterTime: {1674519196 7}, signature: { hash: {0 [154 29 28 221 113 60 177 235 1 108 192 125 68 226 245 137 251 139 223 94]}, keyId: 7127357047842013184.000000 } }, $db: "admin", $readPreference: { mode: "primary" } }
```
After these changes, the behavior is that restoring to a shared cluster (always will be connecting through atlas proxy) is that it ignores admin collections in the export it is restoring from.

**For dump**, these changes will skip trying to dump users/roles from the admin DB, and explicitly error if you try to dump from those collections or the admin DB if you're connected to an atlas proxy. It also makes it so that no intents are created for admin collections if a dump is running connected to an atlas proxy.

Also turns out that I need to also disable oplog dumping and replay from mongodump and mongorestore respectively. Oplog dumping seems to work, but oplog restoring fails with the following error:

```
2023-01-24T16:05:34.721-0500    Failed: restore error: error applying oplog: applyOps: (AtlasError) (Unauthorized) not authorized on admin to execute command { applyOps: [[{ts {1674592679 10}} {t 159} {h <nil>} {v 2} {op i} {ns test.regularcol} {o [{_id 0}]} {lsid [{id {4 [91 204 15 135 195 165 70 162 162 54 58 17 222 14 102 254]}} {uid {0 [117 218 168 183 175 47 46 144 84 62 115 174 168 88 246 113 208 165 191 247 111 254 241 125 101 84 176 146 118 7 18 193]}}]} {txnNumber 1} {prevOpTime [{ts {1674592679 9}} {t 159}]}]], lsid: { id: {4 [178 248 238 127 0 29 68 63 157 176 204 223 99 132 13 220]} }, $clusterTime: { clusterTime: {1674594334 16}, signature: { hash: {0 [0 218 2 95 41 132 114 201 143 211 115 83 98 36 128 221 239 143 89 254]}, keyId: 7127357047842013184.000000 } }, $db: "admin", $readPreference: { mode: "primary" } }
```


mongorestore/mongorestore.go
```diff
@@ -126,26 +127,6 @@ func New(opts Options) (*MongoRestore, error) {
 	return restore, nil
 }
 
-// SupportsCollectionUUID was removed from common/db/command.go, so copied to here
-func SupportsCollectionUUID(sp *db.SessionProvider) (bool, error) {
```
rohshar, line 130: FCV 3.6 isn't supported by mongorestore, and listCollections with atlas proxy would break this function. Best to remove it.

rohshar: I am waiting on confirmation from louisa that I can remove this.

common/db/command.go
```diff
@@ -127,6 +128,28 @@ func (sp *SessionProvider) DatabaseNames() ([]string, error) {
...
+	if result.Err() != nil {
+		if strings.Contains(result.Err().Error(), "no such command") {
+			return false, nil
+		}
+		return false, result.Err()
+	}
+
+	return true, nil
+}
+
```
ahmedssulaiman, line 152: This section looks good to me.

mongodump/mongodump_test.go
```diff
@@ -828,6 +828,44 @@ func TestMongoDumpBSONLongCollectionName(t *testing.T) {
 	})
 }
 
+func TestMongoDumpAtlasProxy(t *testing.T) {
```
autarch, line 831: It'd be nice to use `testify` for new test methods. I've been trying to move this code away from `convey`.

mongodump/mongodump_test.go
```diff
@@ -828,6 +828,44 @@ func TestMongoDumpBSONLongCollectionName(t *testing.T) {
...
+			So(err, ShouldBeNil)
+			So(dbNames, ShouldNotContain, "admin")
+		})
+
+		Convey("With DumpDBUsersAndRoles is true", func() {
+			// This case should error because it has explicitly been set to dump users and roles for a DB, but thats
+			// not possible with an atlas proxy.
+			md.OutputOptions.DumpDBUsersAndRoles = true
+			err = md.ValidateOptions()
+			So(err, ShouldBeError)
```
autarch, line 860: It'd be good to test that the error content or type matches what we expect, not just that we got any error.

mongorestore/metadata_test.go
```diff
@@ -22,13 +22,45 @@ import (
...
+			So(err, ShouldBeNil)
+			So(restore.ShouldRestoreUsersAndRoles(), ShouldBeFalse)
+		})
+
+		Convey("With RestoreDBUsersAndRoles is true", func() {
+			// This case should error because it has explicitly been set to restore users and roles, but thats
+			// not possible with an atlas proxy.
+			restore.InputOptions.RestoreDBUsersAndRoles = true
+			err = restore.ParseAndValidateOptions()
+			So(err, ShouldBeError)
```
autarch, line 56: Same comment re: checking the error (and using `testify`).

autarch (changes requested):
I just had some comments on the tests. Otherwise LGTM!

mongodump/mongodump_test.go
rohshar: Should we wait until we have a longer term plan for the tools? The testing in general is a bit of a mess, and I'm not sure if its worth doing something like that till we have a plan.

mongorestore/metadata_test.go
rohshar: 👍

mongodump/mongodump_test.go
autarch: I've been using `testify` for new tests so there's already some code using it.

mongorestore/mongorestore.go
```diff
@@ -198,6 +179,20 @@ func (restore *MongoRestore) ParseAndValidateOptions() error {
 		log.Logv(log.DebugLow, "restoring to a sharded system")
 	}
 
+	restore.isAtlasProxy, err = restore.SessionProvider.IsAtlasProxy()
+	if err != nil {
+		return err
+	}
+	if restore.isAtlasProxy {
+		if restore.InputOptions.RestoreDBUsersAndRoles || restore.ToolOptions.Namespace.DB == "admin" {
+			return fmt.Errorf("cannot restore to the admin database when connected to an atlas proxy")
```
tfogo, line 188: "atlas proxy" isn't a term users know. We should refer to them as "MongoDB Atlas free or shared clusters".

tfogo (changes requested):
Just one comment about error messages. Otherwise, great work!

mongorestore/mongorestore.go
rohshar: Seems that Ops Manager 5.0 does support FCV 3.4. However, --preserveUUID is only use for 3.6+. Since Ops Manager is the only user of this with 3.4, we should be able to remove this. Note, if anyone tries to run mongorestore with FCV 3.4 with --preserveUUID, the behavior will be undefined.

tfogo (approved):
LGTM. Let's make sure to open a DOCSP ticket to get this behavior added to the docs too.

rohshar:
DOCS ticket is here: https://jira.mongodb.org/browse/DOCS-15850

autarch (approved):
LGTM!

PR #501, opened 2023-01-25
TOOLS-3203 Ignore config DB for full dump unless specified

rohshar:
We intermittently get bugs about authorization errors because we don't skip all of the config.* collections that we should in mongodump and mongorestore. 

For **mongodump** instead of blacklisting collections, we can skip the config DB entirely on a full dump, unless it is explicitly specified (eg by passing --db="config").

I didn't modify `shouldSkipSystemNamespace`, and that means that when specifying --db="config", you still won't get collections in the system blacklist dumped. The only way to include collections in the blacklist is to specify both --db="config" and --collection=<collection_in_blacklist>.

For **mongorestore**, instead of creating intents on all DBs, we skip the config DB entirely on a full dump, unless one of the two following scenarios happen:

1. --db="config" is passed in
2. Any --nsInclude filter(s) are passed. In this case we rely on the Matcher matchstring and include all namespaces in the export that match the configured patterns. The result will be that config namespaces will be skipped, unless matched by this filter.

### DIFF
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -586,6 +586,8 @@ func TestMongoDumpSkipsConfigDBForFullDump(t *testing.T) {
 	require.NoError(t, err)
 	require.NotContains(t, dbNames, "config")
 	require.NotContains(t, dbNames, "local")
+	session.Database("config").Collection("testcol").Drop(nil)
+	session.Database("local").Collection("testcol").Drop(nil)
 }
 
 func TestMongoDumpValidateOptions(t *testing.T) {
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -565,29 +565,6 @@ func testDumpOneCollection(md *MongoDump, dumpDir string) {
 	})
 }
 
-func TestMongoDumpSkipsConfigDBForFullDump(t *testing.T) {
-	testtype.SkipUnlessTestType(t, testtype.IntegrationTestType)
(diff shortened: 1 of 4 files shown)


rohshar:
I will open a DOCS ticket once the PR is approved.

mongodump/mongodump_test.go
```diff
@@ -1865,6 +1866,8 @@ func TestFailDuringResharding(t *testing.T) {
 		md.ToolOptions.Namespace = &options.Namespace{}
 		err = md.Init()
 		So(err, ShouldBeNil)
+		// Hack to not fail validating inputs on both --db and --oplog being specified.
+		md.ToolOptions.DB = "config"
```
rohshar, line 1870: Putting this after .Init() is a temporary hack I made to get around this error message: https://github.com/mongodb/mongo-tools/blob/master/mongodump/mongodump.go#L117. Without specifying config, dumping the config DB is no longer possible, so some of the downstream errors could never happen.

This is an indicator of a new limitation. We now essentially no longer support --oplog with dumping the "config" DB at all ("config" isn't dumped by default, and can't specify --db and use --oplog). I'm not totally sure that is acceptable. If that limitation is acceptable, it might be better for me to truncate this test heavily (or remove it).

rohshar:
This PR will be a draft for now, until we want to do a major release.

FGasper:
Fixed via https://github.com/mongodb/mongo-tools/pull/543

PR #504, opened 2023-02-02
TOOLS-3241 Fix TestFailDuringResharding so it doesn't sometimes error with a different message

rohshar:
In the last 3 sub-tests of TestFailDuringResharding, it expects to fail when trying to dump the oplog, but it fails earlier, when it tries to create the intents for a database. This is a flaky test, most of the time it seems that it doesn't error early, but it is possible. We can skip creating the intents for the database if we specify a database, but it should still try to dump the oplog since its specified. This will ensure that it doesn't error on the earlier case and fail on that string comparison.

An example error is here: https://evergreen.mongodb.com/lobster/evergreen/test/mongo_tools_rhel83_s390x_integration_5.0_cluster_patch_5d5402b0da91ca3a99a3d19f0265d98cae16a7f6_63dbec3ed1fe074e9be5a1da_23_02_02_17_00_47/0/mongodump/#bookmarks=0%2C212&l=1&shareLine=131

### DIFF
--- a/mongodump/mongodump.go
+++ b/mongodump/mongodump.go
@@ -296,6 +296,27 @@ func (dump *MongoDump) Dump() (err error) {
 		return fmt.Errorf("error connecting to host: %v", err)
 	}
 
+	// If oplog capturing is enabled, we first check the most recent
+	// oplog entry and save its timestamp, this will let us later
+	// copy all oplog entries that occurred while dumping, creating
+	// what is effectively a point-in-time snapshot.
+	if dump.OutputOptions.Oplog {
+		err := dump.determineOplogCollectionName()
+		if err != nil {
+			return fmt.Errorf("error finding oplog: %v", err)
+		}
+		log.Logvf(log.Info, "getting most recent oplog timestamp")
+		dump.oplogStart, err = dump.getOplogCopyStartTime()
+		if err != nil {
+			return fmt.Errorf("error getting oplog start: %v", err)
+		}
+	}
+
+	if failpoint.Enabled(failpoint.PauseBeforeDumping) {
+		log.Logvf(log.Info, "failpoint.PauseBeforeDumping: sleeping 15 sec")
+		time.Sleep(15 * time.Second)
+	}
+
 	// switch on what kind of execution to do
 	switch {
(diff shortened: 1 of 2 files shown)


tfogo:
So this does fix it, but I think this is not the most correct way to do it. You actually can't set `--db` with `--oplog` and it will fail during option parsing. So this setup isn't actually a valid set of options.

The reason it's failing is because the `PauseBeforeDumping` failpoint causes mongodump to pause after the intents are created, which is the point at which we error if `config.reshardingOperations` exists. So there's a 1 second window in which this could happen.

I think the best fix here is to move this block:

```go
if failpoint.Enabled(failpoint.PauseBeforeDumping) {
  log.Logvf(log.Info, "failpoint.PauseBeforeDumping: sleeping 15 sec")
  time.Sleep(15 * time.Second)
}
```

To above:

```
// switch on what kind of execution to do
```

Where we create intents.

This means `config.reshardingOperations` will be in the oplog but not in listCollections.

Even in this case there's still a race because it looks like there's a bug in mongodump. `dump.oplogStart, err = dump.getOplogCopyStartTime()` is called after we create the intents, which means it's possible that we miss DDL events between creating intents and dumping. I think this block should be moved up too, and it should be before `PauseBeforeDumping`. Let me know if you have any questions.

rohshar:
It does make sense that we need to start looking at the oplog from a time before or equal to where we create the events. I've made these changes.

tfogo:
I actually think there was still a race with the test. Since it created `config.reshardingOperations` in a loop, the collection could still exist while creating intents. I made it so it waits 2 secs before creating the collection and it only does it once. Still not great, but I think that was the original intent of the test.

tfogo:
Let me know what you think, but LGTM if it seems okay to you.

tfogo (approved):

rohshar:
Its a little weird. It doesn't look like it was creating the collection in a loop, it would just do one of the cases in the select statement (the default because the other one wouldn't have a message yet). I think your change is good, because it makes the logic more clear. I assume that you also wanted to change the similar cases later in the test.

tfogo (approved):
LGTM

PR #513, opened 2023-02-16
TOOLS-3243 Set password from user prompt in the connection string

autarch:


autarch:
This was a one line fix. Most of the PR is new tests.

common/options/options.go
```diff
@@ -694,6 +694,7 @@ func (opts *ToolOptions) NormalizeOptionsAndURI() error {
 			return fmt.Errorf("error reading password: %v", err)
 		}
 		opts.Auth.Password = pass
+		opts.ConnString.Password = pass
```
john-morales, line 697: (Ah, I missed seeing that - thank you! 👏 )

tfogo (approved):
Nice! LGTM!

common/options/options_test.go
```diff
@@ -1091,3 +1092,135 @@ func TestDeprecationWarning(t *testing.T) {
 		})
 	})
 }
+
+func TestPasswordPrompt(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	pw := "some-password"
+	expectPrompt := "Enter password for mongo user:\n"
```
FGasper, line 1100: Testing on a specific, human-readable string will cause a spurious failure if we ever reword this phrase. For example,  (e.g., to have the actual username rather than `mongo user`). The test would be more resilient if it checked just for, e.g., `password.*:\n`

autarch: Good call, done.

edobranov (approved):

FGasper (approved):
LGTM

PR #519, opened 2023-02-21
TOOLS-3252: Test support for Columnstore Indexes

tdq45gj:
This PR adds new integration tests for Columnstore Indexes in `mongodump` and `mongorestore`. 

In `mongodump`:
- Create collections with a Columnstore Index created with `columnstoreProjection` key.
- Dump the collections.
- Check the dumped meta files for the correct number of Columnstore Indexes, Columnstore Index name, and `columnstoreProjection` key.

In `mongorestore`:
- Create a collection with a Columnstore Index created with `columnstoreProjection` key.
- Dump the collection and restore it. 
- Check the restored collection for the correct number of Columnstore Indexes, Columnstore Index name, and `columnstoreProjection` key.

### DIFF
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -2005,6 +2005,7 @@ func TestFailDuringResharding(t *testing.T) {
 	})
 }
 
+// TestMongoDumpColumnstoreIndexes tests dumping a collection with Columnstore Indexes.
 func TestMongoDumpColumnstoreIndexes(t *testing.T) {
 	testtype.SkipUnlessTestType(t, testtype.IntegrationTestType)
 	log.SetWriter(ioutil.Discard)
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -348,6 +348,33 @@ func setUpDBView(dbName string, colName string) error {
 	return nil
 }
 
+func setUpColumnstoreIndex(dbName string, colName string) error {
+	sessionProvider, _, err := testutil.GetBareSessionProvider()
+	if err != nil {
+		return err
+	}
+
+	createIndexCmd := bson.D{
+		{"createIndexes", colName},
+		{"indexes", bson.A{
+			bson.D{
+				{"key", bson.D{{"$**", "columnstore"}}},
+				{"name", "dump_columnstore_test"},
+				{"columnstoreProjection", bson.D{
+					{`"`, 0},
+				}},
+			},
+		}},
+	}
+	var r bson.M
+	err = sessionProvider.Run(createIndexCmd, &r, dbName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func turnOnProfiling(dbName string) error {
 	sessionProvider, _, err := testutil.GetBareSessionProvider()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mongodump/mongodump_test.go
```diff
@@ -1977,3 +2004,83 @@ func TestFailDuringResharding(t *testing.T) {
...
+
+	c2, err := countMetaDataFiles(dumpDBDir)
+	require.NoError(t, err)
+	require.Equal(t, c1, c2)
+
+	metaFiles, err := getMatchingFiles(dumpDBDir, ".*\\.metadata\\.json")
+	require.NoError(t, err)
+	require.Greater(t, len(metaFiles), 0)
+
+	oneMetaFile, err := os.Open(util.ToUniversalPath(filepath.Join(dumpDBDir, metaFiles[0])))
```
rohshar, line 2058: Would it be difficult to have it check the columnstore indexes for all relevant collections?

mongorestore/mongorestore_test.go
```diff
@@ -2061,8 +2061,9 @@ func TestRestoreTimeseriesCollections(t *testing.T) {
 // ----------------------------------------------------------------------
 
 type indexInfo struct {
-	name string
-	keys []string
+	name                  string
+	keys                  []string
+	columnstoreProjection map[string]int32
```
rohshar, line 2066: Add a comment here

mongorestore/mongorestore_test.go
```diff
@@ -2345,3 +2346,223 @@ func runMongodump(t *testing.T, dir, db, collection string, args ...string) stri
...
+	})
+
+	res := session.Database("admin").RunCommand(context.Background(), bson.M{"replSetGetStatus": 1})
+	if res.Err() != nil {
+		t.Skip("server is not part of a replicaset so we cannot test restore from oplog")
+	}
+	t.Run("restore from oplog", func(t *testing.T) {
+		testRestoreColumnstoreIndexFromOplog(t)
+	})
+	//t.Run("restore from oplog with default index name", func(t *testing.T) {
```
rohshar, line 2374: Did you mean to remove this?

mongorestore/mongorestore_test.go
```diff
@@ -2345,3 +2346,223 @@ func runMongodump(t *testing.T, dir, db, collection string, args ...string) stri
...
+	//	testRestoreColumnstoreIndexFromOplog(t, "custom index name")
+	//})
+}
+
+func testRestoreColumnstoreIndexFromDump(t *testing.T) {
```
rohshar, line 2379: I know there aren't comments on the old code which is where the structure of these functions and the test as a whole are from, but I think its important that we add robust comments here. Lets add docstrings to all functions you add, so its very easy to follow along with what is happening at each point.

rohshar (commented):
A few comments on clarity. Lets add a PR description.

mongodump/mongodump_test.go
```diff
@@ -348,6 +348,33 @@ func setUpDBView(dbName string, colName string) error {
...
+	}
+
+	createIndexCmd := bson.D{
+		{"createIndexes", colName},
+		{"indexes", bson.A{
+			bson.D{
+				{"key", bson.D{{"$**", "columnstore"}}},
+				{"name", "dump_columnstore_test"},
+				{"columnstoreProjection", bson.D{
+					{`"`, 0},
```
autarch, line 364: Is this really legal syntax for this? That's really odd. Even if it's legal, can we instead do `{ "_id", 1 }` so it's less surprising?

mongorestore/mongorestore_test.go
```diff
@@ -2063,6 +2063,8 @@ func TestRestoreTimeseriesCollections(t *testing.T) {
 type indexInfo struct {
 	name string
 	keys []string
+	// columnstoreProjection contains info of the columnstoreProjection key in columnstore indexes.
```
autarch, line 2066: nit: s/of the/about the/

mongorestore/mongorestore_test.go
```diff
@@ -2345,3 +2347,228 @@ func runMongodump(t *testing.T, dir, db, collection string, args ...string) stri
...
+	testDB := session.Database(dbName)
+	defer func() {
+		err = testDB.Drop(nil)
+		if err != nil {
+			t.Fatalf("Failed to drop test database: %v", err)
+		}
+	}()
+
+	key := "$**"
+	columnstoreProjection := map[string]int32{"price": 0}
```
autarch, line 2395: While `: 0` apparently works, all the doc examples I've seen for indexes use `: 1` so let's use that instead.

mongorestore/mongorestore_test.go
```diff
@@ -2345,3 +2347,228 @@ func runMongodump(t *testing.T, dir, db, collection string, args ...string) stri
...
```
autarch, line 2555: Let's use `: 1` here as well (and anywhere else I've missed).

autarch (approved):
LGTM % a few small things + Rohan's comments. No need to request another review once my comments are addressed.

mongodump/mongodump_test.go
tdq45gj: No, just updated the test to check all collections.

PR #520, opened 2023-02-21
TOOLS-3255: Fix qa-tests-3.4

tdq45gj:
This PR fixes test failures on server 3.4. This fix should make `IsAtlasProxy` method in `SessionProvider` backward compatible with server 3.4.

### DIFF
--- a/common/db/command.go
+++ b/common/db/command.go
@@ -145,7 +145,7 @@ func (sp *SessionProvider) IsAtlasProxy() (bool, error) {
 			return false, nil
 		}
 		// For server 3.4 and below.
-		if strings.Contains(result.Err().Error(), "no such cmd") {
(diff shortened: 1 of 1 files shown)


rohshar (commented):
Can you add a description to the PR? We generally want all PRs to have one.

tdq45gj:
> Can you add a description to the PR? We generally want all PRs to have one.

Just added a description.

rohshar (approved):

rohshar (changes requested):
Actually can you add comments to the code too, describing which if clause is for which version?

tdq45gj:
> Actually can you add comments to the code too, describing which if clause is for which version?

Not sure if we can find the exact server version. I added it in as a comment in the code even though it might not be very accurate.

rohshar (approved):

PR #525, opened 2023-02-24
TOOLS-3256: Remove push tasks

tdq45gj:
This PR removes all `push` tasks from `common.yml`

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -2229,8 +2229,6 @@ buildvariants:
       - name: "dist"
       - name: "sign"
         run_on: rhel80-small
-      - name: "push"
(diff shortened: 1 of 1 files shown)


tfogo (changes requested):
So we don't want to remove the push tasks. We still need those to run when we do a release. What we wan to do is make push only run on git tags. e.g.:

```
- name: "push"
  run_on: rhel80-small
  git_tag_only: true
```

See: https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#limiting-when-a-task-will-run

tfogo (approved):
LGTM

PR #544, opened 2023-03-31
TOOLS-3279: Fix segmentation faults when tests fail.

FGasper:
In several places we used t.Errorf (i.e., fail the test, but continue) when the intent appears to have been t.Fatalf (fail and stop the test). This changeset makes that substitution.

Also included is a bit more clarity in a few cases where we skip tests.


autarch (approved):
LGTM!

suspicious-mango (approved):
LGTM

suspicious-mango (approved):
LGTM. had to double check what `restore, err := getRestoreWithArgs(args...)` does and whether `defer restore.Close()` needs to be called before we fatally exit in mongorestore/mongorestore_test.go. The answer is I don't think so.

PR #551, opened 2023-04-10
TOOLS-2722: Add macOS ARM to Evergreen builds.

FGasper:


release/platform/platform.go
```diff
@@ -364,10 +364,18 @@ var platforms = []Platform{
 		BuildTags: defaultBuildTags,
 	},
 	{
-		Name:      "macos",
-		Arch:      ArchX86_64,
-		OS:        OSMac,
-		BuildTags: defaultBuildTags,
+		Name:        "macos",
+		VariantName: "macos-arm",
```
mmcclimon, line 368: The docs for VariantName in the Platform struct say setting it should only be done for special builds, and in general we should just use the Name+Arch pair. Unless I'm missing something, I think you should do that here, and remove VariantName.

(I was wondering here if you couldn't have duplicate names, but there are multiple ubuntu1804 names defined here, so I reckon it's ok to just do the normal thing for macos too.)

common-pvt.yml
```diff
@@ -225,17 +225,25 @@ buildvariants:
...
 
   #######################################
-  #           macOS Buildvariant        #
+  #         macOS Buildvariants         #
   #######################################
 
-  - name: macOS-1014-ssl
-    display_name: MacOS 10.14 SSL
+  - name: macOS-1014-x86_64-ssl
+    display_name: MacOS 10.14 x86_64 SSL
```
mmcclimon, line 232: (question) Why are we saying MacOS 10.14 here? The latest macOS versions can definitely run on x86_64, but I'm also not sure where we display this, or if we do special things that only work on versions prior to 10.15 (which seems unlikely to me).

release/platform/platform.go
```diff
@@ -119,7 +119,7 @@ func DetectLocal() (Platform, error) {
 		}
 		return pf, nil
 	case "Darwin":
-		pf, ok := GetByVariant("macos")
+		pf, ok := GetByVariant("macos-arm")
```
mmcclimon, line 122: With the caveat that I don't totally understand the build script: is this sufficient? (I don't _think_ so, but I'm not totally sure.) Presumably we still want to support building on x86 macs, and I don't think this would let that happen. (I think also this interacts with the other change I suggested, to use the name/arch pair.)

mmcclimon (changes requested):
I think this is not quite right for x86 mac builds, but I have not actually gone downstairs to fetch the one I have lying around so I can check for myself. (I don't have a go toolchain installed on it, though I could if it would be useful to do so!)

release/platform/platform.go
FGasper: It broke when I had just `macos`, I think because the `Name` has to match Evergreen. I’ll play with it a bit and see if I can get it to work without VariantName.

FGasper:
> I think this is not quite right for x86 mac builds, but I have not actually gone downstairs to fetch the one I have lying around so I can check for myself. (I don't have a go toolchain installed on it, though I could if it would be useful to do so!)

Looks like we need to go through this to test it properly: https://wiki.corp.mongodb.com/display/KERNEL/Access+ZAP+-+zSeries+%28s390x%29%2C+POWER+%28ppc64le%29%2C+ARM+%28arm64%29+and+MacOS

release/platform/platform.go
```diff
@@ -38,9 +38,9 @@ const (
 	// While arm64 and aarch64 are the same architecture, some Linux distros
 	// use arm64 (Debian and RHEL) and others use aarch64 (Amazon 2).
 	ArchAarch64 Arch = "aarch64"
-	ArchS390x        = "s390x"
-	ArchPpc64le      = "ppc64le"
-	ArchX86_64       = "x86_64"
+	ArchS390x   Arch = "s390x"
```
huan-Mongo, line 41: [nit] I think only have the first element typed is a more idiomatic way inside Golang such as 
```
const (
	ArchArm64 Arch = "arm64"
        ArchAarch64 = "aarch64"
        ArchS390x  = "s390x"
        ...
```

huan-Mongo (approved):
LGTM % nit comment

release/platform/platform.go
FGasper: That only works for `iota`, I think. Example: https://go.dev/play/p/IAbrnm9DkIf

mmcclimon: We should probably do the same thing with all these other constants (OS, Pkg, Repo), which have the same problem!

mmcclimon (approved):
LGTM % I think we should fix the typed constant issue. I'll approve anyway, but happy to take another look after if that turns out to have annoying knock-on effects (I can imagine it might require changing type signatures in some other places).

common-pvt.yml
tfogo: Yeah I think that's what happened. We set version compatibility here: https://github.com/mongodb/mongo-tools/blob/e62cf8866c5b98c45faf2fbe354703969aa076c6/set_goenv.sh#L24-L33

So nothing special for 10.14. We should upgrade if we can.

release/platform/platform.go
```diff
@@ -96,12 +94,21 @@ func GetFromEnv() (Platform, error) {
...
+
+	cmd := exec.Command("uname", "-sm")
 	out, err := cmd.Output()
 	if err != nil {
 		return Platform{}, fmt.Errorf("failed to run uname: %w", err)
 	}
-	kernelName := strings.TrimSpace(string(out))
+
+	kernelNameAndArch := strings.TrimSpace(string(out))
+	pieces := strings.Split(kernelNameAndArch, " ")
```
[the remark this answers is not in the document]
FGasper: `Fields` has its own quirks given that it uses unicode.IsSpace rather than ASCII spaces. But tabs and multiple spaces are a legit concern.

Fixed in 39a5fc70.

release/platform/platform.go
FGasper: @mmcclimon Fixed in 13a93d2a.

common-pvt.yml
FGasper: macOS 10.15 works, but 11.00 consistently [breaks](https://spruce.mongodb.com/version/643dac32850e61bd621796dc/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). I’ve left it at 10.15; I’ll open a separate investigation into why 11.00 fails.

tfogo (approved):
LGTM

PR #553, opened 2023-04-20
TOOLS-3052: Add builds for Amazon Linux 2023

FGasper:
Note that, due to the unavailability of MongoDB builds for this distribution, currently no tests run that require MongoDB (e.g., integration).


FGasper:
We’ll defer on this until there’s server support.

PR #560, opened 2023-05-16
Added support for no admin command error - restoring into DocumentDB

maorbril-coinbase:
When running restore against DocumentDB on a mongodump taken from Atlas.
This is Mongo5.0 restoring to DocumentDB 5.0

### DIFF
--- a/common/db/command.go
+++ b/common/db/command.go
@@ -144,6 +144,9 @@ func (sp *SessionProvider) IsAtlasProxy() (bool, error) {
 		if strings.Contains(result.Err().Error(), "CommandNotFound") {
 			return false, nil
 		}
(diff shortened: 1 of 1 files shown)


tfogo:
Hi @maorbril-coinbase, thank you for submitting a PR.

In order to support a use case of the database tools, we want to make sure we fully test it. We currently don't test against databases that attempt to be MongoDB-compatible such as CosmosDB or DocumentDB. Given that DocumentDB only partially implements the MongoDB API, it would take considerable work for us to test against it, document limitations, add guardrails, find workarounds, etc.

We don't have any plans to test against and support DocumentDB at this time. Because of this, I'm going to close this PR. 

Please feel free to maintain your own fork of the tools if you want to add any behavior that isn't officially supported.

Thanks,
Tim

PR #563, opened 2023-05-17
TOOLS-3062 Add Amazon Linux 2023 x86 builds.

FGasper:
Note that, due to the unavailability of MongoDB builds for this distribution, currently no tests run that require MongoDB (e.g., integration).

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -2682,6 +2682,39 @@ buildvariants:
         run_on: rhel80-small
         git_tag_only: true
 
+  - name: amazon2023
+    display_name: Amazon Linux 2023
+    run_on:
+      - amazon2023-small
+    stepback: false
+    expansions:
+      <<:
+        [
+          *mongod_ssl_startup_args,
+          *mongo_ssl_startup_args,
+          *mongod_tls_startup_args,
+          *mongo_tls_startup_args,
+        ]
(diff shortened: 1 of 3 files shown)


common.yml
```diff
@@ -2682,6 +2682,39 @@ buildvariants:
...
+      mongo_os: amazon2023
+      mongo_edition: enterprise
+      smoke_use_ssl: --use-ssl
+      resmoke_use_ssl: _ssl
+      smoke_use_tls: --use-tls
+      resmoke_use_tls: _tls
+      excludes: requires_mmap_available,requires_large_ram,requires_mongo_24,requires_mongo_26,requires_mongo_30
+      resmoke_args: -j 2
+      edition: "enterprise"
+      USE_SSL: "true"
```
mmcclimon, line 2707: Since we're not running mongo-requiring tasks here, I might get rid of this and replace it with a comment explaining why these tasks are different than the other Linux distros. (I could also see an argument for leaving them in place, but I think a comment would be useful in that case too.)

mmcclimon (changes requested):
A few questions; I don't feel particularly strongly about them so if you _do_, send it back and I can approve.

tfogo (approved):
LGTM

PR #566, opened 2023-05-24
TOOLS-3055 Create 100.7.1’s changelog entry.

FGasper:


CHANGELOG.md
```diff
@@ -1,5 +1,48 @@
 # Database Tools Changelog
 
+## 100.7.1
+
+_Released 2023-05-24_
+
+We are pleased to announce version 100.7.1 of the MongoDB Database Tools.
+
+This release adds several new builds and fixes a few bugs.
```
tfogo, line 9: I think we should add some more highlights here. The MacOS 11.0 ARM build is something that has been requested a lot so it might be nice to list the new platforms in the summary. The extra prominence might be good.

tfogo (changes requested):

CHANGELOG.md
FGasper: This is done.

tfogo (approved):
LGTM

PR #568, opened 2023-05-30
TOOLS-3309 Use `aarch64` for releases on RHEL platforms

tdq45gj:
Updated code comment in `platform.go` and changed arch names for releases on RHEL platforms.

### DIFF
--- a/release/platform/platform_test.go
+++ b/release/platform/platform_test.go
@@ -116,3 +116,18 @@ func TestPlatformsHaveNoDuplicates(t *testing.T) {
 		assert.Equal(t, 1, dupes, "platform %v only occurs once in platforms list", p1)
 	}
 }
+
+func TestPlatformCorrectArch(t *testing.T) {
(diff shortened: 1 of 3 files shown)


FGasper (changes requested):
Can we maybe add a unit test to catch this kind of mistake in the future? Something that detects its own platform and ensures that platform.go is correct for itself?

release/platform/platform_test.go
```diff
@@ -116,3 +116,18 @@ func TestPlatformsHaveNoDuplicates(t *testing.T) {
 		assert.Equal(t, 1, dupes, "platform %v only occurs once in platforms list", p1)
 	}
 }
+
+func TestPlatformCorrectArch(t *testing.T) {
```
FGasper, line 120: I was thinking something that would actually determine the local Linux distro and check that the arch is correct. But this probably suffices.

FGasper (approved):
LGTM

autarch (approved):
LGTM!

PR #575, opened 2023-06-16
[TOOLS-3319] Change mongorestore behavior to ignore admin collections on archive restore

rohshar:
Previously, we have made it so that dump and restore wouldn't touch the admin DB if connected to an atlas proxy in most cases. However, we missed the case in which archive is used in mongorestore from an export created from a non atlas proxy host (meaning that it would contain admin collections). This uses a special (and confusing) workflow, which needs its own handling to skip parts of the export.

I tested these changes with the failing atlas proxy tests: https://spruce.mongodb.com/version/6492223b32f41764587b254b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC, and this all should work.

### DIFF
--- a/common/archive/demultiplexer.go
+++ b/common/archive/demultiplexer.go
@@ -64,6 +64,11 @@ func CreateDemux(namespaceMetadatas []*CollectionMetadata, in io.Reader, isAtlas
 		IsAtlasProxy:    isAtlasProxy,
 	}
 	for _, cm := range namespaceMetadatas {
+		if isAtlasProxy && cm.Database == "admin" {
+			log.Logvf(log.Info, "skipping creating namespace status for %s.%s", cm.Database, cm.Collection)
+			continue
+		}
+
 		var ns string
 		if cm.Type == "timeseries" {
 			ns = cm.Database + ".system.buckets." + cm.Collection
--- a/common/archive/demultiplexer.go
+++ b/common/archive/demultiplexer.go
@@ -128,13 +128,13 @@ func (demux *Demultiplexer) HeaderBSON(buf []byte) error {
 	if colHeader.Collection == "" {
 		return newError("collection header is missing a Collection")
 	}
-	demux.currentNamespace = colHeader.Database + "." + colHeader.Collection
 
 	if demux.IsAtlasProxy && colHeader.Database == "admin" {
 		log.Logvf(log.Info, "skipping consuming archive entry for %v", demux.currentNamespace)
 		return nil
 	}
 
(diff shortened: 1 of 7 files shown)


autarch (approved):
LGTM (I don't love making this change in so many places, but I think any sort of major refactoring of the tools code is out of scope).

rohshar:
I will say that this was a nightmare to investigate and try to figure out.

rohshar:
Until I find a better place to document the process of running atlas proxy tests with unreleased versions of mongodump and mongorestore, I will do so here:

1. Add whatever changes you need to the mongo-tools code.
2. Add changes to the release infrastructure as follows: https://github.com/mongodb/mongo-tools/pull/575/commits/7eb24e7ee3d5913021f2e52867dbca2b407dc550
3. Kick off mongo-tools push patches: **evergreen patch -p mongo-tools -y -v all -t "push" -f**. The changes in the previous step made it so that push will work with patches, and the infra is already set up to only place them in the unstable file locations.
4. Make changes to the atlas proxy code to turn up the verbosity and download / unpack the correct unstable binaries: https://github.com/rohshar/atlasproxy/pull/1/files. The additional change you'll probably have to make is to sub out the **gb462ee7** with **g<7-digits-of-base-commit>**.
5. Run an atlas proxy patch **evergreen patch -p atlasproxy -y -v all -t "RunBackupIntegrationTests" -f**. RunBackupIntegrationTests are the tests that use mongodump and mongorestore.
6. Check the results. Only the RHEL 7 tasks will run correctly, the AL2 tasks need a different number of commit hash digits specified in step 4. An example patch: https://spruce.mongodb.com/version/6493565ae3c331b7138a3fee/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

tfogo (approved):
LGTM

common/archive/demultiplexer.go
```diff
@@ -217,6 +230,11 @@ func (demux *Demultiplexer) BodyBSON(buf []byte) error {
 		return newError("collection data without a collection header")
 	}
 
+	// For atlas proxy archive restores, ignore collections from the admin DB.
+	if demux.IsAtlasProxy && strings.Contains(demux.currentNamespace, "admin") {
+		return nil
+	}
+
```
pierre-gautier, line 237: it seems this line also ignores all collections containing "admin", is it really intended ?

pierre-gautier: @rohshar @tfogo @autarch

PR #579, opened 2023-06-26
TOOLS-3324 Dump entire config db when --db=config

tfogo:
If a user explicitly specifies `--db=config` then we provide the entire collection database. 

We continue to skip most config collections by default.

### DIFF
--- a/mongodump/prepare_test.go
+++ b/mongodump/prepare_test.go
@@ -144,8 +144,49 @@ func TestShouldSkipSystemNamespace(t *testing.T) {
 			output: true,
 		},
 		{
+			db:     "config",
+			coll:   "chunks",
+			output: false,
+		},
+		{
+			db:     "config",
+			coll:   "collections",
+			output: false,
+		},
+		{
+			db:     "config",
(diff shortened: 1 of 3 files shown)


mongodump/prepare_test.go
```diff
@@ -142,6 +143,12 @@ func TestShouldSkipSystemNamespace(t *testing.T) {
 			coll:   "foo",
 			output: true,
 		},
+		{
+			db:       "config",
+			coll:     "foo",
```
rohshar, line 148: Can we also add a test for a collection that would be included normally like `chunks` in ConfigCollectionsToKeep.

rohshar (commented):
Can you add a comment to config.go that explains the behavior of dumping/restoring config DB collections based on passed in options?

mongodump/prepare_test.go
tfogo: Added

tfogo:
Good calls! made those changes @rohshar

rohshar (approved):

PR #586, opened 2023-07-13
TOOLS-3338 Fix issue where mongorestore ignored any namespace including "admin"

jddubois:
When using Atlas Proxy, all collections from the `admin` database should be ignored. This changes fixes an issue with the [initial implementation](https://github.com/mongodb/mongo-tools/pull/575) of this feature that resulted in any namespace including `"admin"` being excluded.

### DIFF
--- a/mongorestore/mongorestore_archive_test.go
+++ b/mongorestore/mongorestore_archive_test.go
@@ -266,7 +266,7 @@ func requireCollectionHasNumDocuments(t *testing.T, collection *mongo.Collection
 	require := require.New(t)
 	count, err := collection.CountDocuments(context.Background(), bson.M{})
 	require.NoError(err, "can count documents")
-	require.EqualValues(0, count, "found %d document(s)", numDocuments)
+	require.EqualValues(numDocuments, count, "found %d document(s)", count)
 }
 
 func createCollectionWithTestDocument(t *testing.T, db *mongo.Database, collectionName string) *mongo.Collection {
--- a/mongorestore/mongorestore_archive_test.go
+++ b/mongorestore/mongorestore_archive_test.go
@@ -35,9 +35,14 @@ func init() {
 }
 
 var (
-	testArchive          = "testdata/test.bar.archive"
-	testArchiveWithOplog = "testdata/dump-w-oplog.archive"
-	testBadFormatArchive = "testdata/bad-format.archive"
+	testArchive                          = "testdata/test.bar.archive"
+	testArchiveWithOplog                 = "testdata/dump-w-oplog.archive"
+	testBadFormatArchive                 = "testdata/bad-format.archive"
+	adminPrefixedCollectionName          = "admins"
+	nonAdminCollectionName               = "test"
+	adminAndPeriodPrefixedCollectionName = "admin.test"
+	adminDBName                          = "admin"
+	adminSuffixedDBName                  = "testadmin"
 )
 
 func TestMongorestoreShortArchive(t *testing.T) {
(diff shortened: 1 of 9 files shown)


mongorestore/mongorestore_archive_test.go
```diff
@@ -132,3 +138,174 @@ func TestMongorestoreBadFormatArchive(t *testing.T) {
...
+func TestMongorestoreArchiveAdminNamespaces(t *testing.T) {
+	require := require.New(t)
+
+	testtype.SkipUnlessTestType(t, testtype.IntegrationTestType)
+
+	session, err := testutil.GetBareSession()
+	require.NoError(err, "can connect to server")
+
+	fcv := testutil.GetFCV(session)
+	if cmp, err := testutil.CompareFCV(fcv, "4.2"); err != nil || cmp < 0 {
```
jddubois, line 157: Looks like these tests aren't compatible with the MMAPv1 storage engine- I think setting this to 4.2 is OK because atlas shared clusters only run 6.0?

tfogo: Yeah I think that's fine. What was the error you saw though?

jddubois:
```
error demultiplexing archive; archive finished but contained files were unfinished ([admin.system.indexes])
```

IamXander (changes requested):
This seems totally right! Nice work!

mongorestore/mongorestore_archive_test.go
tfogo: Hmmm I'm not sure how that's specific to MMAPv1. Can you link me to the failing tests on evergreen?

jddubois: Here's an [example](https://spruce.mongodb.com/task/mongo_tools_amazon_integration_3.4_mmapv1_patch_369ced42d4aad50c1036534f3a541051f7618863_64b0513c57e85aa71dcf0dc9_23_07_13_19_32_13/tests?execution=0&sortBy=STATUS&sortDir=ASC)

IamXander (approved):
I love to see that you are adding more testing!

rohshar (approved):
lgtm as long as only running the test on v4.2 or higher is resolved.

tfogo:
@jddubois So it's not the fact that your test is failing on MMAPv1. It's failing everywhere. But there is some bug with the testing infra that means that all integration tests (except for the mmap tests) are actually getting skipped because `TOOLS_TESTING_INTEGRATION` is not set correctly. So although it looks like the tests are passing but actually they're just getting skipped.

Bug was probably introduced by #290. The `TOOLS_TESTING_INTEGRATION` environment variable should be set here: https://github.com/mongodb/mongo-tools/blob/master/buildscript/build.go#L132 

Could you take a couple of hours to look into this to see if you can figure out what's going wrong? If you need help let me know. If you can't see what's wrong, don't spend too much time on it, let me know and we can get somebody more acquainted with the testing to help out.

jddubois:
For visibility- @tfogo discovered that the reason tests weren't actually running is caused by `test:srv` being called directly after `test:integration` which led to the integration results being overwritten by the `srv` results. Because there was only one srv test, I moved it into the unit tests and removed the `test:srv` calls from all integration evergreen tasks.

tfogo (approved):
LGTM! Good job on this. I think there's no need to create a ticket for the mmap issue. It should be a very rare edge case that isn't supported anyway so hopefully no user should run into it.

PR #587, opened 2023-07-14
TOOLS-3344 Fix issue where mongodump failed against Atlas Online Archive and AWS DocumentDB

jddubois:
Fixes issues where `mongodump` crashed when run against Atlas Online Archive or AWS DocumentDB by handling their errors properly. This will still break if any of these error messages change in the future.

I tested by running my local `mongosync` binary against an online archive cluster directly, and against an AWS DocumentDB instance tunneled through EC2 (instructions [here](https://docs.aws.amazon.com/documentdb/latest/developerguide/connect-from-outside-a-vpc.html)).

### DIFF
--- a/common/db/command.go
+++ b/common/db/command.go
@@ -140,18 +140,34 @@ func (sp *SessionProvider) IsAtlasProxy() (bool, error) {
 		context.Background(),
 		&bson.M{"atlasVersion": 1},
 	)
-	if result.Err() != nil {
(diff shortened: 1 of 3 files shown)


tfogo (changes requested):
This seems fragile since it depends on errors not changing. Why don't we just assume it's not Atlas if the command errors at all? Probably a safe bet and it's more robust.

tfogo (approved):
LGTM!

PR #593, opened 2023-08-01
TOOLS-3349 Add integration tests for config db dump and restore

jddubois:
This change adds integration tests to verify the behavior of `mongodump` and `mongorestore` on collections in the `config` database.

### DIFF
--- a/mongorestore/mongorestore_test.go
+++ b/mongorestore/mongorestore_test.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/mongodb/mongo-tools/common/db"
+	"github.com/mongodb/mongo-tools/common/dumprestore"
 	"github.com/mongodb/mongo-tools/common/log"
 	"github.com/mongodb/mongo-tools/common/options"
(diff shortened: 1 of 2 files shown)


mongorestore/mongorestore_test.go
```diff
@@ -2580,3 +2608,122 @@ func testRestoreColumnstoreIndexFromOplog(t *testing.T) {
...
+			require.EqualValues(0, result.Failures, "mongorestore reports 0 failures")
+
+			for _, collection := range collections {
+				r := collection.FindOne(context.Background(), testDocument)
+				require.NoError(r.Err(), "expected document")
+			}
+
+		},
+		"--db", "config",
+		"--excludeCollection", "transactions",
```
jddubois, line 2665: I had to exclude transactions here because we get an error message when attempting to clear them out, and we get a duplicate key error on restore when attempting to restore. I'm not sure how else to get around this- any other ideas?

tfogo (approved):
LGTM

PR #596, opened 2023-08-07
TOOLS-3365 Prep release for 100.7.5

tfogo:
Adds a new build target `checkMinVersion` which checks the version of Go being used. It will error if it is not at least the minimum version set in `build.go`. `checkMinVersion` is run during the `dist` task before binaries are built.

Removes support for PPC Ubuntu 18.04 since it is no longer supported as of 4.2.17 and 4.4.10. Evergreen builds are no longer run for this distro.

Changed the SUSE 15 build host to suse15sp4 since this is the recommended distro and has go 1.19.12.

Reverts #582 since a bug in goconvey was causing test failures.


tfogo:
Looks like suse15sp4 has some weirdness with the python environment for the latest versions of the server. Just going to ignore that for now though and will fix it during the 7.0 support work.

buildscript/build.go
```diff
@@ -27,6 +30,35 @@ var pkgNames = []string{
...
+// minimumGoVersion must be prefixed with v to be parsed by golang.org/x/mod/semver
+var minimumGoVersion = "v1.19.12"
+
+func CheckMinimumGoVersion(ctx *task.Context) error {
+	goVersionStr, err := runCmd(ctx, "go", "version")
+	if err != nil {
+		return fmt.Errorf("failed to get current go version: %w", err)
+	}
+
+	_, _ = ctx.Write([]byte(fmt.Sprintf("Found Go version \"%s\"\n", goVersionStr)))
```
rohshar, line 42: This is just to log to stdout?

rohshar (approved):

buildscript/build.go
tfogo: Yep, seems to be the way to do it in goke so it's formatted correctly.

PR #601, opened 2023-08-16
Fix error when compiling for Debian ARM64

baschny:
Trying to compile under docker with debian11 under ARM, error is:
```
panic: ubuntu1804 platform name changed
```

I provide this patch on a "works for me" basis, as I have no clue how this platform.go works or is supposed to work.

### DIFF
--- a/release/platform/platform.go
+++ b/release/platform/platform.go
@@ -122,9 +122,9 @@ func DetectLocal() (Platform, error) {
 
 	switch kernelName {
 	case "Linux":
-		pf, ok := GetByOsAndArch("ubuntu1804", archName)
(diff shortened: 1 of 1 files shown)


baschny:
Please disregard this PR, this does not really solve the issue. The main problem seems to be the logic with the "platform detection" based on the currently running kernel version.

When compiling with Docker the kernel could be not "original one" which comes from the distribution you are compiling for. For example if I cross-compile on my Mac using the ARM64 Debian 11 images, I get an `uname -sm` of:
```
Linux aarch64
```

It seems that the code is trying to detect the distribution using `uname -sm` which doesn't seem logical.

Maybe consider using https://github.com/zcalusic/sysinfo to detect current distribution?

PR #605, opened 2023-08-22
TOOLS-3371 Add 7.0 tests and remove superfluous linux tests

tfogo:
- Removes tests from all platforms except:
    - RHEL 8.0
    - Windows
    - RHEL 6.2 w/o SASL or Kerberos
- Adds 7.0 tests
- Removes old rapid release tests
- Removes `test/legacy42/jstests/tool/gridfs.js`. This was causing a setup failure for the sharded cluster, but the test is covered by `test/qa-tests/jstests/files/*` so it is not necessary.


pbrissaud:
Hello, 
What is the ETA for Mongo 7 support? It was released almost two months ago, and we don't have a working dumping tool yet...

tfogo:
@pbrissaud #607 will add official 7.0 support. I expect we will cut a new release next week.

That branch is really just adding the necessary tests for 7.0. The tests passed without any changes to the tools so you should be able to use 100.8.0 just fine with 7.0 if you need.

Apologies for the delay. Unfortunately there was a lot of tech debt in our testing we needed to sort out.

FYI I'll be closing this PR in favor of #607.

PR #607, opened 2023-09-26
TOOLS-3370 TOOLS-3371

tfogo:
Sorry for the large PR. This PR does several things:

- Remove tests for all variants except for:
    - Windows
    - RHEL 8
    - RHEL 6 w/o SASL/Kerberos (Splunk build)
- Add tests for 7.0.
- Replace the python script for downloading server and shell versions with Go code.
- Download the correct `jstestshell` for each version.
- Upgrade to Go 1.20.
- Run `gofmt` with 1.20 which changes indentation on several comment blocks.
- Make the `h` field `omitempty` when marshaling oplog entries in mongodump. This is because the `h` field was removed in 7.1 since it was unused for many versions. Also proactively strip `h` fields when running `applyOps` for 7.0+.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/mongodb/mongo-tools
 
-go 1.19
+go 1.20
```
mmcclimon, line 3: I think `minimumGoVersion` in buildscript/build.go also needs to be updated accordingly.

release/version/version.go
```diff
@@ -34,23 +30,26 @@ func Parse(desc string) (Version, error) {
...
 
 	maj, err := strconv.Atoi(parts[0])
 	if err != nil {
 		return Version{}, fmt.Errorf("failed to parse major version %q", parts[0])
 	}
 
-	min, err := strconv.Atoi(parts[1])
-	if err != nil {
-		return Version{}, fmt.Errorf("failed to parse minor version %q", parts[1])
+	min := 0
```
mmcclimon, line 39: (nit) `min` is a builtin in go1.21. It's fine to use it (it works), but it's maybe confusing because editors will highlight it differently; I might use `minor` here instead.

release/evergreen/evergreen.go
```diff
@@ -100,6 +100,43 @@ func GetArtifactsForTask(id string) ([]Artifact, error) {
 	return task.Artifacts, nil
 }
 
+// GetTasksForVersion gets all the evergreen tasks associated with a version.
```
mmcclimon, line 103: This comment is copy-pasted incorrectly.

mongorestore/oplog.go
```diff
@@ -210,6 +210,14 @@ func (restore *MongoRestore) HandleNonTxnOp(oplogCtx *oplogContext, op db.Oplog)
 		return fmt.Errorf("error filtering UUIDs from oplog: %v", err)
 	}
 
+	// The h field was removed in 7.1.0-rc0 as part of SERVER-69062
+	if restore.serverVersion.GTE(db.Version{7, 0, 0}) {
```
mmcclimon, line 214: Based on your comments elsewhere I _suspect_ it won't matter, but is it safe to strip the `h` fields even for 7.0 patch versions? This will do so for 7.0.1, for example.

release/release.go
```diff
@@ -55,51 +58,104 @@ func main() {
...
+					return nil
+				},
+			},
+			{
+				Name: "get-version",
+				Action: func(cCtx *cli.Context) error {
+					v, err := version.GetCurrent()
+					if err != nil {
+						log.Fatalf("Failed to get current version: %v", err)
+					}
```
mmcclimon, line 84: (nit) You could just return err here and the other places you log.Fatal from an Action, but it hardly matters.

release/release.go
```diff
@@ -1297,3 +1353,180 @@ func linuxRelease(v version.Version) {
...
+		fmt.Printf("extracting to: %v\n", tempDir)
+		untargz(tempPath, tempDir)
+	}
+
+	binFiles, err := filepath.Glob(path.Join(tempDir, "mongodb-*", "bin", "*"))
+
+	for _, f := range binFiles {
+		if filepath.Ext(f) != ".pdb" {
+			fmt.Printf("Move %s to %s\n", f, filepath.Join("bin", filepath.Base(f)))
+			os.Rename(f, filepath.Join("bin", filepath.Base(f)))
```
mmcclimon, line 1421: Should we check the error here?

mmcclimon (changes requested):
In general this looks good, and I am always happy to see big hunks of YAML deleted. I didn't read the common.yml changes _super_ closely, because if evergreen is happy than mostly I am too.

Just a few things that need tweaking, I think!

mongorestore/oplog.go
tfogo: Yeah I wasn't 100% sure how 7.1.0-rc0 would evaluate so I just did this since it doesn't matter.

tfogo:
@mmcclimon I addressed your comments.

release/release.go
```diff
@@ -1297,3 +1353,181 @@ func linuxRelease(v version.Version) {
...
+
+	fmt.Printf("extension: %v\n", filepath.Ext(filename))
+
+	switch filepath.Ext(filename) {
+	case ".zip":
+		fmt.Printf("extracting to: %v\n", tempDir)
+		unzip(tempPath, tempDir)
+	case ".tgz":
+		fmt.Printf("extracting to: %v\n", tempDir)
+		untargz(tempPath, tempDir)
```
tdq45gj, line 1413: Maybe we can add a case "default" and error out.

release/release.go
```diff
@@ -1297,3 +1353,181 @@ func linuxRelease(v version.Version) {
...
+
+	url, githash, serverVersion, err := feed.FindURLHashAndVersion(v, pf.Name, string(pf.Arch), "enterprise")
+	check(err, "get URL from JSON feed")
+
+	fmt.Printf("URL: %v\n", url)
+	fmt.Printf("GitHash: %v\n", githash)
+	fmt.Printf("Version: %v\n", serverVersion)
+
+	downloadBinaries(url)
+	if semver.Compare(fmt.Sprintf("v%s", serverVersion), "v6.0.0") > -1 {
```
tdq45gj, line 1384:
```suggestion
	if semver.Compare(fmt.Sprintf("v%s", serverVersion), "v6.0.0") >= 0 {
```
nit: a little bit hard to read.

tdq45gj (approved):
LGTM % some nits.

release/release.go
tfogo: Good idea

release/release.go
tfogo: Also good suggestion

tfogo:
@tdq45gj I made the changes you suggested

mmcclimon (approved):
LGTM, thanks!

PR #613, opened 2023-10-18
disable some writeConcern tests and updated minimumGoVersion

Becky-Hu-82:
see title


tfogo (approved):
LGTM. You can run the patch to check dist on rhel 6 or just merge and see if it fails on the waterfall.

PR #618, opened 2023-11-06
TOOLS-3411 Fix issue where sort order of views was not always preserved on dump

jddubois:
Fixes an issue where `mongodump` did not always preserve the order of options for a view or collection. This could result in different sort orders than expected on views. It is easily fixed by moving from the unordered `bson.M` type to the ordered `bson.D` type for collection options. Index options are still a `bson.M`, but the keys are `bson.D` so they are not affected by this issue.

I tested manually by viewing the metadata JSON files produced by `mongodump` on both this branch and master. I also verified that the added test succeeds every time on this branch, but fails most of the time on master.

### DIFF
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -12,7 +12,7 @@ import (
 	"crypto/sha1"
 	"encoding/base64"
 	"fmt"
-	"github.com/stretchr/testify/require"
+	"io"
 	"io/ioutil"
 	"math/rand"
 	"os"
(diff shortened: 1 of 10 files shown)


mongodump/mongodump_test.go
```diff
@@ -2091,3 +2104,86 @@ func TestMongoDumpColumnstoreIndexes(t *testing.T) {
...
+			{Key: "_id", Value: 0},
+		}}},
+		bson.D{{Key: "$sort", Value: bson.D{
+			{Key: "year", Value: 1},
+			{Key: "sID", Value: -1},
+			{Key: "name", Value: 1},
+			{Key: "score", Value: 1},
+		}}},
+	}
+
```
tdq45gj, line 2138: Given that the original bug of reordering is probabilistic, can we shuffle the pipeline in this test to test different orderings?

tfogo (approved):
LGTM

tdq45gj (approved):
Thanks, LGTM

PR #621, opened 2023-11-13
TOOLS-3302 Add Support for Debian 12

tdq45gj:
This PR adds Database Tools support for Debian 12 and prepares a release on Debian 12. It also adds the logic to download server versions from Evergreen server release tasks if URL is not available in JSON feed.

Update: I updated the script to download mongod/mongos. Now most integration tests pass % one system failure
https://spruce.mongodb.com/version/6552e6162a60ed0c9ea671a9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


etc/repo-config.yml
```diff
@@ -461,6 +461,17 @@ repos:
     repos:
       - apt/debian/dists/bullseye/mongodb-enterprise
 
+  - name: debian12
+    type: deb
+    edition: enterprise
+    code_name: "bookworm"
+    bucket: repo.mongodb.com
```
tfogo, line 468: Do we need to add a community repo too?

release/platform/platform.go
```diff
@@ -388,6 +388,14 @@ var platforms = []Platform{
 		Repos:     []Repo{RepoEnterprise, RepoOrg},
 		BuildTags: defaultBuildTags,
 	},
+	{
+		Name:      "debian12",
+		Arch:      ArchX86_64,
+		OS:        OSLinux,
+		Pkg:       PkgDeb,
+		Repos:     []Repo{RepoEnterprise, RepoOrg},
```
tfogo, line 396: Seems like we do if there is RepoOrg here?

tfogo (changes requested):
I'm worried that we're not running any tests due to full.json not having the info. Can you ask in #server-release why that's the case? Surely it should be updated if the server has been published.

release/platform/platform.go
tdq45gj: Yes, added.

tdq45gj:
> I'm worried that we're not running any tests due to full.json not having the info. Can you ask in #server-release why that's the case? Surely it should be updated if the server has been published.

@tfogo I asked in [this thread](https://mongodb.slack.com/archives/C1XQ502TA/p1699909842334009). It looks like we'll need to wait for the next RCs and GAs to test on Debian 12. But I'll check tomorrow to see if the released versions will be added. If not, are we okay with releasing without running integration tests?

tfogo (approved):
LGTM!

jddubois (approved):
I don't feel entirely confident in my review here since this area of the codebase is new to me, but there isn't anything that sticks out!

PR #634, opened 2024-03-01
DEVPROD-2553: migrate to garasign for signing

dylrich:
This commit removes our use of the legacy Notary service in favor of Garasign for both PGP signing as well as Authenticode signing. MacOS artifacts are still signed with MacOS Notary. Our signing code has been moved to a single script that branches based on the operating system we are signing for. Notably, this commit has removed signing for our RPMs. This was done because we were signing those RPMs with a now-EOL server key, which leads me to believe that no one was actually verifying these signatures in practice. Users would have had to magically know which key to use, and we have not be publicizing this information anywhere. If removal of RPM signing breaks someone's workflow, it can be added back in, though users can get signed RPMs directly from repo.mongodb.{com|org} instead.


rohshar:
This approach LGTM. Let's add the variables and kick off some patches to verify that the signing works.

rohshar (approved):
Took a look at some artifacts, and they look good to me.

PR #639, opened 2024-03-27
TOOLS-3499 Remove all tests and config for the long obsolete sasl and ssl build tags

autarch:
This also removes some internal code that checked for a `BuiltWithSSL` var that was always true.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -1858,7 +1858,7 @@ buildvariants:
 
  # This build is specifically for Splunk
   - name: rhel62-no-kerberos
-    display_name: "# RHEL 6.2 w/o SASL or Kerberos"
+    display_name: "# RHEL 6.2 w/o Kerberos"
     expansions:
       <<:
         [
--- a/common.yml
+++ b/common.yml
@@ -1856,8 +1856,8 @@ buildvariants:
         run_on: rhel80-small
         git_tag_only: true
 
-  # This build is specifically for Splunk
-  - name: rhel62-no-sasl-or-kerberos
+ # This build is specifically for Splunk
+  - name: rhel62-no-kerberos
     display_name: "# RHEL 6.2 w/o SASL or Kerberos"
     expansions:
       <<:
@@ -2296,7 +2296,7 @@ github_pr_aliases:
     task: "integration-4.2"
   # Run a subset of integration tests against the 4.4 version of MongoDB
   # Server on all variants where that is the most recent supported version.
-  - variant: "^rhel62-no-sasl-or-kerberos$"
+  - variant: "^rhel62-no-kerberos$"
     task: "integration-4.4"
   # Run a subset of integration tests against the 5.0 version of MongoDB
   # Server on all variants where that is the most recent supported version.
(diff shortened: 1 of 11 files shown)


common.yml
```diff
@@ -1856,8 +1856,8 @@ buildvariants:
         run_on: rhel80-small
         git_tag_only: true
 
-  # This build is specifically for Splunk
-  - name: rhel62-no-sasl-or-kerberos
+ # This build is specifically for Splunk
+  - name: rhel62-no-kerberos
     display_name: "# RHEL 6.2 w/o SASL or Kerberos"
```
edobranov, line 1861:
```suggestion
    display_name: "# RHEL 6.2 w/o Kerberos"
```

edobranov (approved):
LGTM

installer/rpm/mongodb-database-tools.spec
```diff
@@ -8,7 +8,7 @@ Group: Applications/Databases
 License: Apache
 URL:        http://www.mongodb.com
 Vendor:     MongoDB
-Requires: openssl, cyrus-sasl, cyrus-sasl-plain, cyrus-sasl-gssapi
+Requires: cyrus-sasl, cyrus-sasl-plain, cyrus-sasl-gssapi
```
tfogo, line 11: So we're removing the sasl tag which would link to libsasl, but we still require these sasl libs, is that right?

tfogo:
This looks correct to me, but I also don't know much about rpm or msi specs, or exactly what we should expect to be linking to. Could you add a reviewer from the Go driver team? They might have a better idea of what we should be linking to. If you can find somebody who is able to review the rpm or msi specs, that would be nice too.

I trust that this is okay, I just want to be very careful with changes like this.

installer/rpm/mongodb-database-tools.spec
autarch: I was trying to be conservative here because I'm  not sure if we really need them. But looking at this more closely I suspect we can remove them.

autarch: I realized that I could test this. I ran a RHEL 8 Docker image locally and installed the tools, along with all of the above packages (including `openssl`). Then I forcibly deleted all the files for these libraries. The installed tools binaries continued to run without any linker issues. So I will remove all of these dependencies.

autarch:
> This looks correct to me, but I also don't know much about rpm or msi specs, or exactly what we should expect to be linking to. Could you add a reviewer from the Go driver team? They might have a better idea of what we should be linking to. If you can find somebody who is able to review the rpm or msi specs, that would be nice too.

I asked and it was confirmed that the Go driver no longer links to OpenSSL or SASL libs. I think that plus my test with a Docker image makes me comfortable with the RPM spec change. I'm asking to see if I can find someone who understands MSIs.

autarch:
Someone here at MongoDB who understands MSIs took a look at this PR and said it looked fine.

I also ran a patch to build the MSI and installed it on an Evergreen host. I confirmed that A) `libsasl.dll` is not present after it's installed; and B) the tools binaries run without issue.

I think for the release notes we should recommend that people delete any `libsasl.dll` installed alongside older versions of the tools.

PR #646, opened 2024-05-08
TOOLS-3519 Document the need to disable writes when dumping/restoring via oplog.

FGasper:


mongodump/options.go
```diff
@@ -57,7 +57,7 @@ func (inputOptions *InputOptions) GetQuery() ([]byte, error) {
 type OutputOptions struct {
 	Out                        string   `long:"out" value-name:"<directory-path>" short:"o" description:"output directory, or '-' for stdout (default: 'dump')"`
 	Gzip                       bool     `long:"gzip" description:"compress archive or collection output with Gzip"`
-	Oplog                      bool     `long:"oplog" description:"use oplog for taking a point-in-time snapshot"`
+	Oplog                      bool     `long:"oplog" description:"for taking a point-in-time snapshot. For sharded clusters, you MUST disable writes first!"`
```
mvankeulen94, line 60: I think we can instead say "for taking a point-in-time snapshot on replica sets that are not part of sharded clusters".
We may have to update this later as part of tools compatibility with invisible sharding.

mongorestore/options.go
```diff
@@ -51,7 +51,7 @@ const (
 // InputOptions defines the set of options to use in configuring the restore process.
 type InputOptions struct {
 	Objcheck               bool   `long:"objcheck" description:"validate all objects before inserting"`
-	OplogReplay            bool   `long:"oplogReplay" description:"replay oplog for point-in-time restore"`
+	OplogReplay            bool   `long:"oplogReplay" description:"replay oplog for point-in-time restore. For sharded clusters, you MUST disable writes first!"`
```
mvankeulen94, line 54: ditto above comment

mvankeulen94 (changes requested):
thank you for taking this on! I have some alternate wording suggestions

mongodump/options.go
FGasper: The JIRA ticket says:
> I've seen users attempt to use --oplog in multi-shard environments to back up their clusters without pausing writes. We should update the docs for mongodump and also for mongoreplay --oplogReplay so we don't endorse this practice anymore. 

I understood “this practice” to mean doing a backup/restore against a sharded cluster without pausing writes. But it sounds from your suggestion like we just want to forbid use against sharded clusters in any circumstance. Is that the case?

mvankeulen94: It's more because the benefit of --oplog mode is to allow capturing any writes that occur during the backup. With sharded clusters, we recommend pausing writes in any case, so there isn't much point in using --oplog.

FGasper: OK. So should we _fail_ attempts to use `--oplog` against sharded clusters?

mvankeulen94: We currently fail to execute --oplog when run on a mongos, but we don't stop users from running --oplog on individual shards. We could consider doing this for the future, but that would break user workflows, since unfortunately users are doing this today (without pausing writes). For now I'm hoping we can proceed with a docs change. I'm planning some coordinated messaging to guide people away from these workflows for the future.

mvankeulen94 (changes requested):
thanks for the changes! I've added a couple more comments

mongodump/options.go
FGasper: Is it worth warning if the replica set in question is a shard? That way no one’s workflow suffers, but we better maximize the # of users who will see the guidance.

mvankeulen94: IMO it's preferable to keep the change as docs-only to avoid further complexity.

mvankeulen94 (approved):
looks great, thanks!

PR #650, opened 2024-05-09
TOOLS-3535 Add gosec as a linter and generate a SARIF report as part of release

jddubois:
This PR adds infrastructure to install dev tools including `precious`, which we use to run `gosec` as a linter. As part of the release process, this will generate a SARIF report. If there are new issues in the code base, then the release will be blocked.

It also updates the `lint-go` Evergreen task to run our new linting, which right now will just run `gosec`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *.zip
 
 !**/testdata/**
+/dev-bin
 
 /mongorestore/testdata/txntest/oplog.bson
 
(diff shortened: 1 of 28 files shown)


autarch:
Here's an example of a SARIF report containing issues: https://parsley.mongodb.com/evergreen/mongo_tools_sarif_gosec_SARIF_report_patch_170b6f2557fa659b74e8268afcfaadd7c8d608df_664c34b2ceb3230007aa18a8_24_05_21_05_44_19/0/task?bookmarks=0,3737&selectedLineRange=L91

PR #651, opened 2024-05-10
Fix lint-js and SRV test failures and clean up options testing code

autarch:
This PR fixes an issue with the Evergreen config for our lint-js testing.

It fixes the SRV tests by mocking DNS so we no longer rely on a particular cluster existing in Atlas when the tests are run. Since these tests are no longer special, the `test:srv` build target is gone and these tests are just run as normal unit tests.

It also cleans up the options testing code to use testify.


autarch:
Argh, I thought I fixed the `srv` failure but now it's back :(

autarch:
@tfogo Is this change to mock DNS in the SRV tests ok? I think we're still testing the same stuff, but I wanted to run it by you.

FGasper (approved):
LGTM. Thanks!

tfogo (approved):
Nice, I think mocking it works. LGTM!

jddubois (approved):

PR #653, opened 2024-05-20
TOOLS-3534 Run `snyk test` in CI as part of release

autarch:


autarch:
See https://parsley.mongodb.com/evergreen/mongo_tools_snyk_test_check_snyk_results_patch_170b6f2557fa659b74e8268afcfaadd7c8d608df_664ad2a47af71b0007206ebc_24_05_20_04_33_42/0/task?bookmarks=0,99&shareLine=69 for an example of what this looks like when it runs.

common.yml
```diff
@@ -660,6 +661,24 @@ tasks:
...
+    commands:
+      - func: "fetch source"
+      - command: expansions.update
+      - command: shell.exec
+        params:
+          working_dir: src/github.com/mongodb/mongo-tools
+          script: |
+            ${_set_shell_env}
+            set -e
+            curl -o ./snyk https://static.snyk.io/cli/latest/snyk-linux
```
huan-Mongo, line 678: I think in the long term we should consider store this file internally.

autarch: Do you mean store it in the repo? This isn't FOSS, so I wouldn't want to put it in this repo.

huan-Mongo: I am not thinking about storing it in a repo. Maybe a permanent S3 storage? Maybe we can ask the DevProd team for idea.

autarch: We already download all sorts of other stuff from the internet at large. What makes this particular thing different?

dragonlee8: This is an external dependency which is not managed by any packaging system either. Let's say if synk website is down for one day, our release will be blocked.

huan-Mongo: This is an external dependency which is not managed by any packaging system either. Let's say if synk website is down for one day, our release will be blocked.

autarch: Again, we already have a number of other similar unmanaged deps (precious, etc.). If there were a well organized way of vendoring these things that wasn't just a one off that'd be great. I will bring this up with DevProd.

huan-Mongo: Yea, that's why I said we should do it "in the long term". We don't need to do it now. Besides, I feel more confident to download from Github than other third party websites, since if GitHub goes down, we probably won't be able to release anyway.

autarch: Fair point re: GitHub. FWIW, `static.snyk.io` appears to be served via Akamai, so I'd expect it to be fairly reliable.

edobranov (approved):

PR #654, opened 2024-05-20
TOOLS-3533 Add tooling to generate and check the SBOM Lite file

autarch:


common.yml
```diff
@@ -517,16 +517,14 @@ functions:
...
-          ./etc/generate-notices.pl > ./THIRD-PARTY-NOTICES
-          git diff --exit-code
+  "check sbom lite":
+    command: shell.exec
+    type: test
+    params:
+      working_dir: src/github.com/mongodb/mongo-tools
+      script: |
+        ${_set_shell_env}
+        ./scripts/regenerate-and-diff-sbom.sh
```
huan-Mongo, line 527: I remember we wanted to get rid of all the shell script and in favor of Go code?

buildscript/sbom.go
```diff
@@ -0,0 +1,10 @@
+package buildscript
+
+import (
+	"github.com/craiggwilson/goke/pkg/sh"
+	"github.com/craiggwilson/goke/task"
+)
+
+func WriteSBOMLite(ctx *task.Context) error {
+	return sh.Run(ctx, "scripts/regenerate-sbom.sh")
```
huan-Mongo, line 9: Why don't we write this as Go code?

huan-Mongo (changes requested):
Some questions about the shell script to be created. And do you have a sample SBOM file that's generated?

autarch:
> Some questions about the shell script to be created. And do you have a sample SBOM file that's generated?

The SBOM is checked into the repo. You didn't see it in the Files Changed tab because it's huge, so it's collapsed by default.

buildscript/sbom.go
autarch: If you look at the script you'll see it has a lot of pipes. We'd have to reimplement that all in Go. I just copied most of this from our docs for the `silkbomb` tool. See https://docs.devprod.prod.corp.mongodb.com/mms/python/src/sbom/silkbomb/docs/PURLS#golang-purl-example

scripts/regenerate-sbom.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
edobranov, line 1: Could we add a few comments explaining what's going on here? The `jq` and `sed` commands in particular are difficult to parse unless you already know what the output looks like.

edobranov (approved):
LGTM % one suggestion!

scripts/regenerate-sbom.sh
autarch: Added.

common.yml
autarch: If I rewrote this in Go all it would do is execute exactly the same commands. I don't think that would improve things much. I think the case for "don't use shell" is more for complex programs like the code we have in Mongosync to download Server versions.

huan-Mongo: I think the purpose of "not use shell" is for better cross-platform support. Since this is a release script we probably can only run it on Linux, I agree it's probably fine to write in shell.

huan-Mongo (approved):
LGTM!

PR #655, opened 2024-05-21
TOOLS-3535 Add gosec as a linter and generate a SARIF report as part of release

autarch:
This PR adds infrastructure to install dev tools including `precious`, which we use to run `gosec` as a linter. As part of the release process, this will generate a SARIF report. If there are new issues in the code base, then the release will be blocked.

It also updates the `lint-go` Evergreen task to run our new linting, which right now will just run `gosec`.

This also includes comments to disable gosec's linting in a few spots. These comments end up in the eventual SARIF report explaining why we are suppressing a particular error.


autarch:
Here's an example of a SARIF report containing issues: https://parsley.mongodb.com/evergreen/mongo_tools_sarif_gosec_SARIF_report_patch_170b6f2557fa659b74e8268afcfaadd7c8d608df_664c34b2ceb3230007aa18a8_24_05_21_05_44_19/0/task?bookmarks=0,3737&selectedLineRange=L91

And an example of a report where all issues have been resolved: https://parsley.mongodb.com/evergreen/mongo_tools_sarif_gosec_SARIF_report_patch_170b6f2557fa659b74e8268afcfaadd7c8d608df_664c9f8f4bc7390007c3b715_24_05_21_13_20_16/0/task?bookmarks=0,261&shareLine=111

edobranov:
Going to remove myself for now since I'll be OOO and back on 5/30. Feel free to re-add me if you haven't gotten someone else to take a look by then!

buildscript/sa.go
```diff
@@ -4,13 +4,199 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	devBinExe, err := devBinFile(exeName)
+	if err != nil {
+		return err
+	}
+
+	exists, err := executableExistsWithVersion(ctx, devBinExe, preciousVersion)
```
huan-Mongo, line 137: Why this check `preciousVersion` instead of `toolVersion`?

buildscript/util.go
```diff
@@ -0,0 +1,205 @@
...
+		if matches {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func pathExists(p string) (bool, error) {
+	_, err := os.Stat(p)
```
huan-Mongo, line 107: I assume this function name suggests it checks if the path exists as a `directory`.  Maybe we should also check if the `FileInfo` it returns is a `DIR`.

precious.toml
```diff
@@ -0,0 +1,28 @@
...
+    "release/**/*.go",
+    "vendor/**/*.go",
+]
+invoke.per-dir-or-once = 7
+working-dir = "root"
+path-args = "dir"
+# This wrapper lets us control the output with an env var so we can generate a
+# SARIF report in Evergreen but use the default format for linting.
+cmd = [
+    "$PRECIOUS_ROOT/etc/gosec-wrapper.sh",
```
huan-Mongo, line 19: Is it required to run the script on Windows as well? It seems the rest of the release script all support Windows.

huan-Mongo (changes requested):

precious.toml
autarch: This will not need to run on Windows. We only need this to run in CI on _one_ distro, not all of them.

buildscript/util.go
autarch: I renamed this to `fileExists` and made it check that the thing is actually a file.

huan-Mongo (approved):
LGTM!

buildscript/util.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return false, err
+}
+
+func versionMatches(ctx *task.Context, exe, wantVersion string) (bool, error) {
+	out, err := sh.RunOutput(ctx, exe, "--version")
+	if err != nil {
+		return false, err
+	}
+	return strings.Contains(out, wantVersion), nil
```
suspicious-mango, line 122: It's probably not going to be a problem anytime soon, but `wantVersion="0.7.2"` would match versions like `"0.7.21"`, `"10.7.2"`, etc.

ghost (approved):
LGTM % maybe how we match the version string, but may also not be a problem.

buildscript/util.go
autarch: Good point. I changed this to use a regex with `\b` on either side of the version string.

PR #658, opened 2024-05-28
TOOLS-3551 Fix mongorestore test failure with Server 8.0+

autarch:
This test checks expects an error when restoring a timeseries collection where its `system.buckets` already exists. This no longer causes an error with 8.0.

### DIFF
--- a/mongorestore/mongorestore_test.go
+++ b/mongorestore/mongorestore_test.go
@@ -1753,7 +1753,7 @@ func TestRestoreTimeseriesCollections(t *testing.T) {
 			So(err, ShouldBeNil)
 			// In the 8.0 release, this no longer leads to an error, so
 			// there's nothing to test here.
-			if restore.serverVersion.LT(db.Version{8, 0, 0}) {
+			if restore.serverVersion.GTE(db.Version{8, 0, 0}) {
 				return
 			}
 
--- a/mongorestore/mongorestore_test.go
+++ b/mongorestore/mongorestore_test.go
@@ -1754,6 +1754,7 @@ func TestRestoreTimeseriesCollections(t *testing.T) {
 			// In the 8.0 release, this no longer leads to an error, so
 			// there's nothing to test here.
 			if restore.serverVersion.GTE(db.Version{8, 0, 0}) {
+				SkipSo()
 				return
 			}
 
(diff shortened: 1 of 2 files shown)


mongorestore/mongorestore_test.go
```diff
@@ -1749,10 +1749,16 @@ func TestRestoreTimeseriesCollections(t *testing.T) {
 		})
 
 		Convey("restoring a timeseries collection when the system.buckets collection already exists on the destination should fail", func() {
-			testdb.RunCommand(context.Background(), bson.M{"create": "system.buckets.foo_ts"})
-			args = append(args, DirectoryOption, "testdata/timeseries_tests/ts_dump")
 			restore, err := getRestoreWithArgs(args...)
 			So(err, ShouldBeNil)
+			// In the 8.0 release, this no longer leads to an error, so
+			// there's nothing to test here.
+			if restore.serverVersion.LT(db.Version{8, 0, 0}) {
```
huan-Mongo, line 1756: Doesn't the comparison mean 
if server version < 8.0, then 
     return
? 

I thought we were only going to test when the server version < 8.0.

huan-Mongo (changes requested):

mongorestore/mongorestore_test.go
autarch: Oops, yeah, I got this backwards. I wonder if the pre-release is comparing as less than 8.0.0, which is why this passed. I guess I'll find out now that I fixed the comparison.

suspicious-mango (approved):
LGTM

autarch:
This is fixed properly now. Here's a run for "latest" where the test is skipped;

https://parsley.mongodb.com/evergreen/mongo_tools_rhel80_integration_latest_patch_706422de3f408f79e136e78bf84c7f7760e702d7_6657f93f7304b30007a3b272_24_05_30_03_57_52/0/task?bookmarks=0,19586&shareLine=13123

    [2024/05/30 12:11:46.118]        |     restoring a timeseries collection when the system.buckets collection already exists on the destination should fail ✔⚠

And one for 7.0 where it runs:

https://parsley.mongodb.com/evergreen/mongo_tools_rhel80_integration_7.0_patch_706422de3f408f79e136e78bf84c7f7760e702d7_6657f93f7304b30007a3b272_24_05_30_03_57_52/0/task?bookmarks=0,21656&shareLine=14010

    [2024/05/30 12:06:30.409]        |     restoring a timeseries collection when the system.buckets collection already exists on the destination should fail ✔2024-05-30T04:06:24.544+0000	using default 'dump' directory

The "⚠" seems to be how `goconvey` indicates a skipped assertion.

huan-Mongo (approved):
LGTM!

PR #660, opened 2024-05-28
Restore the `THIRD-PARTY-NOTICES` file

autarch:
Many of the licenses have text saying something to the effect that we must reproduce the license verbatim from the code that the license covers. The SBOM file probably isn't sufficient in those cases.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -517,6 +517,17 @@ functions:
 
           echo -n "$MONGOD_URI" > MONGOD_URI
 
+  "check third-party notices":
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: src/github.com/mongodb/mongo-tools
+        script: |
+          set -e
+          set -x
+          ./etc/generate-notices.pl > ./THIRD-PARTY-NOTICES
+          git diff --exit-code
+
   "check sbom lite":
     command: shell.exec
     type: test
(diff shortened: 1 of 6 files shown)


common.yml
```diff
@@ -1691,10 +1702,15 @@ tasks:
...
+  - name: check-third-party-notices
+    commands:
+      - func: "fetch source"
+      - func: "check third-party notices"
+
   - name: check-sbom-lite
     commands:
       - func: "fetch source"
-      - func: "check sbom lite"
+      - func: "check third-party notices"
```
mmcclimon, line 1713: This change isn't quite right; this should still be "check sbom lite".

etc/generate-notices.pl
```diff
@@ -0,0 +1,438 @@
...
+);
+
+print forked_licenses();
+
+for my $entry (sort { $a->[0] cmp $b->[0] or $a->[1] cmp $b->[1] } @license_files) {
+    ( my $package_name = $entry->[0] ) =~ s{vendor/}{};
+    my $file = basename($entry->[1]);
+    my $license_text = do { local ( @ARGV, $/ ) = $entry->[1]; <> };
+    $license_text =~ s/ +$//mg;
+    say "" x 70;
```
mmcclimon, line 27: (I know you didn't write this code, but lol at `say "" x 70`, to be really sure you printed nothing.)

mmcclimon (approved):
LGTM % the one thing in the evergreen file, but I trust you can do that without needing another review.

(Also, thanks for the throwback to reviewing perl.)

common.yml
autarch: Oops, thanks. I really wish Evergreen config had one _less_ level of structure (the variants -> tasks -> functions -> commands thing is one level too many).

etc/generate-notices.pl
autarch: Heh, I never noticed this.

common.yml
autarch: Fixed.

ghost (approved):

PR #661, opened 2024-05-29
TOOLS-3554 Really update most of our dependencies

autarch:
The previous updating I did as part of TOOLS-3533 did not actually upgrade a number of things. Apparently `go get -u` doesn't upgrade everything, so I updated each dep individually with `go get -u $pkg`.
    
This PR also adds some comments in `go.mod` about things we don't want to upgrade, and moves all the indirect deps to the same `require` block.

Finally, I had to change a test that compared two floats to account for some rounding error. I'm not sure why this didn't fail before.


dragonlee8 (approved):
LGTM! I am just curious why `go get -u` didn't upgrade all the dependencies...

huan-Mongo (approved):
LGTM! I am just curious why go get -u didn't upgrade all the dependencies...

huan-Mongo (approved):

huan-Mongo:
Sorry I commented above with wrong github account again...

suspicious-mango (approved):
LGTM

autarch:
> LGTM! I am just curious why go get -u didn't upgrade all the dependencies...

Apparently you have to run `go get -u all`. Without the `all` it updates ... some things. I don't know how it chooses. AFAICT, this is not documented in the output from `go help get`.

PR #662, opened 2024-05-29
TOOLS-3532 Record releases in Papertrail

autarch:
This actually records every single PR run in Papertrail, but only properly tagged releases are recorded using the product name "database-tools". Everything else uses "database-tools-dev". Configuring things this way makes it much simpler to be sure that this Papertrail integration will actually work when we do a release.


autarch:
The papertrail CI task fails on one distro right now. This is an issue that DevProd is aware of and will fix.

common.yml
```diff
@@ -1801,10 +1834,12 @@ buildvariants:
       - name: "dist"
       - name: "sign"
         run_on: rhel87-small
+      - name: "record-in-papertrail"
```
huan-Mongo, line 1837: I assume we want to record every release in the papertrail, why not extend the `push` task instead?

release/release.go
```diff
@@ -1544,3 +1552,47 @@ func downloadArtifacts(v string, artifactNames []string) {
...
+// _would_ be used if we were to upload them as official releases.
+//
+// This doesn't _guarantee_ uniqueness since we could run the exact same
+// commit in Evergreen twice, which would generate the same names, but it's
+// close enough. Papertrail will still record the information we give even if
+// we use the same filenames for two different CI runs.
+func renameReleaseFilesForPapetrail() {
+	pf, err := platform.GetFromEnv()
+	if err != nil {
+		log.Fatalf("could not find platform: %v", err)
```
huan-Mongo, line 1568: It seems the caller function returns an error, should we return the err instead of panic inside the function?

release/release.go
```diff
@@ -1544,3 +1552,47 @@ func downloadArtifacts(v string, artifactNames []string) {
...
+	releaseFiles, err := filepath.Glob(glob)
+	if err != nil {
+		log.Fatalf("could not get files matching %q: %v", glob, err)
+	}
+
+	baseName := basenameForPlatformAndVersion(pf, v)
+	for _, file := range releaseFiles {
+		ext := filepath.Ext(file)
+		newName := baseName + ext
+		if err := os.Rename(file, newName); err != nil {
```
huan-Mongo, line 1586: I think the second param of `os.Rename` is a path instead of a filename. Is it intentional or an overlook?

huan-Mongo (changes requested):
In general it looks good!

common.yml
autarch: Yes, good call. I changed this.

release/release.go
autarch: Done.

release/release.go
autarch: I'm not sure what you mean. This works as written. Here's the output from running this locally as a test:

```
renamed "release.tgz" to "mongodb-database-tools-ubuntu2204-x86_64-100.9.4-34-g427bcf56-dirty.tgz"
```

That's exactly how we want the rename to work.

suspicious-mango (approved):
LTGM. Typos and a question.

release/release.go
autarch: I don't think there's much of a difference. If the function returns an error it will ultimately get logged and cause the program to error out.

huan-Mongo: I think there is still some difference? The caller of the function might implement additional logic after the panic error, it's better to return an error instead of panic in place.

huan-Mongo (changes requested):
Just one more question on the `os.Rename` function.

release/release.go
autarch: A "just a file name" is still a path. It's _not_ a fully-qualified path. This all works fine because this code is run in the same directory as the files being renamed, so we don't need to provide a longer path.

huan-Mongo: Thanks, that's why my question was if this is intentional or an overlook?

huan-Mongo (approved):
LGTM!

PR #663, opened 2024-05-31
TOOLS-3550 Check that the Augmented SBOM file is up to date, not just that it exists

autarch:
This changes our Evergreen config so that we check that the Augmented SBOM is up to date when doing a release, as opposed to just checking that the file exists.

This also renames the Augmented SBOM file to match the release tag style we use and the common extension for an SBOM.


suspicious-mango:
Is this also part of TOOLS-3550?

autarch:
> Is this also part of [TOOLS-3550](https://jira.mongodb.org/browse/TOOLS-3550)?

Yeah, more or less. I updated the title. It's not strictly _required_ per TOOLS-3550 but adding this automation prevents us from doing a release with any third-party deps w/ vulnerabilities. If we did that we'd just have to do another release soon after, which would be a waste of time.

huan-Mongo (approved):
LGTM!

huan-Mongo (approved):
LGTM!

suspicious-mango (approved):
LGTM

PR #664, opened 2024-05-31
Pass `--file=./go.mod` to `snyk test`

autarch:
Right now it uses `go.mod` by default, but if for some reason we ever added a deps file for another language, like a `package.json`, it might start testing that instead.


mmcclimon (approved):
Oh sure, LGTM! (I am pleased that my unrelated whinging led to some fruitful outcome, in the end.)

PR #665, opened 2024-06-03
TOOLS-3550 Set `IS_FAKE_TAG` env var when `fake_tag_for_release_testing` param is set

autarch:
If this env var is true then it will block doing a real release. This provides a bit of extra safety when using this fake tag param.


mmcclimon (approved):
LGTM, thanks!

suspicious-mango (approved):
LGTM

PR #667, opened 2024-06-04
TOOLS-3550 Make `regenerate-and-diff-augmented-sbom.sh` exit 0 when there's no tag

autarch:
The `check-augmented-sbom` task gets called for `master` branch runs, not just releases, so we can't assume it will always be run when there's a tag to be checked.

This also updates the Augmented SBOM file to account for changes to our deps that were merged in other PRs.


suspicious-mango (approved):
LGTM

PR #668, opened 2024-06-05
TOOLS-3555 Remove all support for Debian 8 (and 7!)

autarch:
We can't publish packages for Debian 8 to barque any more, and this distro has been EOL since June, 2020.

This PR also removes some config related to Server 3.4, which we no longer test with. Finally, this also removes the long-unused `common-pvt.yml` file.


huan-Mongo (approved):
LGTM!

ghost (approved):
LGTM

PR #670, opened 2024-06-07
Use prettier to lint & tidy JSMarkdown & YAML

autarch:


precious.toml
```diff
@@ -26,3 +26,41 @@ cmd = [
...
+[commands.prettier-md]
+type = "both"
+invoke = "once"
+working-dir = "root"
+path-args = "file"
+include = [ "*.md" ]
+exclude = [
+    ".github/pull_request_template.md",
+    ".github/top-comment-content.md",
+]
```
mmcclimon, line 52: (nit) You don't need these here; copy-pasted from elsewhere.

SERVERVENDORING.md
mmcclimon, line 1: (Not for this PR, but: I suspect we could get rid of this file entirely. It says we only do it for 4.2 and older but 4.2 is past EOL so I would highly doubt we'd ever vendor the tools for it again.)

mmcclimon (approved):
LGTM! I don't see anything here that enforces this in CI, but I think that's fine too.

.gitignore
```diff
@@ -18,6 +18,7 @@
 /purls.txt
 /silkbomb.env
 /release.*
+!/RELEASE.md
```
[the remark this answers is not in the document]
autarch: I think I needed this for `precious`, since it uses a crate under the hood that parses the `.gitignore` to ignore files for tidying & linting.

precious.toml
autarch: Fixed.

suspicious-mango (approved):
LGTM

PR #672, opened 2024-06-07
Install all dev tools under dev-bin and add tidying with goimports

autarch:
Installing all the tools under `dev-bin` avoids conflicts with other projects that might be using the same tools. This also adds tidying with `goimports` to the precious config, and tidies our code with it.

I also fixes the regex to check whether an existing tool has the expected version.

### DIFF
--- a/precious.toml
+++ b/precious.toml
@@ -66,5 +66,5 @@ lint-failure-exit-codes = [1]
 ignore-stderr = ".*"
 
 # It'd be nice to use prettier to tidy the JS code too, but it seems to make
-# changes that cause it to fail to parse with the verisons of the Mongo shell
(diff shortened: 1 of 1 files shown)


mmcclimon (commented):
I think this is fine; I haven't yet looked at all the changes, because: I suspect it would be useful, before adopting a new tool here, to see what this change would look like in mongosync. Could you whip something up for that? I wouldn't imagine it would take a super long time, and then I think it would be easier for me (and probably, the rest of the team) to compare and contrast?

mmcclimon (approved):
Sure, LGTM!

(I get confused when I see these tools PRs because I'm like...didn't I _just_ review this for Dave?)

Becky-Hu-82 (approved):
LGTM!

PR #674, opened 2024-06-10
TOOLS-3559 Add the Augmented to all release artifacts (tarballs, debs, etc.)

autarch:
This also significantly changes the release code so that the files generated by the `dist` task following the naming scheme of the file we would/will release. For non-tag builds, these files will include the git commit in their name. When we do a release, the files that we release and make available for download should only include the tag.


release/release.go
```diff
@@ -385,20 +368,32 @@ func buildRPM() {
...
-	//                                             |----- mongodb-database-tools/
-	//                                                              |--- staticFiles
+	// |-- SOURCES/
+	// |   |-- mongodb-database-tools.tar.gz:
+	//         |
+	//         mongodb-database-tools/
+	//         |-- usr/
+	//             |-- bin/
+	//             |   |-- bsondump
+	//             |    |--- mongo*
```
suspicious-mango, line 392: not: this looks misaligned.

release/release.go
```diff
@@ -1553,48 +1544,73 @@ func downloadArtifacts(v string, artifactNames []string) {
...
 
-	v, err := version.GetCurrent()
-	if err != nil {
-		return errors.Wrap(err, "could not get version")
+	var (
+		sourceFile string
+		tag        = os.Getenv("EVG_TRIGGERED_BY_TAG")
+	)
+	if tag != "" {
+		sourceFile = filepath.Join(repoRoot, "ssdlc", tag+tag+".bom.json")
```
suspicious-mango, line 1568: is `tag` supposed to appear twice here?

scripts/sign_artifacts.sh
```diff
@@ -72,13 +77,17 @@ case $MONGO_OS in
...
-    authenticode_sign "release.msi"
-    pgp_sign "release.zip" "release.zip.sig"
+    msifile=$(ls "$glob_base.msi")
+    authenticode_sign "$msifile"
+    zipfile=$(ls "$glob_base.zip")
+    pgp_sign "$zipfile" "$zipfile.sig"
     ;;
 
   *)
+    ls -l # XXX - debug
```
suspicious-mango, line 87:
```suggestion
```

suspicious-mango (changes requested):

scripts/sign_artifacts.sh
autarch: This PR is not ready for review. I didn't request a review on it, did I?

release/release.go
autarch: Fixed.

huan-Mongo (approved):
LGTM!

PR #676, opened 2024-06-11
TOOLS-3527 Make sure the Augmented SBOM has an empty `vulnerabilities` key if there are no vulnerabilities

autarch:
The `silkbomb` tool (or Silk?) should be doing this for us, but it's not. This PR uses `jq` to append this key if it's not present.

This also fixes the `regenerate-augmented-sbom.sh` script to exit `0` if there is no tag. This script is run for the `main` branch as well as for tags.


suspicious-mango (approved):
LGTM

huan-Mongo (approved):
LGTM!

PR #677, opened 2024-06-14
Fix several issues in the release code

autarch:
* When creating a `version.Version`, we should not include the commit in the struct if we have a proper release tag. This fixes an issue where filenames for artifacts from a release tag incorrectly included the git commit.
* Fixed a bug in a script used for Papertrail integration where release tags caused the script to just fail.
* Improved some logging.


huan-Mongo (approved):
LGTM!

suspicious-mango (approved):
LGTM

PR #678, opened 2024-06-14
Remove the `push` task dependency for the `gosec-SARIF-report` task

autarch:
Generating the SARIF report doesn't depend on the code being pushed anywhere.


mmcclimon (approved):
LGTM!

PR #681, opened 2024-06-14
Write out an `.evergreen.yml` file before running `evergreen validate`

autarch:
This task never worked properly, but in the past `evergreen validate` would incorrectly exit with `0` when it couldn't authenticate itself.


huan-Mongo (approved):
LGTM!

PR #682, opened 2024-06-14
Add `eslint` to `precious` config and remove `lint-js` task

autarch:
This also updates our `eslint` version to 8.57.0, which is the most recent version that works with our YAML config file for `eslint`. Updating our config for the 9.x version didn't seem trivial.


autarch:
If you disable showing whitespace in the files view it greatly reduces the size of the PR.

edobranov (approved):

tdq45gj (approved):
LGTM!

PR #685, opened 2024-06-17
Update the SBOM Lite with the latest version of the Silkbomb tool

autarch:
This added a new `"vulnerabilities": []` key. This isn't a _meaningful_ change, but if it's missing from the checked-in file then the `check-sbom-lite` task fails.


suspicious-mango (approved):
LGTM

PR #687, opened 2024-06-19
TOOLS-3572 Fix the version used for Debian packages and add tests for version code

autarch:
For non-stable releases, we want to use `VERSION~DATE.COMMIT` (e.g. ` mongodb-database-tools_100.9.5+20240619.d8f2d106_amd64.deb`) in the Debian package name. For stable releases, the commit should not be included.

This also removes the `Pre` field from the version. This was no longer used with our current tagging scheme, because we're not doing rc releases.

### DIFF
--- a/release/version/version.go
+++ b/release/version/version.go
@@ -13,42 +13,32 @@ type Version struct {
 	Major int
 	Minor int
 	Patch int
-	Pre   string
 
 	Commit string
 }
 
 func Parse(desc string) (Version, error) {
-	if desc[0] == 'r' || desc[0] == 'v' {
-		desc = desc[1:]
+	// This throws out the part after the dash, but if this was called from
+	// `GetCurrent`, then we'll capture the commit in the `Commit` field in
+	// that func.
+	parts := strings.Split(strings.Split(desc, "-")[0], ".")
+	if len(parts) < 2 || len(parts) > 3 {
+		return Version{}, fmt.Errorf("could not find a two- or three-part dotted version in %q", desc)
 	}
-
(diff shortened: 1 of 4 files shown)


release/version/version.go
```diff
@@ -13,42 +13,32 @@ type Version struct {
...
-			return Version{}, fmt.Errorf("failed to parse minor version %q", parts[1])
-		}
+	minor, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return Version{}, fmt.Errorf("failed to parse minor version %q", parts[1])
 	}
 
-	pat := 0
+	var patch int
 	if len(parts) > 2 {
```
huan-Mongo, line 40: []
```suggestion
	if len(parts) == 3 {
```

release/version/version.go
```diff
@@ -83,34 +72,43 @@ func GetCurrent() (Version, error) {
...
-		vStr = fmt.Sprintf("%s-%s", vStr, v.Pre)
+	return fmt.Sprintf("%d.%d.%d", v.Major, v.Minor, v.Patch)
+}
+
+func (v Version) StringWithCommit() string {
+	if v.Commit == "" {
+		return v.String()
 	}
-	return vStr
+	return fmt.Sprintf("%s-g%s", v.String(), v.Commit[:8])
```
huan-Mongo, line 87: is it intentionally to add `g` inside the string `%s-g%s` ?

huan-Mongo (changes requested):
Looks good. Just one question.

huan-Mongo (approved):
LGTM!

PR #689, opened 2024-06-20
TOOLS-3574 Run tests against 8.0 and add ubuntu2404 variants

jddubois:
Runs integration tests against MongoDB 8.0 and adds dist, sign, and push tasks for ubuntu2404 variants.


release/platform/platform.go
```diff
@@ -536,7 +536,7 @@ var platforms = []Platform{
 		OS:        OSLinux,
 		Pkg:       PkgDeb,
 		Repos:     []Repo{RepoEnterprise, RepoOrg},
-		BuildTags: []string{"failpoints"},
+		BuildTags: defaultBuildTags,
```
jddubois, line 539: This works as is, I don't think there's any reason we need to not support Kerberos on ubuntu arm variants.

autarch: I'm pretty sure this wasn't done intentionally when these platforms were added. We probably should've caught it in PR review. Good to fix it now!

autarch (approved):
LGTM!

tdq45gj (commented):
Do we need to push to the Ubuntu2404 server repo? If so, I think we should add the new platforms to `etc/repo-config.yml` as well.

jddubois:
> repo-config.yml

Great catch- thanks! Updated that with codename `noble`, which I see some references to in Slack so I think that's correct.

tdq45gj (approved):
LGTM!

PR #690, opened 2024-06-25
Update Augmented SBOM for 100.9.5 to match what's in Silk

autarch:
I think the `timestamp` field got out of sync with what's in Silk because of some confusion on my part.

This also updates the CI scripts we use to check that the SBOM files are up to date to _not_ ignore changes to the timestamp or version. If those fields are updated then this is not spurious, and the diff represents a real change that we need to include in that file.


jddubois (approved):
Seems much safer to just diff everything, LGTM!

PR #691, opened 2024-06-28
TOOLS-3546 Support vectored inserts oplog format

tdq45gj:
This PR makes the tools treat vectored inserts as non-transaction by checking the `multiOpType` field in the oplog, meaning that a vectored insert oplog will be applied on sight instead of being buffered until it commits.
It also adds tests to verify that mongodump can dump an oplog that's a vectored insert and mongorestore can restore the dumped oplog into destination.


mtrussotto (commented):
Code looks good, there should probably be tests where there are multiple linked oplog entries with multiOpType.

tdq45gj:
@mtrussotto Good call. Added.

mtrussotto (approved):

jddubois (approved):
LGTM, thanks! For visibility- we discussed a few things over slack:

* Why is it safe to perform vectored inserts non-transactionally?
   The `multiOpType` field is only set outside of transactions.
* Why did we need the `replSet` changes?
   These tests are not supported in standalone, and we didn't have a way to check this 
   before (they only run via `intergration-8.0-cluster`)

PR #692, opened 2024-07-03
TOOLS-3575 Add coverage for mixed-schema timeseries buckets

tdq45gj:
This PR makes mongorestore handle `CannotInsertTimeseriesBucketsWithMixedSchema` by modifying the `timeseriesBucketsMayHaveMixedSchemaData` collection parameter. It adds a test to verify that mongodump can dump a timeseries bucket with mixed schema and another test to verify that restoring a bucket with mixed schema sets `timeseriesBucketsMayHaveMixedSchemaData` to false.


mongodump/mongodump_test.go
```diff
@@ -332,6 +332,60 @@ func setUpTimeseries(dbName string, colName string) error {
...
+			{"metaField", "m"},
+		}},
+	}
+
+	createRes := sessionProvider.DB(dbName).RunCommand(nil, createCmd)
+	if createRes.Err() != nil {
+		return createRes.Err()
+	}
+
+	// SERVER-84531 was only backported to 7.3.
```
jddubois, line 363: It looks like backports were approved for v6.0 and v7.0. Can we create a ticket blocked on those backports to update these tests, and link a TODO here?

mongorestore/mongorestore_test.go
```diff
@@ -1658,6 +1658,92 @@ func createTimeseries(dbName, coll string, client *mongo.Client) {
...
+
+	defer sessionProvider.Close()
+
+	session, err := sessionProvider.GetSession()
+	if err != nil {
+		t.Fatalf("No client available")
+	}
+
+	fcv := testutil.GetFCV(session)
+	// The server fix for SERVER-84531 was only backported to 7.3.
```
jddubois, line 1679: Same here

mongorestore/mongorestore_test.go
```diff
@@ -1658,6 +1658,92 @@ func createTimeseries(dbName, coll string, client *mongo.Client) {
...
+	require.NoError(t, err)
+	require.Equal(t, int64(2), count)
+
+	count, err = bucketColl.CountDocuments(ctx, bson.M{})
+	require.NoError(t, err)
+	require.Equal(t, int64(1), count)
+
+	hasMixedSchema, err := timeseriesBucketsMayHaveMixedSchemaData(bucketColl)
+	require.NoError(t, err)
+	require.False(t, hasMixedSchema)
```
jddubois, line 1714: I'm a bit confused why this is succeeding- my understanding of the server change is that we need to set hasMixedSchema=true in order to perform these direct writes?

tdq45gj:
@jddubois @autarch I made the changes to mongorestore but had to disable the test on v8.0 because of [SERVER-92222](https://jira.mongodb.org/browse/SERVER-92222). The server bug is on v8.0, which gives a wrong error code for the `CannotInsertTimeseriesBucketsWithMixedSchema` error.

common/db/db.go
```diff
@@ -548,6 +551,27 @@ func CanIgnoreError(err error) bool {
...
+	if err == nil {
+		return false
+	}
+
+	switch mongoErr := err.(type) {
+	case mongo.WriteError:
+		return mongoErr.Code == ErrCannotInsertTimeseriesBucketsWithMixedSchema
+
+	case mongo.BulkWriteException:
+		for _, writeErr := range mongoErr.WriteErrors {
```
jddubois, line 565: It looks like this is only checking the first error? Should we instead be returning true if any or all are 408 errors?

mongorestore/restore.go
```diff
@@ -530,7 +531,23 @@ func (restore *MongoRestore) RestoreCollectionToDB(dbName, colName string,
...
+				logicalColName, nameErr := db.GetTimeseriesColNameFromBucket(colName)
+				if nameErr != nil {
+					resultChan <- result.withErr(nameErr)
+					return
+				}
+				if collModErr := restore.AllowMixedSchemaInTimeseriesBucket(dbName, logicalColName); collModErr != nil {
+					resultChan <- result.withErr(errors.Wrap(collModErr, "failed to handle mixed schema in a timeseries bucket"))
+					return
+				}
+				bwResult, bwErr = bulk.Flush()
```
jddubois, line 548: Is it safe to call `ResetBulk()` twice here, or should we use `TryFlush`?

jddubois (commented):
A few more comments- where did you disable the tests? Maybe we should just wait until the backport is merged? It looks like it's in code review.

common/db/db.go
tdq45gj: Good catch!

tdq45gj:
@jddubois The test is disabled on v8.0 [here](https://github.com/mongodb/mongo-tools/pull/692/files#diff-57bd455e42f464e58586dd36657a3f51c0cbd32868d4679f119910874c964d68R1689-R1694). I think we need to wait for the bug fix to be backported to v8.0 and released. Given that the tools and server cyclic release dependency is still a problem, I don't think we can wait for that. I've tested this on v8.0 with its special error code (409) and it works.

mongorestore/mongorestore_test.go
```diff
@@ -1658,6 +1659,163 @@ func createTimeseries(dbName, coll string, client *mongo.Client) {
...
+	// TODO: Enable tests for 6.0, 7.0 and 8.0 (TOOLS-3597).
+	// The server fix for SERVER-84531 was only backported to 7.3.
+	if cmp, err := testutil.CompareFCV(fcv, "7.3"); cmp < 0 {
+		if err != nil {
+			t.Fatalf("Failed to get FCV: %v", err)
+		}
+		t.Skip("Requires server with FCV 7.3 or later")
+	}
+
+	if cmp, err := testutil.CompareFCV(fcv, "8.0"); cmp >= 0 {
```
jddubois, line 1689: Let's just add a TODO here with a linked REP ticket that's blocked on that backport? Or if not a TODO, as long as we have the info in the ticket that's fine.

jddubois (approved):
LGTM!

mongorestore/mongorestore_test.go
tdq45gj: I added that info to [TOOLS-3597](https://jira.mongodb.org/browse/TOOLS-3597).

common/db/db.go
```diff
@@ -548,6 +551,27 @@ func CanIgnoreError(err error) bool {
...
+			if writeErr.Code != ErrCannotInsertTimeseriesBucketsWithMixedSchema {
+				return false
+			}
+		}
```
autarch, line 569: I think the logic here is still incorrect. If there are multiple write errors and the first one _isn't_ the one we check, but a later one is, then we'd return false incorrectly.

common/db/db.go
```diff
@@ -570,3 +594,12 @@ func IsMMAPV1(database *mongo.Database, collectionName string) (bool, error) {
 	_, ok := collStats[numExtents]
 	return ok, nil
 }
+
+// GetTimeseriesColNameFromBucket returns a timeseries collection name from its bucket collection name.
+func GetTimeseriesColNameFromBucket(bucketColName string) (string, error) {
```
autarch, line 599: nit: We usually abbreviate "collection" to "coll", not "col".

mongorestore/restore.go
```diff
@@ -530,7 +531,23 @@ func (restore *MongoRestore) RestoreCollectionToDB(dbName, colName string,
...
+
+			if db.TimeseriesBucketNeedsMixedSchema(bwErr) {
+				// Modify the timeseries collection and retry flushing the bulk writer.
+				logicalColName, nameErr := db.GetTimeseriesColNameFromBucket(colName)
+				if nameErr != nil {
+					resultChan <- result.withErr(nameErr)
+					return
+				}
+				if collModErr := restore.AllowMixedSchemaInTimeseriesBucket(dbName, logicalColName); collModErr != nil {
+					resultChan <- result.withErr(errors.Wrap(collModErr, "failed to handle mixed schema in a timeseries bucket"))
```
autarch, line 545:
```suggestion
					resultChan <- result.withErr(errors.Wrap(collModErr, "failed to enable mixed schema in a timeseries bucket"))
```

But actually, I think this wrapping should be done in the `AllowMixedSchemaInTimeseriesBucket` method, rather than here.

autarch (changes requested):
This looks good overall but there are a few things I think need to be updated.

common/db/db.go
tdq45gj: I changed the logic a bit for it to return true if any of the errors is the one we check.

common/db/db.go
autarch: Did you change this locally but not push it?

autarch (approved):
LGTM % one thing from the previous review. No need for another review.

common/db/db.go
tdq45gj: oops, I changed the parameter name but omitted the function name.

autarch: Also the variable _in_ the func. Just search for "colname".

PR #697, opened 2024-07-11
Update dependabot.yml

tfogo:
I need to hand this off to somebody else. If we want to change how this works in the future, that's fine, but I need to hand this off now.


autarch (approved):
LGTM!

PR #698, opened 2024-07-15
TOOLS-3613 Add qa and legacy tests against 8.0

tdq45gj:


mmcclimon (approved):
I don't really know this code at all, but the new 8.0 blocks look just like the existing 7.0 blocks, which is good enough for me. LGTM, assuming you've had a PR run that didn't burn down in flames. Thanks!

autarch (approved):
LGTM!

PR #704, opened 2024-07-18
TOOLS-3616 Update go driver version to 1.16

tdq45gj:


common/options/options.go
```diff
@@ -679,7 +679,7 @@ func (opts *ToolOptions) NormalizeOptionsAndURI() error {
 	if err != nil {
 		return err
 	}
-	err = opts.setOptionsFromURI(cs)
+	err = opts.setOptionsFromURI(*cs)
```
autarch, line 682: I think it would make more sense to change this func to just accept the pointer.

common/options/options.go
```diff
@@ -641,7 +641,7 @@ func (opts *ToolOptions) setURIFromPositionalArg(args []string) ([]string, error
 				return []string{}, fmt.Errorf("too many URIs found in positional arguments: only one URI can be set as a positional argument")
 			}
 			foundURI = true
-			parsedURI = cs
+			parsedURI = *cs
```
autarch, line 644: Why not change the type of `parsedURI` instead?

common/options/options.go
```diff
@@ -417,7 +417,7 @@ func NewURI(unparsed string) (*URI, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error parsing URI from %v: %v", unparsed, err)
 	}
-	return &URI{ConnectionString: cs.String(), ConnString: cs}, nil
+	return &URI{ConnectionString: cs.String(), ConnString: *cs}, nil
```
autarch, line 420: Why not make this field accept a pointer?

autarch (changes requested):
I think instead of de-referencing the pointer everywhere we could just change our types to match what the driver gives us.

mongotop/mongotop.go
```diff
@@ -47,26 +46,37 @@ func (mt *MongoTop) runDiff() (outDiff FormattableDiff, err error) {
...
+		if elem.Key() != "note" {
+			if topinfo[elem.Key()] != (NSTopInfo{}) {
+				return nil, fmt.Errorf("duplicate namespace (%s) in top result", elem.Key())
+			}
+			info := NSTopInfo{}
+			if unmarshalErr := bson.Unmarshal(elem.Value().Document(), &info); unmarshalErr != nil {
+				return nil, unmarshalErr
+			}
+			topinfo[elem.Key()] = info
+		}
```
autarch, line 79: nit: I think this would read better this way:

```suggestion
		if elem.Key() == "note" {
			continue
		}

		if topinfo[elem.Key()] != (NSTopInfo{}) {
			return nil, fmt.Errorf("duplicate namespace (%s) in top result", elem.Key())
		}
		info := NSTopInfo{}
		if unmarshalErr := bson.Unmarshal(elem.Value().Document(), &info); unmarshalErr != nil {
			return nil, unmarshalErr
		}
		topinfo[elem.Key()] = info
```

autarch (approved):
LGTM % one small thing! No need for another review.

edobranov (approved):
LGTM!

PR #707, opened 2024-07-22
TOOLS-3618 Wrap access to global logger in a mutex and add build variant to run tests w/ race detector

autarch:


common.yml
```diff
@@ -2485,17 +2517,17 @@ github_pr_aliases:
   - variant: ".*"
     task: "^push$"
   # Run tests with the race detector enabled.
-  - variant: "ubuntu-race"
+  - variant: "rhel88-race"
     task: "^.*$"
   # Run all integration tests on one variant. We pick RHEL 8.0 because it's
   # a relatively recent platform that supports a wide range of servers.
-  - variant: "rhel88"
+  - variant: "rhel80"
```
tdq45gj, line 2524: I think the comment isn't correct since we moved all the tests to rhel88.

buildscript/build.go
```diff
@@ -166,6 +166,10 @@ func runTests(ctx *task.Context, pkgs []string, testType string) error {
 			env = append(env, testtype.ReplSetTestType+"=true")
 		}
 
+		if ctx.Get("race") == "true" {
+			args = append(args, "-race")
```
tdq45gj, line 170: I'm seeing a [warning](https://parsley.mongodb.com/evergreen/mongo_tools_rhel88_race_integration_4.0_patch_5d25238d53ea5d8a304e954459f659f39a160156_669e9b1105cd0400064099d7_24_07_22_17_46_59/0/task?bookmarks=0,773,7529) about unused argument "race". It seems that it's not working.

tdq45gj (changes requested):
Somehow `-race` doesn't seem to be working.

common.yml
autarch: Argh. I will update this to add a comment to tell people to stop editing this by hand.

buildscript/build.go
autarch: This should be fixedd.

tdq45gj (approved):
LGTM!

edobranov (approved):
LGTM

fgm:
Just to be sure, do these tests actually fail without the fix ? I do not see t.Parallel or specific goroutine code being used to trigger a race condition.

autarch:
> Just to be sure, do these tests actually fail without the fix ? I do not see t.Parallel or specific goroutine code being used to trigger a race condition.

I don't think any tests fail because of the logger issue with the race detector enabled. I guess none of the tests exercise the code in the particular way that your code did. The race detector did find a _different_ race that I fixed as well.

I should note that your use case of embedding the tools code in another tool is not really supported. So we're probably not going to tests specifically for that. But if you want to submit a PR to add tests for the logger issue that'd be great!

fgm:
Understood. The issue is that if the tests do not fail without the patch, they might not act as an efficient protection against future similar regressions. 

Interesting to learn that the use case of importing (not embedding) the code is not supported, considering this is a public API : if it is meant to be private, it might make sense to move it to an `internal` subfolder. Do you know if there is any "official" position on this specific public API one way or the other ?

autarch:
> Understood. The reason I'm asking is that if the tests do not fail without the patch, they might not act as an efficient protection against future similar regressions.
> 
> Interesting to learn that the use case of importing (not embedding) the code is not supported, considering this is a public API (in terms of Go): if it is meant to be private, it might make sense to move it to an `internal` subfolder. Do you know if there is any "official" position on this specific public API one way or the other ?

This is a pretty old codebase (first commits in 2014). I'd guess that when it was created nobody even thought about this use case since it's a set of binaries. I'm not even sure when `internal` became a thing, but it may not have existed in 2014.

I don't think we really have an official position on this at the moment. We're not going to go out of our way to support using this code as a library, but we currently have no plans to make this _harder_ either (for example, by moving everything to `internal`).

And we're happy to get bug reports like this one. We _should_ write all the code to be free from races, just because that heads off future issues when we change the code. That's why I added back running tests under the race detector in CI.

autarch:
Actually, I think I mispoke. Our _official_ position is that using this code as a library is _not_ a supported use case. But like I said, we don't have any plans to prevent this, either. I think this is a "use at your own risk" type of thing.

fgm:
Thanks for the extra info.

It would be good to mention it in the godoc because that is not what one would expect from a package with public symbols. 

Good guess on internal: it appeared in Go 1.4, December 2014.

PR #709, opened 2024-07-22
Skip checking `timestamp` and `bom-ref` fields in `check-augmented-sbom` for 100.10.0 release

tdq45gj:


autarch (changes requested):
Instead of disabling this check entirely, we can change the code in the `regenerate-and-diff-augmented-sbom.sh` to pass `--ignore-matching-lines` for the fields that are spuriously changed, `timestamp` and `bom-ref`.

autarch (changes requested):
This is the check for the SBOM Lite, not the Augmented SBOM (there's two SBOM files). You need to modify the `regenerate-and-diff-augmented-sbom.sh` script.

autarch (approved):
LGTM!

PR #715, opened 2024-07-26
Add script for podman init and start

tdq45gj:


buildscript/deps.go
```diff
@@ -50,6 +55,44 @@ func requirePodman(ctx *task.Context) error {
 	return err
 }
 
+var podmanCurrentMachineRegexp = regexp.MustCompile(`currentmachine:\s*"?([^"\s]*)"?`)
+var podmanMachineStateRegexp = regexp.MustCompile(`machinestate:\s*"?([^"\s]*)"?`)
+
+func startPodmanMachine(ctx *task.Context) error {
```
autarch, line 61: This needs to be a no-op on Linux. There's no VM needed there.

buildscript/deps.go
```diff
@@ -50,6 +55,44 @@ func requirePodman(ctx *task.Context) error {
...
+	out, err := sh.RunOutput(ctx, "podman", "machine", "info")
```
autarch, line 62: I'm guessing this command takes a `--format json` argument which you could use instead of parsing the output with regexes.

buildscript/deps.go
```diff
@@ -50,6 +55,44 @@ func requirePodman(ctx *task.Context) error {
...
+		return errors.Errorf("failed to parse `machinestate` field from podman machine info")
+	}
+	if machineStates[1] == "Running" {
+		return nil
+	}
+
+	return sh.Run(ctx, "podman", "machine", "start")
+}
+
+func stopPodmanMachine(ctx *task.Context) error {
```
autarch, line 92: This should also be a no-op on Linux.

autarch (changes requested):
I had a few comments.

autarch (approved):

PR #722, opened 2024-08-30
Update release docs for third party vulnerability handling

autarch:


tdq45gj (approved):
LGTM, although I still don't have access to the Silk app.

autarch:
> LGTM, although I still don't have access to the Silk app.

There's a DEVPROD ticket to tie this to Mana groups, but for now I think you can just request access if you are doing a release.

PR #723, opened 2024-08-30
TOOLS-3642 Document Security Level of TOOLS tickets

tdq45gj:
This PR documents a guideline for setting the Security Level of tools tickets.

### DIFF
--- a/ticket-security-level.md
+++ b/ticket-security-level.md
@@ -1,0 +1,27 @@
+# Tools Tickets Security Level
+
+This document describes the security level for Jira tickets under the [TOOLS project](https://jira.mongodb.org/projects/TOOLS/issues). 
+Every TOOLS ticket is by default viewable by public users, unless `Security Level` is set to `Mongo Internal`. 
+Once a ticket is marked as `Mongo Internal`, it cannot be viewed by external users and should be excluded from the 
+[changelog](CHANGELOG.md) after every release. This document outlines what tickets should be internal vs external.
+
+## Investigation of HELP Tickets
+
+TOOLS investigation ticket of HELP tickets are created with full ticket description from the HELP ticket, 
+which might contain links or logs specific to customers or internal teams of MongoDB. 
+These tickets are **strictly prohibited** from being accessed publicly.
+
+## Internal Processes
+
+Tickets are created for an internal process changes that is not tied with the code base should in general be marked as 
+internal. Examples include:
+- [TOOLS-3642](https://jira.mongodb.org/browse/TOOLS-3642) 
+- [TOOLS-3644](https://jira.mongodb.org/browse/TOOLS-3644)
+
+## Vulnerabilities
+
+Vulnerabilities are created as internal by default. But once they're resolved, either being mitigated or 
+being marked as false positive, they should be publicly viewable following a new release. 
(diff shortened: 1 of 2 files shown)


ticket-security-level.md
```diff
@@ -0,0 +1,31 @@
+# Tools Tickets Security Level
+
+This document describes the security level for Jira tickets under the
+[TOOLS project](https://jira.mongodb.org/projects/TOOLS/issues). Every TOOLS ticket is by default
+viewable by public users, unless `Security Level` is set to `Mongo Internal`. Once a ticket is
```
autarch, line 5:
```suggestion
viewable by public users, unless the `Security Level` field on the ticket is set to `Mongo Internal`. Once a ticket is
```

ticket-security-level.md
```diff
@@ -0,0 +1,31 @@
...
+marked as `Mongo Internal`, it cannot be viewed by external users and should be excluded from the
+[changelog](CHANGELOG.md) after every release. This document outlines what tickets should be
```
autarch, line 7:
```suggestion
[changelog](CHANGELOG.md) when it's included in a release. This document describes which tickets should be
```

ticket-security-level.md
```diff
@@ -0,0 +1,31 @@
...
+internal vs external.
+
+## Investigation of HELP Tickets
+
+TOOLS investigation ticket of HELP tickets are created with full ticket description from the HELP
+ticket, which might contain links or logs specific to customers or internal teams of MongoDB. These
+tickets are **strictly prohibited** from being accessed publicly.
```
autarch, line 14:
```suggestion
TOOLS investigation ticket of HELP tickets are created with a copy of the full ticket description from the HELP
ticket, which might contain links or logs specific to customers or internal teams of MongoDB. These
tickets **must always be marked as `Mongo internal`**.
```

ticket-security-level.md
```diff
@@ -0,0 +1,31 @@
...
+
+- [TOOLS-3642](https://jira.mongodb.org/browse/TOOLS-3642)
+- [TOOLS-3644](https://jira.mongodb.org/browse/TOOLS-3644)
+
+## Vulnerabilities
+
+Vulnerabilities are created as internal by default. But once they're resolved, either being
+mitigated or being marked as false positive, they should be publicly viewable following a new
+release. Our [release process](RELEASE.md) requires the release manager to mark any resolved
+vulnerability tickets as public and include them in the [changelog](CHANGELOG.md). Examples include:
```
autarch, line 29:
```suggestion
Vulnerabilities are created as internal by default. However, once they're resolved, whether they've been fixed or marked as a false positive, they should be publicly viewable **after** the fix is released. Our [release process](RELEASE.md) requires the release manager to include them in the [changelog](CHANGELOG.md) and to update their `Security Level` to make them public after they are included in a release. Examples include:
```

autarch, line 29: Actually, the release process doesn't call out changing the `Security Level` field. It just says they should be included in the changelog. Can you update the release process docs to also call out changing the field?

autarch (changes requested):
This looks good % some grammar/wording stuff. And there's one small change we should make to the release process docs as well.

autarch (approved):
LGTM!

PR #731, opened 2024-10-29
Fix Ubi download

FGasper:
[ubi](https://github.com/houseabsolute/ubi) recently changed its bootstrap mechanism, which has broken linting in Evergreen CI.

This changeset fixes that by downloading the bootstrap script for the ubi version that this project’s build system uses.


autarch (approved):
LGTM!

PR #741, opened 2025-01-22
TOOLS-3637 Mongodump should check collection metadata consistency

janetio:


common/bsonutil/bsonutil.go
```diff
@@ -475,9 +475,17 @@ func MtoD(m bson.M) bson.D {
...
+// to ExtJSON objects return a consistent BSON object when unmarshaled. This is
+// to ensure collection metadata consistency with mongorestore.
 //
 // It is preferred to be used in mongodump to avoid generating un-reversible ext JSON.
-func MarshalExtJSONReversible(val interface{}, canonical bool, escapeHTML bool) ([]byte, error) {
+func MarshalExtJSONReversible(
+	val interface{},
+	canonical bool,
+	escapeHTML bool,
+	forMongodump bool,
```
janetio, line 487: I could also just make a new function (?) that calls `MarshalExtJSONReversible` + the new BSON equality check logic. Decided to go with this for now, but I think it might create extra confusion with all the booleans passed into this function. However, I'm also not sure if we need to make a new mongodump specific function. Please let me know what you think.

common/bsonutil/bsonutil.go
```diff
@@ -486,5 +494,24 @@ func MarshalExtJSONReversible(val interface{}, canonical bool, escapeHTML bool)
...
 		return nil, errors2.Wrap(unmarshalErr, "marshal is not reversible")
 	}
+
+	if !forMongodump {
+		return jsonBytes, nil
+	}
+
+	originalBSON, err := bson.Marshal(val)
+	if err != nil {
+		return jsonBytes, nil
```
janetio, line 504: We return jsonBytes if we can't marshal `val` into BSON for non-BSON marshalable interfaces.

common/bsonutil/bsonutil.go
```diff
@@ -486,5 +494,24 @@ func MarshalExtJSONReversible(val interface{}, canonical bool, escapeHTML bool)
...
+	}
+	reversedBSON, err := bson.Marshal(reversedVal)
+	if err != nil {
+		return jsonBytes, nil
```
janetio, line 508: Same comment as above.

common/bsonutil/bsonutil_test.go
```diff
@@ -103,35 +103,48 @@ func TestMarshalExtJSONReversible(t *testing.T) {
...
+		reversible           bool
+		convertibleToExtJSON bool
+		expectedJSON         string
 	}{
+		{
+			bson.M{"field1": bson.M{"abc": int64(123)}},
+			true,
+			true,
+			true,
+			`{"field1":{"abc":{"$numberLong":"123"}}}`,
```
janetio, line 117: Added test case for a BSON object that should not error.

common/bsonutil/bsonutil.go
tdq45gj: This is inside `common` library, the function shouldn't contain care if it's for mongodump or not. Maybe a new function with a name such as `MarshalExtJSONWithBSONRoundtripConsistency`.

common/bsonutil/bsonutil.go
```diff
@@ -486,5 +494,24 @@ func MarshalExtJSONReversible(val interface{}, canonical bool, escapeHTML bool)
...
+	}
+	if !bytes.Equal(originalBSON, reversedBSON) {
+		return nil, fmt.Errorf(
+			"inconsistent metadata detected during dump: marshaling BSON to ExtJSON and back resulted in discrepancies",
```
autarch, line 512: I don't think this error message really makes sense for a generic function like this. It doesn't know what it's marshaling, so calling it "metadata" seems wrong.

common/bsonutil/bsonutil.go
autarch: _Why_ is this being returned? Typically, in Go, a func that returns `(foo, error)` returns the zero value for `foo`.

autarch (changes requested):
This looks good overall but I had few comments & questions.

common/bsonutil/bsonutil.go
janetio: Ah okay I see, fixed.

common/bsonutil/bsonutil.go
janetio: Fixed to return nil. I thought there could be the possibility that the metadata was for some reason not BSON marshalable but still okay, but looked at the code again and realized that was never possible anyways.

common/bsonutil/bsonutil_test.go
```diff
@@ -138,3 +138,42 @@ func TestMarshalExtJSONReversible(t *testing.T) {
 		assert.Equal(t, test.expectedJSON, string(json))
 	}
 }
+
+func TestMarshalExtJSONWithBSONRoundtripConsistency(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	tests := []struct {
+		val                  any
+		convertibleToExtJSON bool
```
tdq45gj, line 147: I think the test cases are all convertible to Ext JSON. Can we name it something else? maybe `consistentAfterRoundtripping`.

tdq45gj (changes requested):
Looks good over all, although we're not using the new function anywhere in mongodump.

common/bsonutil/bsonutil_test.go
```diff
@@ -138,3 +138,42 @@ func TestMarshalExtJSONReversible(t *testing.T) {
...
+		expectedJSON         string
+	}{
```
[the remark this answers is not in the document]
janetio: Added!

common/bsonutil/bsonutil_test.go
janetio: Yep done.

common/bsonutil/bsonutil.go
```diff
@@ -477,14 +477,59 @@ func MtoD(m bson.M) bson.D {
...
+func MarshalExtJSONReversible(
+	val interface{},
+	canonical bool,
+	escapeHTML bool,
+) ([]byte, error) {
 	jsonBytes, err := bson.MarshalExtJSON(val, canonical, escapeHTML)
 	if err != nil {
 		return nil, err
 	}
+
```
tdq45gj, line 489:
```suggestion
```

common/bsonutil/bsonutil.go
```diff
@@ -477,14 +477,59 @@ func MtoD(m bson.M) bson.D {
...
 	reversedVal := reflect.New(reflect.TypeOf(val)).Elem().Interface()
+
```
tdq45gj, line 491:
```suggestion
```

tdq45gj (approved):
This looks good except the two empty lines. Great work, thanks!

common/bsonutil/bsonutil.go
```diff
@@ -477,14 +477,58 @@ func MtoD(m bson.M) bson.D {
 // but would return an error if it cannot be reversed by bson.UnmarshalExtJSON.
 //
 // It is preferred to be used in mongodump to avoid generating un-reversible ext JSON.
-func MarshalExtJSONReversible(val interface{}, canonical bool, escapeHTML bool) ([]byte, error) {
+func MarshalExtJSONReversible(
```
autarch, line 480: I found it a bit confusing that this func was changing in this PR, but I eventually realized this was purely cosmetic.

In these sorts of cases, I think there's a few ways to handle this:

1. Make a separate PR for cosmetic changes.
2. Keep it in the same PR, but call it out in the PR description so reviewers know why this is here. You can also add your own comment to the PR calling this out, which is helpful for reviewers.

autarch (approved):
LGTM!

PR #750, opened 2025-01-31
TOOLS-3745: Handle 'system.buckets.' prefix correctly when .bson is restore target 

shyam12860:
When the bson file is passed as the target to mongorestore,
CreateIntentForCollection does not set the collection name correctly
by removing the "system.buckets." prefix.
This will cause errors like:
"Invalid namespace: test.system.buckets.system.buckets" as a result of
adding the same prefix twice to the namespace.

Making this change in the processing of the file so that the intent can
be initialized correctly rather than remove the prefix in the intent, so
that all code that intentionally accesses the collection name without
the prefix (when dropping the collection for example) works correctly.

Bonus: update README.md with instructions for issues running binaries on
Mac OS


mongorestore/filepath.go
```diff
@@ -611,6 +611,9 @@ func (restore *MongoRestore) CreateIntentForCollection(
 	var isTimeseries bool
 	if strings.HasPrefix(bsonFile.Name(), "system.buckets.") {
 		isTimeseries = true
+		// the name of the collection should be without the prefix to allow for operations (like drop) which should be
+		// performed on the timeseries view and not the system.buckets collection.
+		collection = strings.TrimPrefix(collection, "system.buckets.")
```
shyam12860, line 616: Making the change here rather than in `handleBSONInsteadOfDirectory` because CreateIntentsForDB does something similar where it gets the collection as "system.buckets.*" from the filename but then sets the intent's C field correctly without the prefix

FGasper (approved):
LGTM, but the ticket (TOOLS-3745) needs to be the first thing in the PR title, e.g.:
> TOOLS-3745 Handle 'system.buckets.' prefix correctly when .bson is target

PR #754, opened 2025-02-04
TOOLS-3708 Fix incorrect conversion of small, negative float64 index keys

FGasper:
Previously mongorestore’s legacy index conversion logic considered all small-magnitude float64 index values to be ascending indexes. In fact, though, if such index values are negative, the index should be descending.

This fixes that and adds a test to prevent a regression.


common/bsonutil/indexes.go
```diff
@@ -101,8 +102,8 @@ func ConvertLegacyIndexKeys(indexKey bson.D, ns string) {
 				converted = true
 			}
 		case float64:
-			if math.Abs(v-float64(0)) < epsilon {
-				indexKey[j].Value = int32(1)
+			if math.Abs(v) < epsilon {
+				indexKey[j].Value = int32(lo.Ternary(v > 0, 1, -1))
```
shyam12860, line 106: Considering the other cases are converting 0 to int32(1) should this condition be `lo.Ternary(v >= 0, 1, -1)`?

FGasper: Ah yes, good catch!

shyam12860 (approved):

jsflax (approved):

PR #756, opened 2025-02-04
Update most of our dev tools to their latest versions

autarch:
I mostly did this to get a new `ubi` version, since that has a fix for an issue that one of our team members encountered.

This does _not_ upgrade the following:

* `eslint` - would require rewriting our config to a new format
* `golangci-lint` - would require fixing a bunch of new issues it found.
* `gosec` - also finds new issues, all of which appear to be spurious, but this deserves its own PR with a bit more attention.


shyam12860 (approved):

FGasper (approved):
LGTM. Thanks!

PR #758, opened 2025-02-13
TOOLS-3719: fail if --restoreDbUsersAndRoles used on wrong target

shyam12860:
If the target was dumped with --dumpDbUsersAndRoles then we would find the db specific users / roles and the intents would not be nil, so we can use that check after the intents have been created to make sure that this option is used correctly.

### DIFF
--- a/mongorestore/mongorestore.go
+++ b/mongorestore/mongorestore.go
@@ -475,6 +475,19 @@ func (restore *MongoRestore) Restore() Result {
 			"remove the 'config' directory from the dump directory first")}
 	}
 
+	// if --restoreDbUsersAndRoles is used then db specific users and roles ($admin.system.users.bson / $admin.system.roles.bson)
+	// should exist in target directory / archive
+	if restore.InputOptions.RestoreDBUsersAndRoles &&
+		restore.ToolOptions.Namespace.DB != "" &&
+		(restore.manager.Users() == nil && restore.manager.Roles() == nil) {
+		return Result{
+			Err: fmt.Errorf(
+				"cannot find users or roles to restore with --restoreDbUsersAndRoles - " +
+					"use the archive or db directory dumped with --dumpDbUsersAndRoles as the target",
+			),
+		}
+	}
+
 	if restore.InputOptions.OplogFile != "" {
 		err = restore.CreateIntentForOplog()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


mongorestore/mongorestore.go
```diff
@@ -475,6 +475,19 @@ func (restore *MongoRestore) Restore() Result {
...
 
+	// if --restoreDbUsersAndRoles is used then db specific users and roles ($admin.system.users.bson / $admin.system.roles.bson)
+	// should exist in target directory / archive
+	if restore.InputOptions.RestoreDBUsersAndRoles &&
+		restore.ToolOptions.Namespace.DB != "" &&
+		(restore.manager.Users() == nil && restore.manager.Roles() == nil) {
+		return Result{
+			Err: fmt.Errorf(
+				"cannot find users or roles to restore with --restoreDbUsersAndRoles - " +
+					"restore target should be a dump created with --dumpDbUsersAndRoles",
```
FGasper, line 486: Can this message be a bit more helpful as to guiding the user forward?

Maybe:
> This archive does not contain users or roles. `--restoreDbUsersAndRoles` requires an archive created with mongodump’s `--dumpDbUsersAndRoles` option. To restore this archive, omit `--restoreDbUsersAndRoles`.

(Too wordy?)

FGasper (changes requested):
A couple small notes. See what you think.

Thanks!

mongorestore/mongorestore.go
shyam12860: Good idea. Updated with a message that gives the users a clear next step

mongorestore/mongorestore_test.go
```diff
@@ -1071,6 +1072,67 @@ func TestRestoreUsersOrRoles(t *testing.T) {
 				So(collName, ShouldNotEqual, "temproles")
 			}
 		})
+
+		Convey("If --dumpUsersAndRoles was not used with the target", func() {
+			expectedError := "cannot find users or roles to restore with --restoreDbUsersAndRoles - " +
```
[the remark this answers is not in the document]
shyam12860: Good call, I've updated, but let me know if there's a better way to do it.

test/qa-tests/jstests/restore/empty_users_and_roles.js
```diff
@@ -1,33 +0,0 @@
-(function() {
```
jsflax, line 1: Why is this being removed?

jsflax (approved):
One question, otherwise LGTM.

FGasper (changes requested):
LGTM % using errors.Is to test errors.

Thanks!

FGasper (approved):
LGTM.

Thanks!

PR #764, opened 2025-03-03
Remove columnstore index tests

shyam12860:
This feature was removed from the server in
https://jira.mongodb.org/browse/SERVER-91347
and we skipped them in https://jira.mongodb.org/browse/TOOLS-3252 so we
don't need to keep these around anymore.


autarch (approved):
LGTM!

jsflax (approved):

mongorestore/mongorestore_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
```
FGasper, line 14: Odd that this got _added_ as an import.

FGasper (approved):
LGTM. It’d be good, though, to know why that `errors` import got added. I don’t see any new usage of that module.

mongorestore/mongorestore_test.go
shyam12860: Might be because I accepted github's changes in https://github.com/mongodb/mongo-tools/pull/758? 

But the tests passed in that one without this import 🤔

PR #767, opened 2025-03-12
TOOLS-3540 Fix handling of empty timestamps

FGasper:
Because the server generally interprets Timestamp(0, 0) to mean “current time”, previously mongorestore has errantly been restoring backups with this value incorrectly.

The server has added an experimental mechanism to preserve such values easily; however, that is only available in recent builds. For 4.4 and 4.2, though—which mongorestore still supports—we need a backup strategy.

It happens that aggregation pipelines will, in certain configurations, apply zero timestamps literally. This changeset uses that in order to achieve correct restorations of zero timestamps in 4.4 and 4.2: first it inserts the document, then it updates it with a $replaceRoot aggregation.


common/db/db.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	mapset "github.com/deckarep/golang-set/v2"
```
jsflax, line 26: What is this syntax? Package alias?

common/db/empty_timestamp.go
```diff
@@ -0,0 +1,27 @@
+package db
+
+// MongoCanAcceptLiteralZeroTimestamp indicates whether the given server
+// version can accept a literal zero timestamp in a query. See SERVER-88750.
+func MongoCanAcceptLiteralZeroTimestamp(version Version) bool {
```
jsflax, line 5: Probably also worth referencing this TOOLS ticket.

mongorestore/empty_timestamp.go
```diff
@@ -0,0 +1,152 @@
...
+var nineNuls = make([]byte, 9, 9)
+
+// FindZeroTimestamps returns all document paths (each represented as
+// a []string) that contain a zero-value timestamp.
+func FindZeroTimestamps(raw bson.Raw) ([][]string, error) {
+	if !bytesSuggestEmptyTimestamp(raw) {
+		return nil, nil
+	}
+
+	// TODO: log(trace, "found empty-timestamp byte sequence")
```
jsflax, line 20: Leftover? Why can't we trace this?

common/db/db.go
FGasper: Yes. `mapset` is the normal alias for this package.

common/db/empty_timestamp.go
FGasper: done

jsflax (approved):

mongorestore/empty_timestamp.go
FGasper: This was a note-to-self that I wrote while writing this function, but I’m not sure it’s worth logging about TBH.

PR #768, opened 2025-03-12
REP-5475 - mongodump --internalOnlySourceWritesDoneBarrier

rcownie:
REP-5475 - Changes required for the mongodump passthrough testing

New mongodump option --internalOnlySourceWritesDoneBarrier to synchronize with resmoke tests

mongodump/barrier.go implements the wait-until-named-file-exists barrier

### DIFF
--- a/mongodump/barrier.go
+++ b/mongodump/barrier.go
@@ -1,0 +1,44 @@
+// Copyright (C) MongoDB, Inc. 2014-present.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License. You may obtain
+// a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+package mongodump
+
+import (
+	"os"
+	"time"
+
+	"github.com/mongodb/mongo-tools/common/log"
+)
+
+// Wait until a file exists and can be opened for reading
+// This is used only for testing mongodump/mongorestore with resmoke
+// test infrastructure.  The tests will create the barrier file when they have
+// finshed writes to the source cluster.
+func waitForSourceWritesDoneBarrier(barrierName string) {
+	log.Logvf(log.Always, "waitForSourceWritesDoneBarrier: %s", barrierName)
+	start := time.Now()
+	logInterval := time.Minute
+	prevLogTime := start
+	for {
+		f, err := os.Open(barrierName)
+		if err == nil {
+			// We opened the file for reading, so it does exist.
+			f.Close()
+			return
+		}
+		if os.IsNotExist(err) {
+			if time.Since(prevLogTime) >= logInterval {
+				prevLogTime = time.Now()
+				log.Logvf(log.Always, "waitForSourceWritesDoneBarrier: waited %.1f secs for %s",
+					prevLogTime.Sub(start).Seconds(),
+					barrierName)
+			}
+			// Poll for existence of the barrier file every 500msec
+			time.Sleep(500 * time.Millisecond)
+		} else {
+			panic(err)
+		}
+	}
+}
--- a/mongodump/barrier.go
+++ b/mongodump/barrier.go
@@ -10,7 +10,7 @@ import (
 	"time"
 
 	"github.com/mongodb/mongo-tools/common/log"
-    "github.com/pkg/errors"
+	"github.com/pkg/errors"
 )
 
 // Wait until a file exists and can be opened for reading
@@ -18,7 +18,11 @@ import (
 // test infrastructure.  The tests will create the barrier file when they have
 // finshed writes to the source cluster.
 func waitForSourceWritesDoneBarrier(barrierName string) {
-	log.Logvf(log.DebugHigh, "waitForSourceWritesDoneBarrier: initial check for existence of file %s", barrierName)
+	log.Logvf(
+		log.DebugHigh,
+		"waitForSourceWritesDoneBarrier: initial check for existence of file %s",
+		barrierName,
+	)
 	start := time.Now()
 	logInterval := time.Minute
 	prevLogTime := start
@@ -27,15 +31,22 @@ func waitForSourceWritesDoneBarrier(barrierName string) {
 		if err == nil {
 			// We opened the file for reading, so it does exist.
 			f.Close()
-	        log.Logvf(log.DebugHigh, "waitForSourceWritesDoneBarrier: barrier file %s exists - proceed past the barrier", barrierName)
+			log.Logvf(
+				log.DebugHigh,
+				"waitForSourceWritesDoneBarrier: barrier file %s exists - proceed past the barrier",
+				barrierName,
+			)
 			return
 		}
 		if os.IsNotExist(err) {
 			if time.Since(prevLogTime) >= logInterval {
 				prevLogTime = time.Now()
-				log.Logvf(log.DebugHigh, "waitForSourceWritesDoneBarrier: still waiting for existence of file %s after %.1f sec",
+				log.Logvf(
+					log.DebugHigh,
+					"waitForSourceWritesDoneBarrier: still waiting for existence of file %s after %.1f sec",
 					barrierName,
-					prevLogTime.Sub(start).Seconds())
+					prevLogTime.Sub(start).Seconds(),
+				)
 			}
 			// Poll for existence of the barrier file every 500msec
 			time.Sleep(500 * time.Millisecond)
--- a/mongodump/barrier.go
+++ b/mongodump/barrier.go
@@ -1,4 +1,4 @@
-// Copyright (C) MongoDB, Inc. 2014-present.
+// Copyright (C) MongoDB, Inc. 2025-present.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"); you may
 // not use this file except in compliance with the License. You may obtain
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/mongodb/mongo-tools/common/log"
+    "github.com/pkg/errors"
 )
 
 // Wait until a file exists and can be opened for reading
@@ -17,7 +18,7 @@ import (
 // test infrastructure.  The tests will create the barrier file when they have
 // finshed writes to the source cluster.
 func waitForSourceWritesDoneBarrier(barrierName string) {
-	log.Logvf(log.Always, "waitForSourceWritesDoneBarrier: %s", barrierName)
+	log.Logvf(log.DebugHigh, "waitForSourceWritesDoneBarrier: initial check for existence of file %s", barrierName)
 	start := time.Now()
 	logInterval := time.Minute
 	prevLogTime := start
@@ -26,19 +27,20 @@ func waitForSourceWritesDoneBarrier(barrierName string) {
 		if err == nil {
 			// We opened the file for reading, so it does exist.
 			f.Close()
+	        log.Logvf(log.DebugHigh, "waitForSourceWritesDoneBarrier: barrier file %s exists - proceed past the barrier", barrierName)
 			return
 		}
 		if os.IsNotExist(err) {
 			if time.Since(prevLogTime) >= logInterval {
 				prevLogTime = time.Now()
-				log.Logvf(log.Always, "waitForSourceWritesDoneBarrier: waited %.1f secs for %s",
-					prevLogTime.Sub(start).Seconds(),
-					barrierName)
+				log.Logvf(log.DebugHigh, "waitForSourceWritesDoneBarrier: still waiting for existence of file %s after %.1f sec",
+					barrierName,
+					prevLogTime.Sub(start).Seconds())
 			}
 			// Poll for existence of the barrier file every 500msec
 			time.Sleep(500 * time.Millisecond)
 		} else {
-			panic(err)
+			panic(errors.Wrapf(err, "failed to open barrier file %s", barrierName))
 		}
 	}
 }
--- a/mongodump/barrier.go
+++ b/mongodump/barrier.go
@@ -55,7 +55,7 @@ func waitForSourceWritesDoneBarrier(barrierName string) error {
 			time.Sleep(500 * time.Millisecond)
 		} else {
 			// Any other error implies that the resmoke test environment is
-			// irretrievably confused, so it is appropriate to panic here.
+			// irretrievably confused.  The caller must check the error.
 			return errors.Wrapf(err, "failed to open barrier file %#q", barrierName)
 		}
 	}
--- a/mongodump/barrier.go
+++ b/mongodump/barrier.go
@@ -17,7 +17,7 @@ import (
 // This is used only for testing mongodump/mongorestore with resmoke
 // test infrastructure.  The tests will create the barrier file when they have
 // finshed writes to the source cluster.
-func waitForSourceWritesDoneBarrier(barrierName string) {
+func waitForSourceWritesDoneBarrier(barrierName string) error {
 	// This code should only run in the resmoke testing environment. It's harmless and
 	// possibly useful to have verbose logging for testing; and if it accidentally runs
 	// in production, we want to see that in the logs so it's good to be verbose.
@@ -39,7 +39,7 @@ func waitForSourceWritesDoneBarrier(barrierName string) {
 				"waitForSourceWritesDoneBarrier: barrier file %#q exists - proceed past the barrier",
 				barrierName,
 			)
-			return
+			return nil
 		}
 		if os.IsNotExist(err) {
 			if time.Since(prevLogTime) >= logInterval {
@@ -56,7 +56,7 @@ func waitForSourceWritesDoneBarrier(barrierName string) {
 		} else {
 			// Any other error implies that the resmoke test environment is
 			// irretrievably confused, so it is appropriate to panic here.
-			panic(errors.Wrapf(err, "failed to open barrier file %#q", barrierName))
+			return errors.Wrapf(err, "failed to open barrier file %#q", barrierName)
 		}
 	}
 }
--- a/mongodump/barrier.go
+++ b/mongodump/barrier.go
@@ -18,9 +18,12 @@ import (
 // test infrastructure.  The tests will create the barrier file when they have
 // finshed writes to the source cluster.
 func waitForSourceWritesDoneBarrier(barrierName string) {
+	// This code should only run in the resmoke testing environment. It's harmless and
+	// possibly useful to have verbose logging for testing; and if it accidentally runs
+	// in production, we want to see that in the logs so it's good to be verbose.
 	log.Logvf(
-		log.DebugHigh,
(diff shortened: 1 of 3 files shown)


rcownie:
https://jira.mongodb.org/browse/REP-5475

mongodump/barrier.go
```diff
@@ -0,0 +1,44 @@
...
+	"time"
+
+	"github.com/mongodb/mongo-tools/common/log"
+)
+
+// Wait until a file exists and can be opened for reading
+// This is used only for testing mongodump/mongorestore with resmoke
+// test infrastructure.  The tests will create the barrier file when they have
+// finshed writes to the source cluster.
+func waitForSourceWritesDoneBarrier(barrierName string) {
```
ege-mgdb, line 19: Is this barrier file system part of the resmoke test infrastructure?

rcownie: Yes.  In the resmoke passthrough infrastructure, JS tests do updates to the srcCluster for a while, at some point the cluster_to_cluster_replication.py calls replicator.start() to start the replication (by mongosync or other means),
and calls replicator.commit() when the JS tests have finished doing updates.

For mongodump without --oplog, the mongodump can't start copying collections until after the updates have finished.

For mongodump with --oplog, the mongodump can copy collections, while further updates are being captured in the oplog, but it can't choose the end time in the oplog until updates have finished.

So either way, it has to wait at a barrier until there's a call to replicator.commit() in the resmoke python code.

Creating a file seemed as simple as anything else.

ege-mgdb: One question I have here is if it's possible for the file barrier to exist when it shouldn't? Then this code would start executing before the resmoke logic was complete?

If such a thing happened, how would we know when debugging a failed test?

Likewise, if the file is never created for some reason, how would the code behave and how would we debug it?

ege-mgdb: Sorry, for unresolving this conversation: I had thought you meant that this file barrier system already existed as part of resmoke and was commonly used, but, IIUC, what this means is that it's being created to support the mongodump passthrough testing.

rcownie: That would be bad, but the MongodumpReplicatorFixture creates a new temporary directory for each run of mongodump, and uses a barrier file name in that directory, so there's only a small window for a spurious barrier file to get accidentally created, and whoever did it would need to guess an obscure auto-generated temp_dir name.

mongodump/barrier.go
```diff
@@ -0,0 +1,44 @@
...
+	log.Logvf(log.Always, "waitForSourceWritesDoneBarrier: %s", barrierName)
```
ajayvijayakumar123, line 20:
```suggestion
	log.Logvf(log.DebugHigh, "waitForSourceWritesDoneBarrier: %s", barrierName)
```

I'm not too familiar with how we do logging in the tools. But, should this be at the debug level since we won't use this flag in production?

mongodump/barrier.go
```diff
@@ -0,0 +1,44 @@
...
+			if time.Since(prevLogTime) >= logInterval {
+				prevLogTime = time.Now()
+				log.Logvf(log.Always, "waitForSourceWritesDoneBarrier: waited %.1f secs for %s",
+					prevLogTime.Sub(start).Seconds(),
+					barrierName)
+			}
+			// Poll for existence of the barrier file every 500msec
+			time.Sleep(500 * time.Millisecond)
+		} else {
+			panic(err)
```
ajayvijayakumar123, line 41: can we wrap this error with an informative message? (ex: "failed to open barrier file ___ during resmoke testing")

mongodump/barrier.go
```diff
@@ -0,0 +1,44 @@
...
+		f, err := os.Open(barrierName)
+		if err == nil {
+			// We opened the file for reading, so it does exist.
+			f.Close()
+			return
+		}
+		if os.IsNotExist(err) {
+			if time.Since(prevLogTime) >= logInterval {
+				prevLogTime = time.Now()
+				log.Logvf(log.Always, "waitForSourceWritesDoneBarrier: waited %.1f secs for %s",
```
ajayvijayakumar123, line 34:
```suggestion
				log.Logvf(log.Always, "waitForSourceWritesDoneBarrier: waited %.1f secs for %s to exist",
```

mongodump/barrier.go
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) MongoDB, Inc. 2014-present.
```
ajayvijayakumar123, line 1: Can we add internal docs on how the barrier file coordination works for --oplog and normal mongodump? We could add something similar to what we talked about in this slack thread: https://mongodb.slack.com/archives/C02D30K5WN7/p1741961728834479

mongodump/barrier.go
```diff
@@ -0,0 +1,44 @@
...
+//
```
ajayvijayakumar123, line 2: Can we add a reviewer who is familiar with the tools repo? (Dave, Jian, Felipe) ?

ajayvijayakumar123 (changes requested):
Some suggestions!

mongodump/barrier.go
```diff
@@ -0,0 +1,44 @@
...
+	prevLogTime := start
+	for {
+		f, err := os.Open(barrierName)
+		if err == nil {
+			// We opened the file for reading, so it does exist.
+			f.Close()
+			return
+		}
+		if os.IsNotExist(err) {
+			if time.Since(prevLogTime) >= logInterval {
```
ege-mgdb, line 32: just to make sure I understand, this setup will not print a log statement until 1 minute after it has actually started waiting on the barrier file.  And then this works in tandem with the log statement at the top of the function to indicate we have entered and are still in the blocking state?

mongodump/barrier.go
ege-mgdb: Can we update the log message so that it's clear that this log means we are entering the blocking state and the later one means we are still in the blocking state.

ege-mgdb (changes requested):

mongodump/barrier.go
rcownie: Yes.  mongodump will hang waiting for the file, the replication will never be able to reach COMMITTED state.

mongodump/barrier.go
rcownie: I don't think that's necessary for this small change, adding code that will never run in production.

mongodump/barrier.go
rcownie: Done and done

mongodump/barrier.go
rcownie: Done

mongodump/barrier.go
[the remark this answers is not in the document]
rcownie: Yes.  The testing framework has to signal when the source cluster is in its final state.  Otherwise there will be a race and mongodump may capture the state too early, when it's still being changed.

mongodump/barrier.go
ajayvijayakumar123: I still would prefer having someone more familiar with mongo-tools tagged as a reviewer (especially since I'm not very familiar with the codebase; I'm assuming the same with Erich too). If they feel it's not necessary, they can remove themselves. Does that sound good?

ajayvijayakumar123 (changes requested):

mongodump/barrier.go
```diff
@@ -0,0 +1,57 @@
...
+					log.DebugHigh,
+					"waitForSourceWritesDoneBarrier: still waiting for existence of file %s after %.1f sec",
+					barrierName,
+					prevLogTime.Sub(start).Seconds(),
+				)
+			}
+			// Poll for existence of the barrier file every 500msec
+			time.Sleep(500 * time.Millisecond)
+		} else {
+			panic(errors.Wrapf(err, "failed to open barrier file %s", barrierName))
```
FGasper, line 54: Panicking seems a bit extreme here. Can we return an error instead?

(And use `%#q` in the error string to get a nicely-quoted name?)

mongodump/barrier.go
```diff
@@ -0,0 +1,57 @@
...
+				barrierName,
+			)
+			return
+		}
+		if os.IsNotExist(err) {
+			if time.Since(prevLogTime) >= logInterval {
+				prevLogTime = time.Now()
+				log.Logvf(
+					log.DebugHigh,
+					"waitForSourceWritesDoneBarrier: still waiting for existence of file %s after %.1f sec",
```
FGasper, line 46:
```suggestion
					"waitForSourceWritesDoneBarrier: still waiting for existence of file %#q after %.1f sec",
```

mongodump/barrier.go
```diff
@@ -0,0 +1,57 @@
...
+	log.Logvf(
+		log.DebugHigh,
+		"waitForSourceWritesDoneBarrier: initial check for existence of file %s",
+		barrierName,
+	)
+	start := time.Now()
+	logInterval := time.Minute
+	prevLogTime := start
+	for {
+		f, err := os.Open(barrierName)
```
FGasper, line 30: Nit: Do we care, really, about being able to _read_ the file, versus its mere existence?

If all we need to verify is its existence, without any need to _interact_ with the file, then a Stat() seems more apropos.

mongodump/mongodump.go
```diff
@@ -430,6 +440,15 @@ func (dump *MongoDump) Dump() (err error) {
 	// we check to see if the oplog has rolled over (i.e. the most recent entry when
```
FGasper, line 440: This comment seems a bit far removed from its subject. Should be moved now that there’s the intervening barrier check?

mongodump/barrier.go
```diff
@@ -0,0 +1,57 @@
...
+func waitForSourceWritesDoneBarrier(barrierName string) {
+	log.Logvf(
+		log.DebugHigh,
+		"waitForSourceWritesDoneBarrier: initial check for existence of file %s",
+		barrierName,
+	)
+	start := time.Now()
+	logInterval := time.Minute
+	prevLogTime := start
+	for {
```
FGasper, line 29: A setup that’s more idiomatic to Go—and more scalable—would be to have two `time.Ticker`s: one that fires every 500ms to check for the file, and another that fires every 1m to output the log. Then your loop does a `select` on those along with ctx.Done().

It would _probably_ read better since then you don’t have to track the `prevLogTime`.

mongodump/barrier.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/mongodb/mongo-tools/common/log"
+	"github.com/pkg/errors"
+)
+
+// Wait until a file exists and can be opened for reading
+// This is used only for testing mongodump/mongorestore with resmoke
+// test infrastructure.  The tests will create the barrier file when they have
+// finshed writes to the source cluster.
+func waitForSourceWritesDoneBarrier(barrierName string) {
```
FGasper, line 20: This should accept a context, and whatever blocks in here should take the context into account. See my note below about making the `for` loop more idiomatic.

It should also probably return an error rather than panicking (see other note below).

FGasper (changes requested):
This generally looks good. I’ve made some notes; see what you think.

Thanks!

mongodump/barrier.go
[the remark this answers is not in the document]
ege-mgdb: (this may already be addressed with discussions you had with Dave)
My opinion here is that we should have avoid having code that exists just for testing get compiled into the binaries that are released. So, is it possible to have this mechanism only be a part of the integration testing build?

mongodump/barrier.go
ege-mgdb: Going to disagree with @ajayvijayakumar123: if this should never run in production this it's actually critical that it is as noisy as possible when running in production. Basically, if we hide the logs when this is running in production then we'll not be able to detect that something _which should never run in production_ is running in production until it breaks.

If possible, we should not have this code compile into production binaries so that it's actually impossible to run in production :).

ajayvijayakumar123: Ah yup good point. I agree

mongodump/mongodump.go
```diff
@@ -226,6 +226,16 @@ func (dump *MongoDump) Dump() (err error) {
 		return nil
 	}
 
+	if !dump.OutputOptions.Oplog && (dump.InputOptions.SourceWritesDoneBarrier != "") {
```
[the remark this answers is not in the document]
FGasper: Quibble: `verifyCollectionExists` is for backups of a specific collection. It’s a no-op otherwise. So either the backup will fail, or that collection will be included.

That said, I agree that the existence check should probably follow the barrier.

mongodump/barrier.go
rcownie: Undone and undone, and I added a comment to explain the choice of log.Always.

mongodump/barrier.go
rcownie: I think panic() is appropriate here, since this code is only used in the resmoke test environment and anything strange happening to that file would indicate some level of chaos in the test.

Changed %s to %#q throughout.

mongodump/barrier.go
rcownie: As discussed here, the code gets more complicated for no real benefit if you go that way and try to get it right.

https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go

mongodump/mongodump.go
rcownie: Done

mongodump/barrier.go
rcownie: It doesn't need to be scalable. And I don't think two Tickers is simpler than one variable for prevLogTime.  The Ticker approach would allow the two activities to operate with no phase relationship, but we don't need or even want that - if they got unsync'ed it would allow substantial windows of time when the file might already exist, but we print a message saying we're still waiting.  By printing the message shortly after doing the file check, that race becomes improbable.

mongodump/barrier.go
rcownie: Nothing in mongodump is using any kind of Context.

Your suggestions are appropriate to designing a general-purpose barrier mechanism for wider use.  That is not the goal here: it is code which lives in the mongodump repo and is used in mongodump only for one specific tactical purpose, and is designed to do precisely that and with no complexity beyond that.

mongodump/mongodump.go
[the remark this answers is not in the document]
rcownie: For the non-oplog case, moved the waitForBarrier before the check.

I added a comment about the oplog case.

    // If we were to test with the combination of --oplog --internalSourceWritesOnly
    // *and* a specified collection *and* the collection didn't exist at the time mongodump
    // was started but would be created later and possibly captured in the oplog, then this
    // check could fail. That is beyond the scope of the mongodump passthrough testing,
    // so we don't disrupt the mongodump code more than necessary.
	exists, err := dump.verifyCollectionExists()

mongodump/barrier.go
FGasper: We should design this particular piece to accept a context, then have callers pass in context.Background().

It is a defect of mongodump/restore that they generally don’t use Context. We shouldn’t propagate that to new code.

mongodump/barrier.go
FGasper: I would say it’s marginally simpler because you don’t have a file descriptor to close. You still have a command that can fail, and the error you check for is the same.

All you care about is existence. You might as well use the kernel API that gives you that information directly, IMO. It’s a small point, though, so I’ll close this thread.

mongodump/barrier.go
FGasper: > if they got unsync'ed it would allow substantial windows of time when the file might already exist, but we print a message saying we're still waiting

Generally contexts prevent this.

Anyhow, it’s a small point. Closing this thread.

mongodump/barrier.go
FGasper: General Go practice is to avoid panic() except for indicating purely-internal errors, usually programmer mistakes.

Maybe ping another reviewer for a third opinion?

mongodump/barrier.go
FGasper: That said, I suppose the only thing this function does that blocks (other than filesystem) is time.Sleep, which doesn’t take a context. (It surprises me that the stdlib still lacks a `time.SleepContext` or some other convenient way to do an interruptible wait.)

FGasper (changes requested):
I’d still like to see the panic converted to a returned error. @ajayvijayakumar123, @tdq45gj? (Erich, you don’t pop up as easily in @ completion …)

mongodump/barrier.go
rcownie: Made the change.

ege-mgdb:
> I’d still like to see the panic converted to a returned error. @ajayvijayakumar123, @tdq45gj? (Erich, you don’t pop up as easily in @ completion …)

I've kind of intentionally backed off these PRs bc there were getting to be too many people involved and I didn't to create distractions.

My general opinion on panics or faults is that they should be used when you have entered a state that is unrecoverable (that is, the program has entered a state that should be impossible and where it can no longer trust any decision it makes after that).

ajayvijayakumar123 (approved):
LGTM % a [suggestion](https://github.com/mongodb/mongo-tools/pull/768#discussion_r2016762671) !

mongodump/mongodump.go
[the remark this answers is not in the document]
rcownie: The medium/long-term goal is to get rid of mongodump and mongorestore, by supporting dump-to-file(s) and restore-from-file(s) in mongosync.  So I don't think there's a need to plan for extending the mongodump testing.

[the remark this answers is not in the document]
ajayvijayakumar123: Gotcha. I still think we note this as a non-goal on the scope to call out something that is supported but not testing

mongodump/barrier.go
```diff
@@ -0,0 +1,62 @@
...
+					"waitForSourceWritesDoneBarrier: still waiting for existence of file %#q after %.1f sec",
+					barrierName,
+					prevLogTime.Sub(start).Seconds(),
+				)
+			}
+			// Poll for existence of the barrier file every 500msec
+			time.Sleep(500 * time.Millisecond)
+		} else {
+			// Any other error implies that the resmoke test environment is
+			// irretrievably confused, so it is appropriate to panic here.
```
tdq45gj, line 58: The comment might need to be updated

tdq45gj (approved):
LGTM!

FGasper (approved):
LGTM % @tdq45gj’s request to update the comment about panicking.

mongodump/mongodump.go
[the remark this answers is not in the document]
rcownie: OK, doing that now.

mongodump/barrier.go
rcownie: Done

ege-mgdb (approved):

PR #770, opened 2025-03-13
TOOLS-3764: Run integration tests against server version 8.1

shyam12860:
This commit contains the changes required to run the mongo tools tests against mongo server version 8.1.

Changes to JS test builds:
After SERVER-81339, ShardingTest, and ReplSetTest are not shipped / loaded with the mongo shell by default. They are now modules that need to be imported into the shell environment. We're copying over the files `replsettest.js`,  `shardingtest.js` and `parallelTester.js` from the mongo repo into the `test/shell_common/libs/` directory. Small changes were required in both `ReplSetTest` and `ShardingTest` so that they can be used in the context of our JS tests. The changes to the original versions are listed at the top of each file.

We're loading the libs in `test/shell_common/` into the mongo shell for tests by calling `load_libs.8.1.js` in an `--eval` statement, and the evergreen builds and JS tests have been set up to look for a version specific `load_libs.X.js` file depending on server / shell version. `load_libs.8.1.js` expects the shell to be run from `test/qa-tests` or `test/legacy42`, so we're adding a `test/shell_common/jsconfig.json` that is hidden away during jstests, but moved into the base directory to run `mongo` from there during `create repl_set`. This is temporary, as we should just remove the use of `ReplSetTest` to create replica sets.

Additionally - `replsettest.8.0.js` was added because we thought it was required in the 8.0.5 shell, but tests are passing without it so we're removing it.

Notable changes to tests to work with 8.1.0 mongo server / shell:
- SERVER-83663 - The authSchema was removed in server 8.1, so we have to skip checking for it in `dumprestore_auth2.js`
- SERVER-100146 - We were calling `assert.eq` with more than a bunch of extra string arguments, expecting that they will all be joined into one, but instead the shell was just ignoring them. An unrelated change in 8.1 shell adding another 4th argument now causes assert calls to fail if they have too many arguments, so we're converting all the strings into a single string.
- SERVER-88158 - `applyOps` does not upsert by default, and therefore we need to insert the document into the collection first before trying to update it in `preserve_oplog_structure_order.js`
- SERVER-96103 - `mongofiles_get.js`, Windows specific ecaping the quotes in the get_id JSON input is not required for 8.1.0
- SERVER-95434 - `shardOptions` has been removed, so we use `rsOptions` instead in `topology_helper.js`. This should be a no-op for older shell versions that use `shardOptions`.


test/qa-tests/jstests/files/mongofiles_get.js
```diff
@@ -47,7 +48,8 @@ var testName = 'mongofiles_get';
 
     // ensure tool runs get_id without error
     var idAsJSON = '{"$oid":"' + fileId.valueOf() + '"}';
-    if (_isWindows()) {
+    // no need to escape the string for windows in shell version 8.1.0
```
shyam12860, line 51: Just noticed this. I've changed this comment to be clearer and include the relevant SERVER ticket, will push that change with other changes after review

common.yml
```diff
@@ -170,7 +180,7 @@ functions:
         rm -rf mongodb/db_files mongodb/${logfile|run.log}
         mkdir -p mongodb/db_files
         MONGOD_ARGS="${mongod_args}"
-        if [ "${USE_TLS}" = "true" ]; then
+        if [ "$ name{USE_TLS}" = "true" ]; then
```
FGasper, line 183: Should there be a space here?

common.yml
```diff
@@ -57,7 +59,15 @@ functions:
...
         cd ${test_path}
-        ./mongo --nodb lib/run_mongod.js jstests/tool/*.js
+        if [ -n "${load_libs_version}" ]; then
+          if [ "${load_libs_version}" == "latest" ]; then
+            # set to the maxServerVersion in download_server.go
+            IMPORT_LOAD_LIBS="await import(\"../shell_common/libs/load_libs-8.1.js\");"
+          else
+            IMPORT_LOAD_LIBS="await import(\"../shell_common/libs/load_libs-${load_libs_version}.js\");"
+          fi
+        fi
```
FGasper, line 69: This is nontrivial logic that gets repeated. Can it be put into a function?

common.yml
```diff
@@ -1245,6 +1264,96 @@ tasks:
...
           target: test:awsauth ${testArgs}
 
+  - name: integration-8.1
+    tags: ["8.1"]
+    commands:
+      - func: "fetch source"
+      - command: expansions.update
+      - func: "download mongod and shell"
+        vars:
+          mongo_version: "8.1.0-alpha3"
```
FGasper, line 1274: I guess there’s no way not to encode `alpha3` here? Would be nice not to have to …

common.yml
```diff
@@ -36,6 +36,8 @@ variables:
       mongo_args_tls: "--port 33333 --tls --tlsCAFile common/db/testdata/ca-ia.pem --tlsCertificateKeyFile common/db/testdata/test-server.pem --tlsAllowInvalidCertificates"
       mongod_port: 33333
 
+  max_server_version: "8.1"
```
FGasper, line 39: Where is this used?

test/legacy42/jstests/tool/dumprestore_auth2.js
```diff
@@ -108,11 +113,15 @@ var dumpRestoreAuth2 = function(backup_role, restore_role) {
...
     assert.eq(2, admindb.system.users.getIndexes().length, "didn't maintain user indexes");
     assert.eq(2, admindb.system.roles.getIndexes().length, "didn't maintain role indexes");
-    assert.eq(
-        1, admindb.system.version.find({_id: "authSchema"}).count(), "didn't restore version");
-    assert.docEq(versionDoc,
-                 admindb.system.version.findOne({_id: "authSchema"}),
-                 "version doc wasn't restored properly");
+    // authSchema document has been removed from server version 8.1.0+, no need to check for it
+    // load isAtLeastVersion from qa tests for the time being.
+    if (!isAtLeastVersion(admindb.version(), "8.1.0")) {
```
FGasper, line 118: Could put this into a `adminHasAuthSchemaDocument(client)` function to avoid repeating the `8.1.0`.

test/shell_common/jsconfig.json
FGasper, line 1: What consumes this? Is there some other format besides JSON that we can use for it so that we can add comments?

test/shell_common/libs/parallelTester-8.1.js
FGasper, line 1: We should note that this was copied from the server.

test/shell_common/libs/load_libs-8.1.js
```diff
@@ -0,0 +1,17 @@
...
+  TestData = new Object();
+}
+
+const {ReplSetTest} = await import('../shell_common/libs/replsettest-8.1.js');
+globalThis.ReplSetTest = ReplSetTest
+
+const {ShardingTest} = await import('../shell_common/libs/shardingtest-8.1.js');
+globalThis.ShardingTest = ShardingTest
+
+// in 8.1 shell rawMongoProgramOutput expects one parameter. Change it here specifically when running from 8.1 shell.
```
FGasper, line 15: Should clarify what that parameter is, and maybe link to the ticket that made the change.

FGasper (changes requested):
This looks really good! I’ve noted a few things; see what you think, then shoot it back to me.

common.yml
shyam12860: Looks like I was testing this _before_ 8.1.0-alpha3 made it to full.json, which means I needed to explicitly set the version here. Not required anymore.

common.yml
shyam12860: Looks like we can't call one function in another. Looking into this

test/shell_common/jsconfig.json
shyam12860: This data? It's in an oplog.bson file to be read by mongorestore.

Are you looking to add the same comments here closer to the data itself?

autarch (approved):
LGTM!

FGasper (approved):
LGTM. Thank you for all your work on this … I know this stuff is a haul.

PR #771, opened 2025-03-14
TOOLS-3782: handle missing authSchema doc in 8.1

shyam12860:
In SERVER-83663, the authSchema doc in system.version was removed because server 8.1 only supports one auth version (5) and the doc is not required anymore. mongodump and restore expect the authSchema document to exist in the server in order to use the auth version and determine the compatibility of auth versions between the source and destination servers.

Since there's only one authVersion in 8.1+, we set it to 5 based on the server version. In mongodump's VerifySystemAuthVersion, in place of checking the authSchema doc we check whether users and roles exist in the server, because that was the condition for an authSchema doc to have been written in older server versions.

Unfortunately, we don't currently have a way to find the dump's source server version in mongorestore in bson / directory mode. mongodump does not write the server's version into the dump unless it's used in --archive mode. As a result, we're making mongodump write the server version and tool version into a new prelude.json file in the dump directory, and mongorestore can read from this and use it to determine the dump's server version. If the dump's server version is 8.1.0+, we default to authSchema 5 in GetDumpAuthVersion.


mongodump/mongodump.go
```diff
@@ -946,6 +957,61 @@ func (dump *MongoDump) DumpMetadata() error {
...
+		// most likely means there was no data to dump
+		log.Logvf(log.DebugLow, "parent directory does not exist, not writing prelude.json")
+		return nil
+	}
+	log.Logvf(log.DebugLow, "dumping prelude metadata to file: %s", filename)
+
+	file, err := os.Create(filename)
+	if err != nil {
+		log.Logvf(log.DebugLow, "error opening file: %v", err)
+		return fmt.Errorf("error opening prelude.json file %v", filename)
```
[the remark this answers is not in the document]
FGasper: Also: wrap the error

mongorestore/mongorestore.go
```diff
@@ -96,8 +97,9 @@ type MongoRestore struct {
 	// This is initialized to os.Stdin if unset.
 	InputReader io.Reader
 
-	// Server version for version-specific behavior
-	serverVersion db.Version
+	// Server versions for version-specific behavior
+	dumpServerVersion string
+	serverVersion     db.Version
```
FGasper, line 102: It seems a bit confusing that two “server version” variables in the same struct would have different types. IDEs automate much of the “friction”, of course, but it still seems a bit strange.

Maybe it’s worth parsing dumpServerVersion?

mongorestore/mongorestore.go
```diff
@@ -396,6 +399,10 @@ func (restore *MongoRestore) Restore() Result {
 			}
 			return Result{Err: fmt.Errorf("mongorestore target '%v' invalid: %v", restore.TargetDirectory, err)}
 		}
+		if err = restore.ReadPreludeMetadata(target); err != nil {
+			// don't error out here because older dump versions will not prelude.json
```
FGasper, line 403: Here it’d be better to make ReadPreludeMetadata return (bool, error) to differentiate nonexistence from a bona-fide failure.

mongodump/mongodump.go
```diff
@@ -946,6 +957,61 @@ func (dump *MongoDump) DumpMetadata() error {
...
+	}
+	if dump.OutputOptions.Out == "" {
+		filename = filepath.Join("dump", filename)
+	} else {
+		filename = filepath.Join(dump.OutputOptions.Out, filename)
+	}
+	if dump.OutputOptions.Gzip {
+		filename += ".gz"
+	}
+	if _, err := os.Stat(filepath.Dir(filename)); err != nil {
```
FGasper, line 981: This check probably isn’t necessary. If this directory doesn’t exist, os.Create() below should fail with fs.ErrNotExist, so you can just check for that case.

mongorestore/mongorestore.go
```diff
@@ -662,6 +669,60 @@ func (restore *MongoRestore) Restore() Result {
...
+		if err != nil {
+			log.Logvf(log.DebugLow, "error creating gzip reader: %v", err)
+			return fmt.Errorf("could not read gzip file %s", filename)
+		}
+		defer zipfile.Close()
+		reader = zipfile
+	} else {
+		reader = file
+	}
+	bytes, err := io.ReadAll(reader)
```
FGasper, line 702: I’m _assuming_ that we don’t expect over-large files here … but if there _were_ to be such a file then this could cause an OOM. We can’t really prevent that with simple JSON, though.

shyam12860: Yeah considering we're control the prelude.json file, I figured we don't need to worry about the size of the file for the time being, because it's very small

mongorestore/mongorestore.go
shyam12860: Maybe the ideal is to remove the use of `db.Version` and use a string + semver.Compare throughout the codebase, but for the time being I've just added a `db.StrToVersion` method that will parse dumpSemverVersion.

mongorestore/mongorestore.go
FGasper: If EACCES or other “unexpected” cases happen, they need to take down the process. Thus, I think “don’t error out here …” doesn’t work.

If we want to _ignore_ the nonexistence case, then yeah we need to handle that specially.

mongodump/mongodump.go
```diff
@@ -365,16 +367,17 @@ func (dump *MongoDump) Dump() (err error) {
 		return fmt.Errorf("error dumping metadata: %v", err)
 	}
 
+	dump.serverVersion, err = dump.SessionProvider.ServerVersion()
```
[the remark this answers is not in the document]
shyam12860: I meant to move the error check with this line from inside the `if dump.OutputOptions.Archive != ""`. Currently we don't error out when the server version isn't available, so I'm keeping that behavior

mongodump/mongodump.go
```diff
@@ -946,6 +957,61 @@ func (dump *MongoDump) DumpMetadata() error {
 	return nil
 }
 
+// DumpPreludeMetadata dumps information about the server and the dump in json format
+// Currently only writes the server version and tool version, but we can use this to write other metadata about the dump in the future.
+func (dump *MongoDump) DumpPreludeMetadata() error {
+	preludeJson := map[string]string{
```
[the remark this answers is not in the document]
shyam12860: Non Go writer habits I think

common/auth/auth_info.go
```diff
@@ -63,6 +65,23 @@ func VerifySystemAuthVersion(sessionProvider *db.SessionProvider) error {
...
+		usersExist := session.Database("admin").
+			Collection("system.users").
+			FindOne(context.Background(), bson.D{})
+		rolesExist := session.Database("admin").
+			Collection("system.roles").
+			FindOne(context.Background(), bson.D{})
+		if errors.Is(usersExist.Err(), mongo.ErrNoDocuments) &&
+			errors.Is(rolesExist.Err(), mongo.ErrNoDocuments) {
+			return fmt.Errorf("no users / roles to dump")
+		}
```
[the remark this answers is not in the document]
FGasper: This error may look a bit strange because it won’t be clear which error came from which request.

Maybe just search for the users, check for failures, then search for roles, and check for failures.

I won’t block on this, but I do think it could be tidier.

mongodump/mongodump.go
```diff
@@ -946,6 +957,64 @@ func (dump *MongoDump) DumpMetadata() error {
...
+		defer writer.Close()
+	}
+	bytes, err := json.Marshal(preludeData)
+	if err != nil {
+		return fmt.Errorf("error marshaling prelude data: %w", err)
+	}
+
+	_, err = writer.Write(bytes)
+	if err != nil {
+		return fmt.Errorf("failed to write to prelude.json to file %v: %w", filename, err)
```
FGasper, line 1012:
```suggestion
		return fmt.Errorf("failed to write to prelude.json to file %#q: %w", filename, err)
```

mongorestore/mongorestore.go
```diff
@@ -396,6 +399,14 @@ func (restore *MongoRestore) Restore() Result {
 			}
 			return Result{Err: fmt.Errorf("mongorestore target '%v' invalid: %v", restore.TargetDirectory, err)}
 		}
+		preludeFileExists, err := restore.ReadPreludeMetadata(target)
+		if !preludeFileExists {
+			// don't error out here because older dump versions will not prelude.json
+			log.Logvf(log.DebugLow, "no prelude metadata found in target directory or parent, skipping")
+		} else if err != nil {
+			return Result{Err: fmt.Errorf("error reading dump metadata from prelude.json: %v", err)}
```
FGasper, line 407:
```suggestion
			return Result{Err: fmt.Errorf("error reading dump metadata from prelude.json: %w", err)}
```

mongorestore/mongorestore.go
```diff
@@ -396,6 +399,14 @@ func (restore *MongoRestore) Restore() Result {
 			}
 			return Result{Err: fmt.Errorf("mongorestore target '%v' invalid: %v", restore.TargetDirectory, err)}
 		}
+		preludeFileExists, err := restore.ReadPreludeMetadata(target)
+		if !preludeFileExists {
+			// don't error out here because older dump versions will not prelude.json
```
FGasper, line 404: Can you be more specific than “older”? What specific version added prelude.json?

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+	file, err := os.Open(filepath.Join(target.Path(), filename))
+	if errors.Is(err, os.ErrNotExist) {
+		// if prelude.json doesn't exist, check the parent directory in case db directory was used as target
+		file, err = os.Open(filepath.Join(target.Parent().Path(), filename))
+		if errors.Is(err, os.ErrNotExist) {
+			return false, nil
+		} else if err != nil {
+			return false, err
+		}
+	}
```
FGasper, line 698: Missing handling of non-nil err here.

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+		target = target.Parent()
+	}
+	file, err := os.Open(filepath.Join(target.Path(), filename))
+	if errors.Is(err, os.ErrNotExist) {
+		// if prelude.json doesn't exist, check the parent directory in case db directory was used as target
+		file, err = os.Open(filepath.Join(target.Parent().Path(), filename))
+		if errors.Is(err, os.ErrNotExist) {
+			return false, nil
+		} else if err != nil {
+			return false, err
```
FGasper, line 696:
```suggestion
		parentPath := filepath.Join(target.Parent().Path(), filename)
		file, err = os.Open(parentPath)
		if errors.Is(err, os.ErrNotExist) {
			return false, nil
		} else if err != nil {
			return false, fmt.Errorf("failed to read prelude metadata from file %#q: %w", parentPath, err)
```

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+
+	var prelude map[string]string
+	err = json.Unmarshal(bytes, &prelude)
+	if err != nil {
+		return true, fmt.Errorf("failed to unmarshal prelude metadata from %s: %w", filename, err)
+	}
+
+	dumpVersion, ok := prelude["ServerVersion"]
+	if !ok {
+		return true, fmt.Errorf("ServerVersion key not found in %s", filename)
```
FGasper, line 725:
```suggestion
		return true, fmt.Errorf("ServerVersion key not found in %#q", filename)
```

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+	}
+	bytes, err := io.ReadAll(reader)
+	if err != nil {
+		return true, fmt.Errorf("failed to read prelude metadata from %s: %w", filename, err)
+	}
+
+	var prelude map[string]string
+	err = json.Unmarshal(bytes, &prelude)
+	if err != nil {
+		return true, fmt.Errorf("failed to unmarshal prelude metadata from %s: %w", filename, err)
```
FGasper, line 720:
```suggestion
		return true, fmt.Errorf("failed to unmarshal prelude metadata from %#q: %w", filename, err)
```

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+			return true, fmt.Errorf("failed to open gzip file %s: %w", filename, err)
+		}
+		defer zipfile.Close()
+		reader = zipfile
+	} else {
+		reader = file
+	}
+	bytes, err := io.ReadAll(reader)
+	if err != nil {
+		return true, fmt.Errorf("failed to read prelude metadata from %s: %w", filename, err)
```
FGasper, line 714:
```suggestion
		return true, fmt.Errorf("failed to read prelude metadata from %#q: %w", filename, err)
```

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+		return true, fmt.Errorf("failed to unmarshal prelude metadata from %s: %w", filename, err)
+	}
+
+	dumpVersion, ok := prelude["ServerVersion"]
+	if !ok {
+		return true, fmt.Errorf("ServerVersion key not found in %s", filename)
+	}
+
+	// mongodump sets server version to unknown if it can't get the server version
+	if dumpVersion == "unknown" {
```
FGasper, line 729: This should probably be a constant rather than a literal string that gets copied around.

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+		log.Logvf(log.Info, "server version in prelude.json is 'unknown'")
+		return true, nil
+	}
+
+	restore.dumpServerVersion, err = db.StrToVersion(dumpVersion)
+	if err != nil {
+		return true, fmt.Errorf("failed to parse server version from prelude.json: %w", err)
```
FGasper, line 736:
```suggestion
		return true, fmt.Errorf("failed to parse server version from %#q: %w", filename, err)
```

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+	} else {
+		log.Logvf(log.Info, "successfully parsed prelude metadata from prelude.json")
```
FGasper, line 738:
```suggestion
		log.Logvf(log.Info, "successfully parsed prelude metadata from %#q", filename)
```

mongorestore/mongorestore.go
```diff
@@ -662,6 +673,74 @@ func (restore *MongoRestore) Restore() Result {
...
+		filename += ".gz"
+	}
+
+	var reader io.ReadCloser
+	if !target.IsDir() {
+		target = target.Parent()
+	}
+	file, err := os.Open(filepath.Join(target.Path(), filename))
+	if errors.Is(err, os.ErrNotExist) {
+		// if prelude.json doesn't exist, check the parent directory in case db directory was used as target
```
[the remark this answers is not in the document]
shyam12860: No. There are 2 options - one is the db directory being used as the target, and the other one is the actual `collection_name.bson` file. `target.IsDir()` is handling the latter.

In a situation where the customer's taken a full db dump but point restore to one of the database directories, prelude.json will be in the top level directory, not in the database directory.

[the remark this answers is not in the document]
shyam12860: Let me add to this comment to make that clearer

mongorestore/mongorestore.go
shyam12860: I decided to use the full filepath while logging errors rather than just the filename, and that takes care of this change too.

mongodump/mongodump.go
```diff
@@ -946,6 +957,61 @@ func (dump *MongoDump) DumpMetadata() error {
...
+	if dump.OutputOptions.Out == "" {
+		filename = filepath.Join("dump", filename)
+	} else {
+		filename = filepath.Join(dump.OutputOptions.Out, filename)
+	}
+	if dump.OutputOptions.Gzip {
+		filename += ".gz"
+	}
+	if _, err := os.Stat(filepath.Dir(filename)); err != nil {
+		// most likely means there was no data to dump
```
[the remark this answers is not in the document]
shyam12860: This is already in a `if errors.Is(err, os.ErrExist) {`, should I change the wording on the comment instead?

mongorestore/mongorestore.go
FGasper: It’s probably better not to name the file specifically since it can optionally be `.json.gz` … ?

FGasper (changes requested):

mongodump/mongodump.go
```diff
@@ -365,16 +367,17 @@ func (dump *MongoDump) Dump() (err error) {
 		return fmt.Errorf("error dumping metadata: %v", err)
 	}
 
+	dump.serverVersion, err = dump.SessionProvider.ServerVersion()
+	if err != nil {
+		log.Logvf(log.Always, "warning, couldn't get version information from server: %v", err)
+		dump.serverVersion = "unknown"
```
FGasper, line 373: Also, can this be a constant?

FGasper, line 373:
```suggestion
		dump.serverVersion = serverVersionUnknown
```

mongorestore/mongorestore.go
shyam12860: It's unclear which line the target of this comment is, are you talking about line 702? The non nil error is handled there

shyam12860: Oh the outer `err`, I see

common/const.go
```diff
@@ -0,0 +1,5 @@
+package common
+
+const (
+	ServerVersionUnknown = "unknown"
```
shyam12860, line 4: This is used in mongodump and mongorestore. I couldn't find any place to put the common constants, so I made this, let me know if there's a better option

FGasper (approved):
LGTM. Thank you!

PR #778, opened 2025-03-25
TOOLS-3793: Update package 'net' to v0.36.0

shyam12860:
Ran the following commands:
go mod edit -require=golang.org/x/net@v0.36.0
go mod tidy -v
go mod vendor -v

Looks like net 0.36 requires go version 1.23.0+, so that has been updated. Updating the go version requires an update to set_goenv.sh, and the problem was not obvious in the evergreen task failure, so I'm also adding an echo statement to that script.


FGasper (approved):
LGTM. Thanks!

PR #782, opened 2025-03-31
common.yml: expansions.update correctly for kondukto_token

shyam12860:
Turns out empty expansions.update doesn't update the variable, my testing was wrong.


FGasper (approved):
LGTM. Thanks!

PR #783, opened 2025-03-31
set MinLinuxServerVersion for RHEL 9 to 6.0.0

shyam12860:
6.0.0 is the earliest supported version for RHEL 9, so we should not push new mongo-tools releases to the repos for earlier server versions.


release/platform/platform.go
```diff
@@ -522,36 +522,40 @@ var platforms = []Platform{
...
-		Pkg:       PkgRPM,
-		Repos:     []Repo{RepoOrg, RepoEnterprise},
-		BuildTags: defaultBuildTags,
+		Name:                  "rhel9",
+		Arch:                  ArchPpc64le,
+		OS:                    OSLinux,
+		Pkg:                   PkgRPM,
+		Repos:                 []Repo{RepoOrg, RepoEnterprise},
+		BuildTags:             defaultBuildTags,
+		MinLinuxServerVersion: &version.Version{Major: 6, Minor: 0, Patch: 0},
```
FGasper, line 531: nit: You can omit Minor & Patch here.

I know gopls et al. optionally insert them; I personally find it a bit cleaner to omit them since there’s less code to grok. Totally up to you.

FGasper (approved):
LGTM % one optional nit.

Thanks!

release/platform/platform.go
shyam12860: I don't mind either way, but for now I'm keeping this as is because the rest of the file is not omitting minor and patch

PR #786, opened 2025-03-31
match rhel9's name in etc/repo-config.yml  to platform.go

shyam12860:
Also update augmented sbom

### DIFF
--- a/etc/repo-config.yml
+++ b/etc/repo-config.yml
@@ -120,7 +120,7 @@ repos:
       - yum/redhat/8/mongodb-org
       - yum/redhat/8Server/mongodb-org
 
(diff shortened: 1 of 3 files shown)


shyam12860:
I merged this branch with `master` locally and then ran `EVG_TRIGGERED_BY_TAG=100.12.0 scripts/regenerate-and-diff-augmented-sbom.sh` and it passed. Might've been changing for me earlier because `KONDUKTO_BRANCH` is inferred in that script. This should be the augmented sbom that we merge to master now for 100.12.0,  it has the right vulnerability

FGasper (approved):
LGTM. Thanks!

PR #790, opened 2025-04-04
TOOLS-3799: Use // operator correctly to set empty when generating sbom

shyam12860:
The left side of the alternative operator (//) in jq will not be empty in this case because of the string being appended. This causes "pkg:golang/github.com/mongodb/mongo-tools@" to get included in the list of dependencies for mongo-tools, without the version after "@", even though the `go list -json` output for mongo-tools does not contain a .Module.Version field for mongo-tools.

Compared against the old file this script generates to sends to silkbomb and the only thing removed was
"pkg:golang/github.com/mongodb/mongo-tools@"

Including the new sbom and augmented sbom in this commit.

### DIFF
--- a/cyclonedx.sbom.json
+++ b/cyclonedx.sbom.json
@@ -586,19 +586,6 @@
       "version": "v1.0.1"
     },
     {
-      "bom-ref": "pkg:golang/github.com/mongodb/mongo-tools",
-      "externalReferences": [
-        {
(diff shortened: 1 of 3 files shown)


wratner (approved):
LGTM

FGasper (approved):
LGTM, but we should deduplicate this script with Mongosync if possible.

PR #792, opened 2025-04-14
mongo-tools evergreen for mongodump passthrough tests

rcownie:
1. Evergreen tasks and functions are ported from mongosync/evergreen into mongo-tools/mongodump_passthrough
2. Various changes are needed to match the different directory structure and evergreen environment
3. Slight changes to mongodump/mongorestore to help with debugging passthrough failures
4. New buildvariant "rhel80" used for mongodump passthrough testing only

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -697,7 +697,6 @@ post:
         exit 0
   # Extra post steps needed for mongodump passthrough.
   - func: f_resmoke_report_attach
-  - func: f_gotest_report
   - func: "attach artifacts"
   - func: f_mongo_coredumps_save
 
@@ -719,8 +718,6 @@ timeout:
         fi
   # Extra timeout steps needed for mongodump passthrough.
   - func: f_expansions_write
-  # Hang analyzer is used to upload data files after a timed-out resmoke test.
-  - func: "run hang analyzer"
   - func: "wait for resmoke to shutdown"
 
 tasks:
--- a/common.yml
+++ b/common.yml
@@ -1,3 +1,13 @@
+# Support for mongodump+mongorestore passthrough tests.
+
+include:
+  - filename: "mongodump_passthrough/globals.yml"
+  - filename: "mongodump_passthrough/parameters.yml"
+  - filename: "mongodump_passthrough/modules.yml"
+  - filename: "mongodump_passthrough/functions.yml"
+  - filename: "mongodump_passthrough/tasks.yml"
+  - filename: "mongodump_passthrough/variants.yml"
+
 #######################################
 #    Tools Driver Config for MCI      #
 #######################################
@@ -649,6 +659,18 @@ pre:
   - command: expansions.update
     params:
       file: expansions.yml
+  # Extra pre steps needed for mongodump passthrough.
+  - command: expansions.update
+    params:
+      updates:
+        - key: python
+          value: "/opt/mongodbtoolchain/v4/bin/python3"
+        - key: resmoke_dir
+          value: "src/resmoke"
+        - key: resmoke_venv_dir
+          value: "resmoke-venv"
+        - key: mongosync_binary_folder
+          value: "mongosync-coverage-binary"
 
 post:
   # attach.results works for jstests, gotest.parse_files for golang tests
@@ -673,6 +695,11 @@ post:
           set -v
         rm -rf /data/db/*
         exit 0
+  # Extra post steps needed for mongodump passthrough.
+  - func: f_resmoke_report_attach
+  - func: f_gotest_report
+  - func: "attach artifacts"
+  - func: f_mongo_coredumps_save
 
 timeout:
   - command: shell.exec
@@ -690,6 +717,11 @@ timeout:
           # git the processes a second or two to dump their stacks
           sleep 10
         fi
+  # Extra timeout steps needed for mongodump passthrough.
+  - func: f_expansions_write
+  # Hang analyzer is used to upload data files after a timed-out resmoke test.
+  - func: "run hang analyzer"
+  - func: "wait for resmoke to shutdown"
 
 tasks:
   - name: dist
--- a/mongodump/oplog_dump.go
+++ b/mongodump/oplog_dump.go
@@ -158,7 +158,7 @@ func oplogDocumentValidator(in []byte) error {
 	}
 
 	if ok && nsStr == "admin.system.version" {
-		return fmt.Errorf("cannot dump with oplog if admin.system.version is modified")
+		return fmt.Errorf("cannot dump with oplog if admin.system.version is modified by %v", raw)
 	}
 
 	if _, err := raw.LookupErr("o", "renameCollection"); err == nil {
--- a/mongodump_passthrough/README.md
+++ b/mongodump_passthrough/README.md
@@ -1,0 +1,9 @@
+mongo-tools/mongodump_passthrough contains evergreen .yml files to support resmoke passthrough
+testing of mongodump+mongorestore.
+
+This started out as a copy of the files from mongosync/evergreen, but has minor changes to cope with
+running from mongo-tools, and to avoid conflicts with the mongo-tools/common.yml
+
+These files should be included from mongo-tools/common.yml
+
+We have to be careful to avoid clashes with any task of function names used in common.yml
--- a/mongodump_passthrough/empty.go
+++ b/mongodump_passthrough/empty.go
@@ -1,0 +1,3 @@
+package empty_package_to_satisfy_go_vet
+
+// Empty package so that "go vet" will be happy with this directory.
--- a/mongodump_passthrough/functions.yml
+++ b/mongodump_passthrough/functions.yml
@@ -1,0 +1,773 @@
+# Copied from mongosync/evergreen, with slight changes.
+
+# Functions and variables which differ from mongosync/evergreen/functions.yml are
+# marked as
+#   either: "Added for mongodump_passthrough"
+#   or:     "Changed for mongodump_passthrough"
+#
+# The tests are run by loading the mongosync repo as an evergreen module under src/mongosync.
+# The sources for mongodump-suite-gen, mongodump-task-gen, js tests (with slight modifications
+# for mongodump), and the dump_restore.py are all in the mongosync repo.
+#
+# Changes made to these in the mongosync repo will only be picked up when the _mongosync_pin
+# is changed here to point at the new mongosync commit hash.
+
+variables:
+  _migration_verifier_pin: &_migration_verifier_pin 9348607d61c4c865287828ce78b798a0f1b1cb89
+  _resmoke_dir: &_resmoke_dir "src/resmoke"
+  _src_dir: &src_dir src/mongosync
+  # Added for mongodump_passthrough
+  _mongosync_pin: &_mongosync_pin 6b18a97da5aa5a3ede27a2376a1d5c20dcaf986e
+
+  f_expansions_write: &f_expansions_write
+    command: expansions.write
+    params:
+      file: expansions.yml
+      redacted: true
+
+  # Changed for mongodump_passthrough
+  f_repo_fetch: &f_repo_fetch
+    command: git.get_project
+    params:
+      directory: src/github.com/mongodb/mongo-tools
+      revisions:
+        mongosync: *_mongosync_pin
+
+  # Added for mongodump_passthrough
+  f_link_to_mongo_tools: &f_link_to_mongo_tools
+    command: shell.exec
+    params:
+      shell: bash
+      script: |
+        ln -s ${workdir}/src/github.com/mongodb/mongo-tools src
+
+  "set up logger credentials": &f_set_up_logger_credentials
+    command: subprocess.exec
+    params:
+      binary: "./src/mongosync/evergreen/logger_credentials_setup.sh"
+      add_expansions_to_env: true
+
+  f_resmoke_wheelhouse_setup: &f_resmoke_wheelhouse_setup
+    command: subprocess.exec
+    params:
+      binary: "./src/mongosync/evergreen/resmoke_wheelhouse_setup.sh"
+      add_expansions_to_env: true
+
+  # Function used to download particular bits of resmoke and binaries uploaded previously.
+  f_resmoke_with_binaries_fetch: &f_resmoke_with_binaries_fetch
+    command: s3.get
+    params:
+      aws_key: ${aws_key}
+      aws_secret: ${aws_secret}
+      # Changed for mongodump_passthrough
+      remote_file: mongo-tools/mongodump_passthrough/${build_variant}/${revision}/resmoke-with-binaries/${build_id}.tgz
+      bucket: mciuploads
+      extract_to: *_resmoke_dir
+
+  f_resmoke_wheelhouse_fetch: &f_resmoke_wheelhouse_fetch
+    command: s3.get
+    params:
+      aws_key: ${aws_key}
+      aws_secret: ${aws_secret}
+      # Changed for mongodump_passthrough
+      remote_file: mongo-tools/mongodump_passthrough/${build_variant}/${revision}/resmoke-python-wheelhouse/${build_id}.tgz
+      bucket: mciuploads
+      extract_to: "."
+
+  f_migration_verifier_binary_fetch: &f_migration_verifier_binary_fetch
+    command: s3.get
+    params:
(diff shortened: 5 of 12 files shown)


rcownie:
The check-sbom-lite task is failing apparently because of a disagreement between "go1.23.7" and "go1.23.8".  I don't know where these versions are coming from, so I'd welcome any suggestions for fixing this.  But it seems unrelated to the new functionality.

mongodump_passthrough/README.md
```diff
@@ -0,0 +1,9 @@
+mongo-tools/mongodump_passthrough contains evergreen .yml files to support resmoke passthrough
```
shyam12860, line 1: Considering this is for both mongodump and mongorestore, would it be better to call this directory just `passthrough` rather than `mongodump_passthrough`?

mongodump_passthrough/functions.yml
```diff
@@ -0,0 +1,773 @@
+# Copied from mongosync/evergreen, with slight changes.
+
+# Functions and variables which differ from mongosync/evergreen/functions.yml are
+# marked as
+#   either: "Added for mongodump_passthrough"
+#   or:     "Changed for mongodump_passthrough"
+#
+# The tests are run by loading the mongosync repo as an evergreen module under src/mongosync.
+# The sources for mongodump-suite-gen, mongodump-task-gen, js tests (with slight modifications
```
shyam12860, line 9: Could you tell me why the code for these lives in the mongosync directory? I don't have the context

common.yml
```diff
@@ -673,6 +695,11 @@ post:
           set -v
         rm -rf /data/db/*
         exit 0
+  # Extra post steps needed for mongodump passthrough.
```
shyam12860, line 698: Considering all the commands added to pre / post / timeout are passthrough specific, is there a way to run them conditionally, only for passthrough?

Can running it for every task cause errors or slowdowns?

shyam12860 (approved):

mongodump_passthrough/README.md
rcownie: Tricky one.  In the scope document it was called "mongodump/mongorestore passthrough, in the technical design I shortened that to "mongodump passthrough".  I'm inclined to stick with that because the mongo-tools repo contains other tools, e.g. mongoimport/mongoexport.

mongodump_passthrough/functions.yml
rcownie: Sure.  The mongosync repo has a fairly elaborate multi-language stack of JS code for generating traffic on the source db and verifying consistency between src and dst clusters; python code for "fixtures" which set up the various components of the test infrastructure - the two clusters, the driver for cluster-to-cluster-replication, and the particular method of replication (mongosync, multi-mongosync, and now mongodump+mongorestore); the suite.yml files which configure the fixtures to run a suite of tests; the suite generator which generates the suites; and the task-generator which tells evergreen about the suites.

And that code has to evolve to cope with changes in the server resmoke infrastructure.

So it's good to share as much of that code as possible, andonly have one copy to be maintained.

It also makes it easier to eyeball the code and see which parts (not very many) have different behavior for mongodump/mongorestore vs mongosync.

So the design choice was to leave that code in the mongosync repo and make minimal change to it, but
fetch it from the mongo-tools repo as an evergreen module.

common.yml
shyam12860: Fair enough. I don't see a way to use tags to do this either because you need them to run in a specific order. 

I'm okay with leaving them here if they're quick enough and don't cause failures.

mongodump_passthrough/README.md
shyam12860: Reasonable. I was going to suggest putting them in mongodump/passthrough instead, but these tests are for dump + restore both, so this is fine in that case.

common.yml
autarch: Yeah, there's no way to do this. It's a bit annoying.

common.yml
```diff
@@ -673,6 +695,11 @@ post:
...
+  - func: f_resmoke_report_attach
+  - func: f_gotest_report
```
autarch, line 700: Is this necessary for the resmoke stuff? Did you just copy this from the mongosync config?

common.yml
```diff
@@ -690,6 +717,11 @@ timeout:
           # git the processes a second or two to dump their stacks
           sleep 10
         fi
+  # Extra timeout steps needed for mongodump passthrough.
+  - func: f_expansions_write
+  # Hang analyzer is used to upload data files after a timed-out resmoke test.
+  - func: "run hang analyzer"
```
autarch, line 723: I'm not sure we ever actually run this for Mongosync, because for some reason we _can't_. So I'm not sure if this is useful here either.

mongodump/oplog_dump.go
```diff
@@ -158,7 +158,7 @@ func oplogDocumentValidator(in []byte) error {
 	}
 
 	if ok && nsStr == "admin.system.version" {
-		return fmt.Errorf("cannot dump with oplog if admin.system.version is modified")
+		return fmt.Errorf("cannot dump with oplog if admin.system.version is modified by %v", raw)
```
autarch, line 161: Doesn't printing a `bson.Raw` just print a `[]byte` slice in hex? I think that might be confusing for any users who see this error.

mongodump_passthrough/empty.go
autarch, line 1: You can just add an `exclude = "mongodump_passthrough/**" to the golangci-lint config in this repo's `precious.toml` file.

mongodump_passthrough/functions.yml
```diff
@@ -0,0 +1,773 @@
...
+
+  f_mongosync_binary_fetch: &f_mongosync_binary_fetch
+    command: s3.get
+    params:
+      aws_key: ${aws_key}
+      aws_secret: ${aws_secret}
+      # Changed for mongodump_passthrough
+      remote_file: mongo-tools/mongodump_passthrough/${mongosync_compile_build_variant}/${revision}/${version_id}/${mongosync_binary_folder}/${version_id}.tgz
+      bucket: mciuploads
+      extract_to: "src/mongosync"
```
autarch, line 108: Do you need to do this at all? Do you actually run `mongosync` for tools passthroughs?

mongodump_passthrough/functions.yml
```diff
@@ -0,0 +1,773 @@
...
+  # into src/mongosync/dist.
+  "build mongodump and mongorestore":
+    - command: shell.exec
+      params:
+        working_dir: src/github.com/mongodb/mongo-tools
+        script: |
+          ${PREPARE_SHELL}
+          PATH=$PATH:$HOME
+          echo $PATH 
+          MISE_INSTALL_PATH='${workdir}/.local/bin/mise' sh ${workdir}/src/mongosync/etc/mise.run.sh
```
autarch, line 254:
```suggestion
          MISE_INSTALL_PATH='${workdir}/.local/bin/mise' sh ${workdir}/src/mongosync/etc/mise.run.sh
```

Michael recently made a change to add retrying when this fails. It'd be good to get that into this as well.

mongodump_passthrough/functions.yml
```diff
@@ -0,0 +1,773 @@
...
+      script: |
+        set -o errexit
+        set -o pipefail
+        set -o xtrace
+
+        rm -rf `find suite-config -name "suites"`
+        cd mongodump-suite-config
+        tar cvf ../suite-config/a.tar .
+        cd ../suite-config
+        tar xvf a.tar
```
autarch, line 308: Why use `tar` like this instead of `cp -r`?

mongodump_passthrough/functions.yml
```diff
@@ -0,0 +1,773 @@
...
+          ${PREPARE_SHELL}
+
+          mise install
+          mise exec -- npm install
+
+  # Added for mongodump_passthrough
+  #
+  # This build the mongo-tools executables, then copies mongogump and mongorestore
+  # into src/mongosync/dist.
+  "build mongodump and mongorestore":
```
autarch, line 246: I don't understand why you're using `mise` to build the tools binaries. The mongo-tools repo is not using mise (yet?). So when we do an actual release of the tools, we'll just use whatever `go` is installed on the Evergreen host. I think we should use that same `go` for these tests.

mongodump_passthrough/functions.yml
```diff
@@ -0,0 +1,773 @@
...
+        display_name: "migration_verifier release artifact (compiled)"
+
+  f_migration_verifier_binary_fetch:
+    - *f_migration_verifier_binary_fetch
+
+  f_make_migration_verifier_binary_executable:
+    - *f_make_migration_verifier_binary_executable
+
+  # This uploads everything in src/mongosync/dist/
+  # For mongodump, we don't need the mongosync bi9naries, but we have
```
autarch, line 424:
```suggestion
  # For mongodump, we don't need the mongosync binaries, but we have
```

autarch (changes requested):
This looks good % some questions I had about specific changes. I also have a few general comments:

* It'd be nice to separate out the Go version changes into a separate PR, since it's totally unrelated to this work.
* It seems like you copied a lot of tasks & functions that aren't actually used. It'd be good to remove those. Maybe just leave a comment in where you deleted it if you're concerned about keeping up to date with future changes in the mongosync config.
*

common.yml
rcownie: It's copied from the mongosync config without any deeper analysis.

common.yml
rcownie: I'm slightly inclined to match what mongosync evergreen has as closely as possible.  But if you're 100% sure it's useless, I could take it out.

mongodump/oplog_dump.go
rcownie: It comes out as text, somehow.  I've looked at the logs. bson.Raw has a String() method so that will get it formatted.

"5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any)."

mongodump_passthrough/empty.go
rcownie: The mongo-tools/common.yml has a "vet" task which doesn't go through precious.

  - name: vet
    commands:
      - func: "fetch source"
      - command: shell.exec
        type: test
        params:
          working_dir: src/github.com/mongodb/mongo-tools
          script: |
            ${_set_shell_env}
            set -x
            set -v
            set -e
            go vet -composites=false ./bsondump ./mongo*

mongodump_passthrough/functions.yml
rcownie: I don't build or run any mongosync binaries.  But I copy the mongodump and mongorestore binaries into the mongosync/dist directory, they get packed up and sent to s3 from there, and then they get downloaded back into mongosync/dist.

So we have the same task names and same dependencies as for mongosync passthrough tests, we're just gluing mongodump and mongorestore into that same framework.

rcownie:
> This looks good % some questions I had about specific changes. I also have a few general comments:
> 
> * It'd be nice to separate out the Go version changes into a separate PR, since it's totally unrelated to this work.

Done

> * It seems like you copied a lot of tasks & functions that aren't actually used. It'd be good to remove those. Maybe just leave a comment in where you deleted it if you're concerned about keeping up to date with future changes in the mongosync config.

I pruned it fairly heavily: the tasks+functions here are 892 lines, compared to 1826 lines in mongosync. Might have missed one or two unused functions, IDK, but "evergreen validate common.yml" seems mostly happy.

mongodump_passthrough/functions.yml
rcownie: I'm more confident that I understand the behavior of tar.

"Historic versions of the cp utility had a -r option.  This implementation supports that option, however,
  its behavior is different from historical FreeBSD behavior.  Use of this option is strongly discouraged as
  the behavior is implementation-dependent."

autarch: We only run this stuff on Linux, which has a `cp -r` that works as you'd expect. Using `tar` seems unnecessarily baroque.

common.yml
autarch: Eh, it's fine either way, I think.

mongodump/oplog_dump.go
autarch: Ah, interesting. I think in that case you should use `%s` instead.

mongodump_passthrough/empty.go
autarch: Ah, right. I forgot about that.

autarch: I guess you could just move that to precious too. It'd be fairly trivial.

mongodump_passthrough/functions.yml
autarch: I think a comment explaining this would be good.

autarch:
> > * It seems like you copied a lot of tasks & functions that aren't actually used. It'd be good to remove those. Maybe just leave a comment in where you deleted it if you're concerned about keeping up to date with future changes in the mongosync config.
> 
> I pruned it fairly heavily: the tasks+functions here are 892 lines, compared to 1826 lines in mongosync. Might have missed one or two unused functions, IDK, but "evergreen validate common.yml" seems mostly happy.

I'm pretty sure validate won't complain about unused functions and tasks. I think it'd be good to audit for those and not include them. Otherwise, it's a burden on whoever has to maintain this in the future. They're forced to try figure out why some seemingly unused function is included.

common.yml
rcownie: I removed it.

mongodump_passthrough/functions.yml
rcownie: Added a comment

mongodump_passthrough/functions.yml
rcownie: Done

mongodump_passthrough/functions.yml
rcownie: Baroque is my era.  I'm easily confused by the way cp puts things inside a destination directory if it already exists.

mongodump_passthrough/functions.yml
rcownie: Changed so that we use the same setup as mongo-tools/common.yml "run make target" for code that runs in mongo-tools, but use the mise for code that lives in mongosync repo (the mongodump-suite-gen).

I did try to get everything to run without mise, but the shell setup (sourcing ./set_goenv.sh) gets confusing so I backed off to this.

mongodump_passthrough/functions.yml
rcownie: Done

rcownie:
> > > * It seems like you copied a lot of tasks & functions that aren't actually used. It'd be good to remove those. Maybe just leave a comment in where you deleted it if you're concerned about keeping up to date with future changes in the mongosync config.
> > 
> > 
> > I pruned it fairly heavily: the tasks+functions here are 892 lines, compared to 1826 lines in mongosync. Might have missed one or two unused functions, IDK, but "evergreen validate common.yml" seems mostly happy.
> 
> I'm pretty sure validate won't complain about unused functions and tasks. I think it'd be good to audit for those and not include them. Otherwise, it's a burden on whoever has to maintain this in the future. They're forced to try figure out why some seemingly unused function is included.

It does complain about seemingly-unused tasks (except it can't tell if dynamically-generated tasks will have them as dependencies).  I don't know if that extends to functions as well.

mongodump_passthrough/functions.yml
rcownie: mongosync codebase skews the other way

(python_env) rcownie@M-M3G94W7K61 internal % fgrep %v `cat files2.txt` | wc -l
     171
(python_env) rcownie@M-M3G94W7K61 internal % fgrep %s `cat files2.txt` | wc -l
     621

mongodump_passthrough/functions.yml
autarch: I think this should be changed. This is a very idiosyncratic thing that we don't do elsewhere, AFAIK. It's time to move onto the Classical era, if not Romantic.

autarch (approved):
LGTM % please do change the `tar` thing. I think it is likely to confuse basically most future developers, who have not ever seen this idiom. Keep in mind that nowadays new devs often have much less Unix CLI experience than people from our era.

autarch:
@rcownie I think you missed my comment asking you to change the `tar` thing before you merged this. It's a common practice on our team to do an "LGTM but before you merge please do this one thing" approval.

PR #794, opened 2025-04-17
upgrade toolchain from go1.23.7 to go1.23.8

rcownie:
Toolchain version change needed to pass static tests on rhel80.

Splitting this off from the mongo-tools evergreen changes.


autarch (approved):
LGTM!

PR #795, opened 2025-04-22
TOOLS-3823 enable cron for nightly tests

rcownie:
Add "cron: 0 3 * * *" to enable nightly testing (for mongodump passthrough)

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -2058,6 +2058,7 @@ buildvariants:
 
   - name: rhel88
     display_name: "# RHEL 8.8"
+    cron: 0 3 * * *
     run_on:
       - rhel8.8-large
     expansions:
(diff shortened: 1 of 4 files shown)


ege-mgdb (approved):
lgtm

common.yml
```diff
@@ -2058,6 +2058,7 @@ buildvariants:
 
   - name: rhel88
     display_name: "# RHEL 8.8"
+    cron: 0 3 * * *
```
ajayvijayakumar123, line 2061: Why are we adding a cron job for this variant? It looks like we only use rhel80 (not rhel88) for passthru tests?

ajayvijayakumar123 (changes requested):

common.yml
rcownie: I just figured that if we're adding nightly testing (that we've apparently never had before for mongo-tools ?) we might as well enable the other tests at the same time and get them on the waterfall to see how they look.

Of course it's a one-line change to disable that if it turns out to be unhelpful.

ajayvijayakumar123 (approved):
LGTM!

PR #798, opened 2025-04-30
Add an integration test for dump piped directly to restore.

FGasper:
This requires a fix for a race condition around MongoDump’s internal `storageEngine`.


mongodump/mongodump.go
```diff
@@ -624,29 +626,36 @@ func (dump *MongoDump) DumpIntent(intent *intents.Intent, buffer resettableOutpu
...
-		if err != nil {
-			log.Logvf(log.Always,
-				"failed to determine storage engine, an mmapv1 storage engine could result in"+
-					" inconsistent dump results, error was: %v", err)
-		} else if isMMAPV1 {
-			dump.storageEngine = storageEngineMMAPV1
+
+	var storageEngine storageEngineType
+
+	dump.storageEngine.Store(func(seng storageEngineType) storageEngineType {
```
shyam12860, line 632: "The storage engine cannot change from namespace to namespace, so we set it the first time we reach here"

Couple of questions:
Can we not use serverStatus() or something else to determine the storage engine?

If the storage engine doesn't change, can we do this outside of `DumpIntent`, and call it just once, to avoid the race condition? Only extra thing we'd need to do is pick a random collection to pass to `IsMMAPV1`

shyam12860 (approved):

shyam12860 (commented):
Hit approve by mistake

mongodump/mongodump.go
```diff
@@ -150,7 +150,9 @@ func (dump *MongoDump) Init() error {
 
 	// this would be default, but explicit setting protects us from any
 	// redefinition of the constants.
-	dump.storageEngine = storageEngineUnknown
+	dump.storageEngine.Store(func(set storageEngineType) storageEngineType {
```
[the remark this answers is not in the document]
FGasper: If we end up using it more widely, but since this is lifted from mongosync I feel like it might as well stay as-is. This kind of usage has been fairly rare thus far.

mongodump/mongodump.go
FGasper: Actually … heh.

MMAPv1 was removed in v4.2, which is the oldest version of MongoDB that the tools still support.

shyam12860:
Could we mention in the commit message before merging this that the storageEngine code was causing the race condition, and rather than fix the race condition we can just get rid of MMAPv1 and avoid the problem?

shyam12860 (approved):

PR #801, opened 2025-05-16
TOOLS-3819: update golang/x/net to address vulnerability

shyam12860:
Changes generated are automatic with `go mod edit -require=golang.org/x/net@v0.40.0` and then go mod vendor + tidy


autarch (approved):
LGTM!

PR #804, opened 2025-05-21
100.12.1 augmented sbom with branch_name=master

shyam12860:


autarch (changes requested):
This doesn't look right to me. First, every vuln has `"state": "in_triage"`. That can't be correct.

Second, why is there an _NPM_ package showing up here?!

PR #805, opened 2025-05-21
TOOLS-3889: changelog update for 100.12.1

shyam12860:


CHANGELOG.md
```diff
@@ -1,5 +1,40 @@
...
+
+<ul>
+<li>[<a href='https://jira.mongodb.org/browse/TOOLS-3889'>TOOLS-3889</a>] -         Release Database Tools 100.12.1
+</li>
+</ul>
+
+### Task
+
+<ul>
+<li>[<a href='https://jira.mongodb.org/browse/TOOLS-3682'>TOOLS-3682</a>] -         Move macOS support from macOS 11 to macOS 14
```
FGasper, line 34: Do we still support macOS 11? This sounds like we’ve dropped support for the older version, but the line about “adds support for MacOS 14” sounds like we now support both.

(As a user I’d also wonder: what about macOS 12 & 13? Are those supported?)

FGasper (changes requested):

dsahni (approved):

FGasper (approved):
LGTM. Thanks!

PR #806, opened 2025-05-30
TOOLS-3822 Fix mongorestore’s handling of newlines in namespaces

FGasper:
mongorestore’s namespace matching uses regular expressions, which by default consider newlines _not_ to match the `.` wildcard. As a result, mongorestore has been quietly excluding various newline-containing namespaces from restores.

This changeset fixes that by setting the regular expression flag that makes regexp wildcards match newlines.

The CHANGELOG.md is also tidied here.

### DIFF
--- a/mongorestore/ns/ns.go
+++ b/mongorestore/ns/ns.go
@@ -38,8 +38,8 @@ var (
 
 // Escape escapes instances of '\' and '*' with a backslash.
 func Escape(in string) string {
-	in = strings.Replace(in, `\`, `\\`, -1)
-	in = strings.Replace(in, "*", `\*`, -1)
+	in = strings.ReplaceAll(in, `\`, `\\`)
+	in = strings.ReplaceAll(in, "*", `\*`)
 	return in
 }
 


shyam12860 (approved):

zhenxuanjameszhang (commented):
The change LGTM.

However, I do not think the previous implementation is a bug. Based on my quick search around, regular expression does not consider new line as matches. 

Out of caution, I tested additional regular expression like `(?s)(?s)^(|.[^\])*(.)$` and it seems applying it twice would not hurt. 

However, if we do this change, can the user still exclude namespace if there are new line character or we lose that feature? It seems to me that we would not have that capability anymore.

We can choose to change the default behavior to turn on dotall flag to make this easier for the user. I think this would be helpful but I would include the document change on this to explicitly say so.

FGasper:
As replied in DM: users don’t specify the regexp directly; mongorestore uses regexps to implement [wildcard matching](https://www.mongodb.com/docs/database-tools/mongorestore/#std-option-mongorestore.--nsExclude).

PR #810, opened 2025-06-11
TOOLS-3903 Always include all deps for all platforms in SBOM and only include deps for released binaries

autarch:
Previously, the SBOM would only contain dependencies that were used in the platform on which we generated the SBOM Lite file. We didn't notice this because we did not have any platform-conditional dependencies, but Jane was working on a PR that did add such a dep.

This changes the code that generates the SBOM Lite to get a list of deps for each of our supported platforms.

This _also_ changes the SBOM generation to only look at dependencies of the code for each of the release binaries. Previously, we included all the deps for everything in the repo. There were a huge number of packages that were only used by dev tooling, but not in the released tools.

### DIFF
--- a/cyclonedx.sbom.json
+++ b/cyclonedx.sbom.json
@@ -101,106 +101,6 @@
       "version": "v1.3.2"
     },
     {
-      "bom-ref": "pkg:golang/github.com/aws/aws-sdk-go@v1.53.11",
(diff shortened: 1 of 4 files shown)


mmcclimon (approved):
LGTM, thanks!

PR #816, opened 2025-07-15
Update to golangci-lint v2

FGasper:
This includes code updates to satisfy the new (more stringent) linter rules.

### DIFF
--- a/mongoexport/csv.go
+++ b/mongoexport/csv.go
@@ -61,7 +61,7 @@ func (csvExporter *CSVExportOutput) WriteHeader() error {
 }
 
 // WriteFooter is a no-op for CSV export formats.
-func (_ *CSVExportOutput) WriteFooter() error {
+func (*CSVExportOutput) WriteFooter() error {
 	// no CSV footer
 	return nil
 }
@@ -110,7 +110,7 @@ func (csvExporter *CSVExportOutput) ExportDocument(document bson.D) error {
 // It will also handle dot-delimited field names for nested arrays or documents.
 func extractFieldByName(fieldName string, document interface{}) interface{} {
 	dotParts := strings.Split(fieldName, ".")
-	var subdoc interface{} = document
+	var subdoc = document
 
 	for _, path := range dotParts {
 		docValue := reflect.ValueOf(subdoc)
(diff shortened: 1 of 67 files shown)


mongoexport/csv.go
```diff
@@ -119,13 +119,14 @@ func extractFieldByName(fieldName string, document interface{}) interface{} {
 		}
 		docType := docValue.Type()
 		docKind := docType.Kind()
-		if docKind == reflect.Map {
+		switch docKind {
```
zhenxuanjameszhang, line 122: Does the linter require switch statement instead of if else block? I feel both style should be valid.

mongoexport/mongoexport_test.go
```diff
@@ -190,7 +190,7 @@ func TestMongoExportTOOLS1952(t *testing.T) {
 	err = sessionProvider.Run(bson.D{{"drop", collName}}, &r1, dbName)
 	if err != nil {
 		var commandErr mongo.CommandError
-		if !(errors.As(err, &commandErr) && commandErr.Code == 26) {
+		if !errors.As(err, &commandErr) || commandErr.Code != 26 {
```
zhenxuanjameszhang, line 193: Does the linter require always use one style of boolean logic? I feel either way should be fine but would not mind enforcing this from now on.

mongoimport/mongoimport.go
```diff
@@ -617,8 +624,15 @@ func (imp *MongoImport) getInputReader(in io.Reader) (InputReader, error) {
...
+		return NewTSVInputReader(
+			colSpecs,
+			in,
+			out,
+			imp.IngestOptions.NumDecodingWorkers,
+			ignoreBlanks,
+			imp.InputOptions.UseArrayIndexFields,
+		), nil
 	}
 	return NewJSONInputReader(
```
zhenxuanjameszhang, line 637: nit since it is not related to linter change, I feel this should be in the else statement to be more readable.

zhenxuanjameszhang (approved):
LGTM. Thanks for doing this work to make our code base quality better!

I have a few questions, it would be great if you can answer them. 

Could you please also summarize the lint rule change in the PR DESCRIPTION? I got some idea by reviewing but think it would be nice to have it for reference.

PR #818, opened 2025-07-18
TOOLS-3927 Update Evergreen config to explicitly test with Server 8.2

autarch:
This required a few changes to account for new test failures that popped up when testing with 8.2.0.

The `autoIndexId` option for collection creation has been removed entirely in this release. When we try to restore a collection that includes this option, it will now be removed entirely when restoring to 8.2.0+ clusters.

This PR changes the tests that use this `autoIndexId` option to either not use it (when possible) or we skip those tests with 8.2.0+.

We also needed a new "load libraries" file for 8.2. I just copied the one from 8.1, but these are keyed by Server version, so we need one for each version.

There was also a weird issue with one of the JS tests with 8.2.0 because of updates to the Server shell's JS engine (newer Spidermonkey), so I rewrote it in Go.


mongoexport/mongoexport_test.go
```diff
@@ -253,3 +266,84 @@ func TestMongoExportTOOLS1952(t *testing.T) {
...
+			optionsFunc: func(o Options) Options {
+				o.Query = "{ hello }"
+				return o
+			},
+			errorTestFunc: func(t *testing.T, err error) {
+				require.Regexp(t, `query.+is not valid JSON`, err.Error())
+			},
+		},
+		{
+			name: "invalid sort",
```
shyam12860, line 311: The optionsFunc and errorTestFunc of this test case looks the same as the one above, should they be different?

autarch: It's not _exactly_ the same. I'm setting `o.Query` in one test and `o.Sort` in the other.

PR #820, opened 2025-07-22
TOOLS-3931: fix goroutine leak in `DumpIntents` function

ycycyyc:


shyam12860 (dismissed):

shyam12860:
There isn't a leak in the current code because the channel is tied to the number of parallel workers, but no harm in protecting against future mistakes.

shyam12860 (approved):

ycycyyc:
Thanks. I'm done signing MongoDB Contributor Agreement.

autarch:
Thanks for signing the CLA! I reverted the revert, so your change will be in the next release.

PR #822, opened 2025-07-24
TOOLS-3920 Fix macOS notarization

autarch:
This adds some code to the release process to verify that macOS releases are properly notarized. It also fixes the sign task so that it respects the fake tag so we can trigger actual notarization for testing.


FGasper (approved):
LGTM. Thanks!

PR #824, opened 2025-07-31
Upgrade to golangci-lint 2.3.0

autarch:
The older version we were using crashed with out of memory errors on my computer, which made working on the tools code very difficult. This includes the upgrade and all the fixes automatically applied by the latest version (via the `staticcheck` linter).


bsondump/bsondump_test.go
```diff
@@ -273,7 +272,7 @@ func TestBsondumpMaxBSONSize(t *testing.T) {
 	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
 
 	// 16mb + 16kb
-	maxSize := int(16*math.Pow(1024, 2)) + sixteenKB
+	maxSize := int(16*float64(1024*1024)) + sixteenKB
```
ege-mgdb: Why not just: 
```suggestion
	maxSize :=16*1024*1024 + sixteenKB
```

bsondump/bsondump.go
```diff
@@ -96,7 +95,7 @@ func New(opts Options) (*BSONDump, error) {
 
 	// 16kb + 16mb - This is the maximum size we would get when dumping the
 	// oplog itself. See https://jira.mongodb.org/browse/TOOLS-3001.
-	maxBSONSize := (16 * 1024) + (16 * math.Pow(1024, 2))
+	maxBSONSize := (16 * 1024) + (16 * float64(1024*1024))
```
ege-mgdb:
```suggestion
	maxBSONSize := (16 * 1024) + (16*1024*1024)
```

mongoimport/common.go
```diff
@@ -227,14 +227,12 @@ func getUpsertValue(field string, document bson.D) interface{} {
 		log.Logvf(log.DebugHigh, "no subdoc found for '%v'", left)
 		return nil
 	}
-	switch subDoc.(type) {
+	switch subDoc := subDoc.(type) {
 	case bson.D:
-		//nolint:errcheck
-		subDocD := subDoc.(bson.D)
+		subDocD := subDoc
```
ege-mgdb, line 232: Is this necessary?

autarch: Yes, note that those are two different variables, and `subDocD` will have the type `bson.D`, while `subDoc` is `any`.

PR #830, opened 2025-08-06
TOOLS-3941 - make mongodump_passthrough work with newer pinned mongosync

rcownie:
The devprod tiger team effort made various good changes to mongosync evergreen config and scripts.

This commit adjusts the mongo-tools/mongodump_passthrough scripts to work against a pinned newer mongosync:

  - shell scripts have been moved from src/mongosync/evergreen/foo.sh to src/mongosync/evergreen/scripts/foo.sh
  
  - mongodump_passthrough tests now run with build variant "mongodump_passthru_v", to isolate the specifics of the
     hardware platform from the name used in scripts and configuration of Build Baron context rules

  - "mongodump_passthru_v" is a cheaper amazon2-arm64 instance type, as used in mongosync to reduce costs.
  
  - mongodump_passthrough jstestfuzz scripts are currently disabled.  There's some issue with the new world that I
    don't have time to solve this week, and I'm on PTO next week, so this PR solves 3/4 of the problem and I'll file a
    follow-up ticket to fix jstestfuzz later.

### DIFF
--- a/mongodump_passthrough/functions.yml
+++ b/mongodump_passthrough/functions.yml
@@ -24,6 +24,9 @@ variables:
   # When you update this pin, you may need to make other changes to get mongodump_passthrough
   # working again.
   #
+  # And when you change the _mongosync_pin, you also need to update other pins
+  # which get passed as environment variables to scripts in mongosync/evergreen/scripts.
+  #
   _mongosync_pin: &_mongosync_pin "7126c03d3f87950824b393b5b1157065767c6d44"
 
   f_expansions_write: &f_expansions_write
--- a/mongodump_passthrough/functions.yml
+++ b/mongodump_passthrough/functions.yml
@@ -13,11 +13,18 @@
 # is changed here to point at the new mongosync commit hash.
 
 variables:
-  _migration_verifier_pin: &_migration_verifier_pin 9348607d61c4c865287828ce78b798a0f1b1cb89
+  _migration_verifier_pin: &_migration_verifier_pin "9431222bdc602e21b3cf2a64ee45d5eaceac6060"
   _resmoke_dir: &_resmoke_dir "src/resmoke"
   _src_dir: &src_dir src/mongosync
-  # Added for mongodump_passthrough
-  _mongosync_pin: &_mongosync_pin 55b77da9db675c8f92e288b684617ad8a6c158b3
+  #
+  # WARNING:
+  #
+  # There is a delicate relationship between the mongodump_passthrough evergreen
+  # configuration files, and the pinned mongosync which it fetches as an evergreen module.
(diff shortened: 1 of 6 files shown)


mongodump_passthrough/functions.yml
```diff
@@ -181,15 +188,27 @@ functions:
...
     - *f_link_to_mongo_tools
+    # Check that mongo-tools and mongosync both got cloned to the right place.
     - command: shell.exec
       params:
         shell: bash
         script: |
-          ls -l src
+          if [ ! -d "src/github.com/mongodb/mongo-tools" ]; then
+            echo "ERROR: git.get_project did not clone mongo-tools as required."
+            find . -name "mongo-tools"
```
mankawal, line 201: Were this `find` and the one at L204 dev-only scaffolding?
If not, can you please help me understand how their output is useful if these errors occur?

mongodump_passthrough/variants.yml
```diff
@@ -1,22 +1,28 @@
...
+# This maps to the same amazon2-arm64 platform used for mongosync passthrough testing.
 
 buildvariants:
-  - name: rhel80
+  # We use a strange name here so that the Build Baron context rules can steer
+  # the mongodump passthrough tests, and only those tests, to context "mongodump_passthrough".
+  #
+  # This name must also match some name in mongo-tools/release/platform/platform.go,
+  # so if you change the name here you also need to change that.
+  - name: mongodump_passthru_v
```
mankawal, line 11: Do we expect mongodump passthrough to ever run on more than one platforms?
If so, it might be useful to add the `run_on` variant to the name.

mankawal (commented):
LGTM + a few questions

mongodump_passthrough/variants.yml
rcownie: Not in the foreseeable future.

mongosync passthrough tests only run under two variants, the same platform but with and without race detection.
Running with race detector was out of scope for this project, and with the simpler thread interactions of mongodump/mongorestore it would probably cost money but not add a lot of value.  So I think one variant is ok.

mongodump_passthrough/functions.yml
rcownie: Painful experience with evergreen.  If you don't have your module declared in the buildvariant, then when git.get_project runs it just silently ignores the module, which then leads to confusing symptoms in later functions in the task.

If you're running something which needs that module, then it's better to have it clearly logged as an error in the evergreen task logs, and cause a task failure which will prevent execution of any dependent tasks.

The second possibility is that evergreen *does* fetch your module, but it doesn't put it where you were expecting.
And the "find" helps with that.

mankawal (approved):

PR #836, opened 2025-08-13
Remove code to validate notarization - this is just not working in CI

autarch:


FGasper (approved):
Instead of removing can we just log a warning?

Otherwise 👍

autarch:
> Instead of removing can we just log a warning?
> 
> Otherwise 👍

Log a warning about _what_? If the task doesn't fail, I doubt anyone will notice a warning saying that this didn't do anything.

FGasper:
I was thinking just to turn the Fail() function into a Logf(). But NBD either way.

PR #837, opened 2025-08-14
Fix S3 path when uploading release feed and don't run some tasks on git tag

autarch:
This also includes an update to the Augmented SBOM. This adds a bogus vulnerability, but this is because of a bug in Kondukto that cannot be fixed today.


mmcclimon (approved):
Fine by me, thanks!

(I don't really understand the Kondukto thing but I assume it's just housekeeping stuff and is all fine.)

PR #838, opened 2025-08-14
TOOLS-3929 Add release notes for 100.13.0

autarch:


CHANGELOG.md
```diff
@@ -1,5 +1,87 @@
 # Database Tools Changelog
 
+## 100.13.0
+
+_Released 2025-08-14_
+
+We are pleased to announce version 100.13.0 of the MongoDB Database Tools.
+
+The highlight for this release is Server 8.2.0 support. Starting with 8.2.0, when restoring
```
mankawal, line 9: (not a blocker)
I'm wondering if a bullet-list form works better here, with different sections for general change, security related changes etc.

CHANGELOG.md
```diff
@@ -1,5 +1,87 @@
...
+collections with `mongorestore`, the `autoIndexId` collection creation option will be ignored, since
+this is no longer supported with 8.2.0+.
+
+This release also addresses a few reported security vulnerabilities. Previous releases incorrectly
+included development-only library dependencies in the SBOM, which accounts for the `aws-sdk-go`
+vulnerability report. This package is still included in the released SBOM because of a bug in our
+SSDLC tooling. However, it has never been included in the actual tools code. In addition, this
```
mankawal, line 16:
```
This package is still included in the released SBOM because of a bug in our
SSDLC tooling
```
Do our customer know about SBOM, SSDLC etc? If not, i guess we should simplify this sentence

CHANGELOG.md
autarch: This is just the format we generally use for all releases. It's not ideal, but I don't think it's worth spending too much time on this.

CHANGELOG.md
autarch: The customers who care about this will know what this means, I think.

CHANGELOG.md
autarch: I should also note that merging this PR is currently the very last step in our release process. These notes have already gone out as-is in other places.

autarch: If you want to propose a change to the release notes format, I think the best way to do that is to make a PR that changes the template in `RELEASE.md` in this repo. FWIW, I don't like this format much either, but I would note that the list of tickets part is generated by Jira, so our flexibility is somewhat limited unless we also change whatever in Jira does this (I think it's some custom thing for us).

PR #839, opened 2025-08-18
TOOLS-3960 Remove the `15k_collection.js` test

autarch:
This was taking a very long time with Server 8.2.0 and causing timeouts.

The test itself was added to test a memory leak from the days when the tools were written in C++, and it no longer seems relevant.


rcownie:
LGTM

rcownie (approved):
LGTM

PR #841, opened 2025-08-26
TOOLS-3964: omit the main module from go list for sbom generation

shyam12860:
We fixed this earlier by omitting the `.Module` structs that didn't contain `.Path`, but that was removed by mistake in #810. 
Rather than add it back and depend on `.Path` being empty, I'm adding a filter to remove the main module. This is indicated in the go list json by `.Main` being set to true in the `.Module` struct.

### DIFF
--- a/cyclonedx.sbom.json
+++ b/cyclonedx.sbom.json
@@ -311,19 +311,6 @@
       "version": "v0.0.15"
     },
     {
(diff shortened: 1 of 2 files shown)


autarch:
But I think you need to regen the SBOM, don't you?

PR #861, opened 2026-01-22
Update mongosync branch and pins for passthrough tests

tdq45gj:
The resmoke setup was failing because it was using outdated server pins. This commit updates the pinned mongosync repo so that the passthrough tests can download the most recent server pins.


mmcclimon (approved):
LGTM, thanks!

PR #872, opened 2026-02-03
Centralize the connection logic for tests in one place and fix it for Atlas clusters

autarch:
This makes it possible to run tests against an Atlas cluster. As the code was before, this just didn't work.

### DIFF
--- a/common/testutil/testutil.go
+++ b/common/testutil/testutil.go
@@ -21,6 +21,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/x/mongo/driver/connstring"
 )
 
 // GetBareSession returns an mgo.Session from the environment or
@@ -40,15 +41,50 @@ func GetBareSession() (*mongo.Client, error) {
 // GetBareSessionProvider returns a session provider from the environment or
 // from a default host and port.
 func GetBareSessionProvider() (*db.SessionProvider, *options.ToolOptions, error) {
-	var toolOptions *options.ToolOptions
+	toolOptions, err := GetToolOptions()
+	if err != nil {
+		return nil, nil, fmt.Errorf(
+			"error getting tool options to create a bare session provider: %w",
+			err,
+		)
+	}
 
+	sessionProvider, err := db.NewSessionProvider(*toolOptions)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return sessionProvider, toolOptions, nil
+}
+
+const uriEnvVar = "TOOLS_TESTING_MONGOD"
+
+func GetToolOptions() (*options.ToolOptions, error) {
+	var toolOptions *options.ToolOptions
 	// get ToolOptions from URI or defaults
-	if uri := os.Getenv("TOOLS_TESTING_MONGOD"); uri != "" {
-		fakeArgs := []string{"--uri=" + uri}
-		toolOptions = options.New("mongodump", "", "", "", true, options.EnabledOptions{URI: true})
-		_, err := toolOptions.ParseArgs(fakeArgs)
+	if uri := os.Getenv(uriEnvVar); uri != "" {
+		parse, err := connstring.ParseAndValidate(uri)
 		if err != nil {
-			panic(fmt.Sprintf("Could not parse TOOLS_TESTING_MONGOD environment variable: %v", err))
+			return nil, fmt.Errorf(
+				"%#q from the %#q env var is not a valid connection string: %w",
+				uri,
+				uriEnvVar,
+				err,
+			)
+		}
+
+		fakeArgs := []string{"--uri=" + uri}
+		opts := options.EnabledOptions{Auth: parse.UsernameSet, URI: true}
+		toolOptions = options.New("mongodump", "", "", "", true, opts)
+
+		_, err = toolOptions.ParseArgs(fakeArgs)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"could not create toolOptions with %#q from the %#q env var: %w",
+				uri,
+				uriEnvVar,
+				err,
+			)
 		}
 	} else {
 		ssl := GetSSLOptions()
@@ -66,15 +102,13 @@ func GetBareSessionProvider() (*db.SessionProvider, *options.ToolOptions, error)
 			Namespace:  &options.Namespace{},
 		}
 	}
+
 	err := toolOptions.NormalizeOptionsAndURI()
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
-	sessionProvider, err := db.NewSessionProvider(*toolOptions)
-	if err != nil {
-		return nil, nil, err
-	}
-	return sessionProvider, toolOptions, nil
+
+	return toolOptions, nil
 }
 
 func GetBareArgs() []string {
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -4,7 +4,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "mongo-tools",
```
autarch: NPM made this change automatically when the linting was run. I guess something changed in newer NPM versions?

autarch: NPM made this change automatically when the linting was run. I guess something changed in newer NPM versions?

autarch:
I did this to make it more feasible to test the tools against arbitrary Atlas clusters. The tests don't all pass, but this does allow them to at least run.

common/testutil/testutil.go
```diff
@@ -40,15 +41,43 @@ func GetBareSession() (*mongo.Client, error) {
...
+func GetToolOptions() (*options.ToolOptions, error) {
+	var toolOptions *options.ToolOptions
 	// get ToolOptions from URI or defaults
 	if uri := os.Getenv("TOOLS_TESTING_MONGOD"); uri != "" {
+		parse, err := connstring.ParseAndValidate(uri)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"could not parse TOOLS_TESTING_MONGOD environment variable - %#q: %w",
+				uri,
+				err,
```
FGasper, line 72:
```suggestion
			return nil, fmt.Errorf(
				"%#q is invalid connection string (%#q): %w",
				"TOOLS_TESTING_MONGOD",
				uri,
				err,
```

common/testutil/testutil.go
```diff
@@ -40,15 +41,43 @@ func GetBareSession() (*mongo.Client, error) {
...
+		opts := options.EnabledOptions{Auth: parse.UsernameSet, URI: true}
+		toolOptions = options.New("mongodump", "", "", "", true, opts)
+
+		_, err = toolOptions.ParseArgs(fakeArgs)
 		if err != nil {
-			panic(fmt.Sprintf("Could not parse TOOLS_TESTING_MONGOD environment variable: %v", err))
+			return nil, fmt.Errorf(
+				"could not parse TOOLS_TESTING_MONGOD environment variable - %#q: %w",
+				uri,
+				err,
```
FGasper, line 86:
```suggestion
			return nil, fmt.Errorf(
				"%#q (%#q) is unusable: %w",
				"TOOLS_TESTING_MONGOD",
				uri,
				err,
```

FGasper: (I’m just thinking it makes sense to distinguish the two errors here.)

common/testutil/testutil.go
```diff
@@ -208,3 +235,14 @@ func MakeTempDir(t *testing.T) (string, func()) {
 	}
 	return dir, cleanup
 }
+
+// SkipForRemoteCluster will skip the test if `TOOLS_TESTING_MONGOD` is set, on the assumption that
+// this points to an external Atlas cluster, as opposed to a local MongoDB cluster for testing.
+func SkipForRemoteCluster(t *testing.T, reason string) {
+	if os.Getenv("TOOLS_TESTING_MONGOD") != "" {
```
FGasper: Do you mind making this a `const`? We reference it enough times.

common/testutil/testutil.go
```diff
@@ -208,3 +235,14 @@ func MakeTempDir(t *testing.T) (string, func()) {
 	}
 	return dir, cleanup
 }
+
+// SkipForRemoteCluster will skip the test if `TOOLS_TESTING_MONGOD` is set, on the assumption that
+// this points to an external Atlas cluster, as opposed to a local MongoDB cluster for testing.
```
FGasper: That’s not a safe assumption, though. I’ve used this in local testing.

What is the actual failure? Is there a ticket for it?

mongoexport/mongoexport_test.go
```diff
@@ -35,38 +35,10 @@ var (
...
-			SSL:        &ssl,
-			Connection: connection,
-			Auth:       &auth,
-			Verbosity:  &options.Verbosity{},
-			URI:        &options.URI{},
-		}
+func simpleMongoExportOpts() (Options, error) {
+	toolOptions, err := testutil.GetToolOptions()
+	if err != nil {
+		return Options{}, err
```
FGasper: Consider wrapping.

mongodump/mongodump_test.go
```diff
@@ -58,31 +58,10 @@ const (
...
-			SSL:        &ssl,
-			Connection: connection,
-			Auth:       &auth,
-			Verbosity:  &options.Verbosity{},
-			URI:        &options.URI{},
-		}
+func simpleMongoDumpInstance() (*MongoDump, error) {
+	toolOptions, err := testutil.GetToolOptions()
+	if err != nil {
+		return nil, err
```
FGasper: Consider wrapping.

FGasper (changes requested):
I’ve made a few notes.

More broadly: I don’t think we should assume that the env var is only for Atlas. Can we instead parse the connection string & determine therefrom that it’s Atlas? That logic does exist, as I recall.

common/testutil/testutil.go
autarch: There are various permissions issues when running against an Atlas cluster.

autarch: I don't think we would make a ticket for it. We don't expect these tests to all run successfully against Atlas. But I think having the tests skip in that case can be useful.

common/testutil/testutil.go
autarch: Done.

mongodump/mongodump_test.go
autarch: I added this for all the callers of `GetToolOptions`.

mongoexport/mongoexport_test.go
autarch: done

common/testutil/testutil.go
autarch: I used a slightly different message.

common/testutil/testutil.go
autarch: I used a slightly different message.

autarch:
I did change this to check if the URI is for Atlas. We don't have any code to do this. You're probably thinking of checking for the Atlas _proxy_, which I think is for free tier clusters. I added something a bit dumb that just checks for a mongodb domain in the URI.

common/testutil/testutil.go
```diff
@@ -208,3 +242,34 @@ func MakeTempDir(t *testing.T) (string, func()) {
...
+
+// SkipForAtlasCluster will skip the test if `TOOLS_TESTING_MONGOD` is an Atlas URI.
+func SkipForAtlasCluster(t *testing.T, reason string) {
+	uri := os.Getenv(uriEnvVar)
+	if uri == "" {
+		return
+	}
+
+	for _, d := range atlasDomains {
+		if strings.Contains(uri, d) {
```
FGasper, line 267: nit:
```suggestion
		if strings.HasSuffix(uri, d) {
```

FGasper (approved):
One nit, but otherwise LGTM.

Thanks!

common/testutil/testutil.go
```diff
@@ -208,3 +242,34 @@ func MakeTempDir(t *testing.T) (string, func()) {
 	}
 	return dir, cleanup
 }
+
+var atlasDomains = []string{
```
FGasper, line 246: Where did this list come from?

autarch: I got it from Glean.

common/testutil/testutil.go
autarch: This isn't right. The uri doesn't end with the domain name.

PR #873, opened 2026-02-03
TOOLS-4074 Add a script to create clusters with the `recordIdsReplicated` feature enabled

autarch:

PR #874, opened 2026-02-03
TOOLS-4074 Add a tool to perform manual dump/restore testing across two clusters

autarch:
This was written by Copilot using Claude Opus and reviewed by me.

The goal here was to create something that allows us to test the dump and restore between clusters with and without the replicated record IDs feature enabled. We don't have any non-passthrough tests that do this sort of thing across two clusters, so writing a test that we run manually with either local or Atlas clusters was the best approach for now.

### DIFF
--- a/manual-tests/record-ids-replicated/main.go
+++ b/manual-tests/record-ids-replicated/main.go
@@ -1,0 +1,752 @@
+// This tool tests the ability of mongodump and mongorestore to move data between clusters with and
+// without replicated record IDs enabled.
+//
+// It performs a full dump/restore cycle between two clusters and verifies the data matches.
+//
+// Usage:
+//
+//	go run main.go --source=<uri> --destination=<uri> [--keep-dump]
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"go.mongodb.org/mongo-driver/v2/bson"
+	"go.mongodb.org/mongo-driver/v2/mongo"
+	"go.mongodb.org/mongo-driver/v2/mongo/options"
+)
+
+const (
+	testDBName          = "rir_test_db"
+	docsPerCollection   = 50
+	concurrentOpsDBName = "rir_test_concurrent_db"
+	concurrentOpsColl   = "concurrent_coll"
+	dumpStartDelay      = 2 * time.Second
+	concurrentOpsPeriod = 5 * time.Second
+)
+
+// collectionSpec defines a collection to create for testing.
+type collectionSpec struct {
+	name       string
+	options    *options.CreateCollectionOptionsBuilder
+	indexes    []mongo.IndexModel
+	timeseries *options.TimeSeriesOptionsBuilder
+}
+
+// testContext holds the state needed throughout the test.
+type testContext struct {
+	srcURI      string
+	dstURI      string
+	keepDump    bool
+	srcClient   *mongo.Client
+	dstClient   *mongo.Client
+	dumpDir     string
+	barrierFile string
+}
+
+func main() {
+	srcURI := flag.String("source", "", "MongoDB URI for source cluster (required)")
+	dstURI := flag.String("destination", "", "MongoDB URI for destination cluster (required)")
+	keepDump := flag.Bool("keep-dump", false, "Keep the dump directory for inspection")
+	flag.Parse()
+
+	if *srcURI == "" || *dstURI == "" {
+		flag.Usage()
+		log.Fatal("Both --source and --destination are required")
+	}
+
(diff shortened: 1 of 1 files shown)


autarch:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/874?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#875** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/875?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#874** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/874?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/874?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#873** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/873?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#872** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/872?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

autarch:
See this [eng proposal](https://docs.google.com/document/d/1T9NSY0XsDQHBwvdwN4XEzJLnp3bcNV1JiX8WEfyDln4/edit?tab=t.0#heading=h.6t9nsdzj4hw) for more context.

manual-tests/record-ids-replicated/main.go
```diff
@@ -0,0 +1,721 @@
...
+	}
+
+	log.Fatal(
+		"Cannot determine repo root - run from repo root or manual-tests/record-ids-replicated directory",
+	)
+	return ""
+}
+
+func insertTestDocuments(ctx context.Context, coll *mongo.Collection, spec collectionSpec) error {
+	docs := make([]interface{}, docsPerCollection)
```
FGasper:
```suggestion
	docs := make([]any, docsPerCollection)
```

manual-tests/record-ids-replicated/main.go
```diff
@@ -0,0 +1,721 @@
...
+				{"reading", fmt.Sprintf("reading_%d", i)},
+			}
+		} else {
+			docs[i] = bson.D{
+				{"_id", primitive.NewObjectID()},
+				{"index", i},
+				{"field_a", fmt.Sprintf("value_a_%d", i)},
+				{"field_b", i * 2},
+				{"unique_field", fmt.Sprintf("unique_%s_%d", spec.name, i)},
+				{"sparse_field", func() interface{} {
```
FGasper:
```suggestion
				{"sparse_field", func() any {
```

manual-tests/record-ids-replicated/main.go
```diff
@@ -0,0 +1,721 @@
...
+	ctx context.Context,
+	coll *mongo.Collection,
+) (map[interface{}]bson.D, error) {
+	cursor, err := coll.Find(ctx, bson.D{})
+	if err != nil {
+		return nil, err
+	}
+	defer cursor.Close(ctx)
+
+	docs := make(map[interface{}]bson.D)
```
FGasper: While this does compile—TIL!—it seems like we should type this & type-assert later on.

FGasper (approved):
This is fine, but two broader notes:
1. It would seem sensible to refactor the verification part into something reusable. mongorestore’s `TestPipedDumpRestore`, for example, has no verification, and it’d be great to plug that in there.
2. Could this script gainfully incorporate archive-format?

manual-tests/record-ids-replicated/main.go
autarch: I'm confused. The correct type for the id is `any`. What else would we check? And later, we look things up in these maps by other ids. If there are two keys with equivalent values but different types, they _will_ be treated as different keys. See https://go.dev/play/p/cGjwL5Zy7gv for an example.

manual-tests/record-ids-replicated/main.go
autarch: I fixed all uses of `interface{}` to use `any`.

manual-tests/record-ids-replicated/main.go
FGasper: There can be BSON values that don’t fit into this map. For example, binary strings, or anything else with a variable-size buffer behind it. ([example](https://goplay.tools/snippet/9lmzLzWgcCF))

Migration Verifier’s way around this is `string([]byte{bsonType, buffer...})`.

autarch: I changed this map to use `bson.ObjectID` as the key type. The `_id` values for these docs are controlled by this code, and it always uses an object ID.

PR #875, opened 2026-02-03
TOOLS-4073 Support dump/restore between a cluster with replicated record IDs and one without

autarch:
This was written by Copilot using Claude Opus and reviewed by me. I also made some additional changes based on issues I saw when testing these changes.

There are a few things this does:

- Removes the `recordIdsReplicated` collection option when restoring collections. This option will be removed in a near-future Server version anyway, as it's an internal detail.
- Removes the record ID field from oplog entries (`rid`) when restoring them. Per discussion with Server folks working on this, the cluster will handle assigning record IDs and does not need this information.
- Skips `ci`, `cd`, and `km` ops when restoring from the oplog. These are new ops specific to replicated record IDs and index creation and deletion. They are internal operations that do not need to be replicated when restoring. The cluster will always manage indexes appropriately.

This also undoes some changes to the manual test code from a previous PR. That previous PR changed this code to expect `mongorestore` to fail when trying to run with `--oplogReplay` when going restoring a dump from a cluster with replicated record IDs to one without, as well as the inverse. We now expect this to work.

### DIFF
--- a/common/db/command.go
+++ b/common/db/command.go
@@ -151,11 +151,11 @@ func (sp *SessionProvider) GetNodeType() (NodeType, error) {
 		&bson.M{"ismaster": 1},
 	)
 	if result.Err() != nil {
-		return Unknown, result.Err()
+		return Unknown, fmt.Errorf("running `ismaster` command: %w", result.Err())
 	}
 	err = result.Decode(&masterDoc)
 	if err != nil {
-		return Unknown, err
+		return Unknown, fmt.Errorf("decoding `ismaster` response: %w", err)
 	}
 	if masterDoc.SetName != nil || masterDoc.Hosts != nil {
 		return ReplSet, nil
@@ -181,7 +181,7 @@ func (sp *SessionProvider) IsReplicaSet() (bool, error) {
 func (sp *SessionProvider) IsMongos() (bool, error) {
 	nodeType, err := sp.GetNodeType()
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("getting node type: %w", err)
 	}
 	return nodeType == Mongos, nil
 }
--- a/manual-tests/record-ids-replicated/main.go
+++ b/manual-tests/record-ids-replicated/main.go
@@ -1,10 +1,7 @@
 // This tool tests the ability of mongodump and mongorestore to move data between clusters with and
 // without replicated record IDs enabled.
 //
-// It performs a full dump/restore cycle between two clusters. When both clusters are of the same
-// type (both with or both without recordIdsReplicated), it verifies the data matches after
-// restore. When the clusters differ (one with and one without recordIdsReplicated), it verifies
-// that mongorestore fails when --oplogReplay is used.
+// It performs a full dump/restore cycle between two clusters and verifies the data matches.
 //
 // Usage:
 //
@@ -20,7 +17,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -49,16 +45,13 @@ type collectionSpec struct {
 
 // testContext holds the state needed throughout the test.
 type testContext struct {
-	srcURI                        string
-	dstURI                        string
-	keepDump                      bool
-	srcClient                     *mongo.Client
-	dstClient                     *mongo.Client
-	dumpDir                       string
-	simpleDumpDir                 string
-	barrierFile                   string
-	srcRecordIdsReplicatedEnabled bool
-	dstRecordIdsReplicatedEnabled bool
+	srcURI      string
+	dstURI      string
+	keepDump    bool
+	srcClient   *mongo.Client
+	dstClient   *mongo.Client
+	dumpDir     string
+	barrierFile string
 }
 
 func main() {
@@ -125,42 +118,12 @@ func run(srcURI, dstURI string, keepDump bool) error {
 		return err
 	}
 
-	if tc.srcRecordIdsReplicatedEnabled != tc.dstRecordIdsReplicatedEnabled {
-		// Cycle 1: with oplog replay — expect failure due to cluster type mismatch.
-		log.Println(
-			"Source and destination clusters differ in recordIdsReplicated support; verifying that mongorestore fails with --oplogReplay...",
-		)
-		if err := tc.runRestoreExpectingFailure(); err != nil {
-			return err
-		}
+	if err := tc.runRestore(); err != nil {
+		return err
+	}
 
-		// Cycle 2: without oplog replay — expect success.
-		log.Println("Verifying that mongorestore succeeds without --oplogReplay...")
-		if err := tc.cleanTestDBs(ctx); err != nil {
-			return err
-		}
-		if err := tc.runDumpWithoutOplog(); err != nil {
-			return err
-		}
-		defer tc.removeSimpleDumpDir()
-
-		if err := tc.removeSystemDBsFromSimpleDump(); err != nil {
-			return err
-		}
-		if err := tc.runRestoreWithoutOplog(); err != nil {
-			return err
-		}
-		if err := tc.verifyClustersMatch(ctx); err != nil {
-			return err
-		}
-	} else {
-		if err := tc.runRestore(); err != nil {
-			return err
-		}
-
-		if err := tc.verifyClustersMatch(ctx); err != nil {
-			return err
-		}
+	if err := tc.verifyClustersMatch(ctx); err != nil {
+		return err
 	}
 
 	return nil
@@ -202,53 +165,7 @@ func (tc *testContext) connectToClusters(ctx context.Context) error {
 	tc.dstClient = dstClient
 	log.Println("✅ Connected to destination cluster")
 
-	tc.srcRecordIdsReplicatedEnabled, err = hasRecordIdsReplicated(ctx, srcClient)
-	if err != nil {
-		return fmt.Errorf("checking recordIdsReplicated on source: %w", err)
-	}
-	log.Printf("Source cluster recordIdsReplicated: %t\n", tc.srcRecordIdsReplicatedEnabled)
-
-	tc.dstRecordIdsReplicatedEnabled, err = hasRecordIdsReplicated(ctx, dstClient)
-	if err != nil {
-		return fmt.Errorf("failed to check recordIdsReplicated on destination: %w", err)
-	}
-	log.Printf("Destination cluster recordIdsReplicated: %t\n", tc.dstRecordIdsReplicatedEnabled)
-
 	return nil
-}
-
-// hasRecordIdsReplicated checks whether the given cluster has the featureFlagRecordIdsReplicated
-// parameter enabled.
-func hasRecordIdsReplicated(ctx context.Context, client *mongo.Client) (bool, error) {
-	result := client.Database("admin").RunCommand(
-		ctx,
-		bson.D{
-			{"getParameter", 1},
-			{"featureFlagRecordIdsReplicated", 1},
-		},
-	)
-	if result.Err() != nil {
-		// This is what we get when the parameter name isn't recognized by the Server.
-		const invalidOptionsErrorCode = 72
-		if slices.Contains(mongo.ErrorCodes(result.Err()), invalidOptionsErrorCode) {
-			// If the command fails, the feature flag doesn't exist or isn't enabled. Or maybe something
-			// is totally broken, in which case we will find that out when we attempt other DB
-			// operations.
-			return false, nil
-		}
-		return false, result.Err()
-	}
-
-	var doc struct {
-		FeatureFlagRecordIdsReplicated struct {
-			Value bool `bson:"value"`
-		} `bson:"featureFlagRecordIdsReplicated"`
-	}
-	if err := result.Decode(&doc); err != nil {
-		return false, nil
-	}
-
-	return doc.FeatureFlagRecordIdsReplicated.Value, nil
 }
 
 func (tc *testContext) disconnectClients(ctx context.Context) {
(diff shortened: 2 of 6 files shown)


autarch:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/875?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#875** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/875?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/875?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#874** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/874?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#873** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/873?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#872** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/872?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **2 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>The `filterRecordIds` function seems to run unconditionally, which might affect the new feature's behavior</strong>
   - [Location](https://github.com/mongodb/mongo-tools/blob/ff4ac8fd5b316cae1a32f1e21e21b9c1e390e3b9/mongorestore/oplog.go#L230-L233): `mongorestore/oplog.go:L230-L233`
   - Impact: Record IDs could be stripped from entries even when the feature is enabled, potentially bypassing the intended replication functionality.
   - Suggestion: Consider wrapping the call in a conditional check so it only runs when `recordIdsReplicatedEnabled` is false.

2. <strong>Broad error handling in feature detection might lead to unexpected silent disablement</strong>
   - [Location](https://github.com/mongodb/mongo-tools/blob/ff4ac8fd5b316cae1a32f1e21e21b9c1e390e3b9/common/db/command.go#L205-L210): `common/db/command.go:L205-L210`
   - Impact: Transient issues like network or auth errors could be treated as a "missing feature," causing the restore to proceed in a degraded state.
   - Suggestion: It might be safer to inspect the error and only return `false` if the command is truly unsupported, while propagating other critical errors.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

common/db/command.go
```diff
@@ -187,6 +187,40 @@ func (sp *SessionProvider) IsMongos() (bool, error) {
...
+		context.Background(),
+		bson.D{
+			{"getParameter", 1},
+			{"featureFlagRecordIdsReplicated", 1},
+		},
+	)
+	if result.Err() != nil {
+		// If the command fails, the feature flag doesn't exist or isn't enabled. Or maybe something
+		// is totally broken, in which case we will find that out when we attempt other DB
+		// operations.
```
FGasper: Can we instead check for an error code?

The failure may be intermittent, so ISTM we really shouldn’t just throw it away. At the very least it should be logged, but even then we’d want a good comment about why we can’t reliably fail.

mongorestore/metadata.go
```diff
@@ -299,12 +299,33 @@ func (restore *MongoRestore) UpdateAutoIndexId(options bson.D) bson.D {
...
+		return options
+	}
+
+	for i, elem := range options {
+		if elem.Key == "recordIdsReplicated" {
+			log.Logvf(
+				log.DebugLow,
+				"removing recordIdsReplicated option (target server does not support this feature)",
+			)
+			return append(options[:i], options[i+1:]...)
```
FGasper: IMO slices.Delete() would be clearer.

common/db/oplog.go
```diff
@@ -20,4 +20,5 @@ type Oplog struct {
 	TxnNumber   *int64              `bson:"txnNumber,omitempty"`
 	PrevOpTime  bson.Raw            `bson:"prevOpTime,omitempty"`
 	MultiOpType *int                `bson:"multiOpType,omitempty"`
+	RecordId    *int64              `bson:"rid,omitempty"`
```
FGasper: Record IDs are not all int64s, though. In clustered collections they’re binary strings. And in 5.0 timeseries they’re _strings_.

FGasper (changes requested):
The record ID type issue must be fixed.

We should also add tests to catch this & any other new/unknown BSON types for record IDs.

mongorestore/oplog.go
```diff
@@ -573,6 +587,39 @@ func (restore *MongoRestore) filterHs(op db.Oplog) (db.Oplog, error) {
 	return op, nil
 }
 
+// filterRecordIds removes 'recordId' entries from ops, including nested `applyOps` ops. This is
+// needed when restoring to a server that does not have the `featureFlagRecordIdsReplicated` feature
+// enabled.
+func (restore *MongoRestore) filterRecordIds(op db.Oplog) (db.Oplog, error) {
```
visemet: Why are we trying to support mongodump --oplog + mongorestore --oplogReplay from a source system which uses the recordIdsReplicated collection option by default? The only requirement is an offline export of the data.

mongorestore/oplog.go
```diff
@@ -573,6 +587,39 @@ func (restore *MongoRestore) filterHs(op db.Oplog) (db.Oplog, error) {
...
+	// Check for and filter nested applyOps ops
+	if op.Operation == "c" && isApplyOpsCmd(op.Object) {
+		ops, err := unwrapNestedApplyOps(op.Object)
+		if err != nil {
+			return db.Oplog{}, err
+		}
+
+		filtered := make([]db.Oplog, len(ops))
+		for i, v := range ops {
+			filtered[i], err = restore.filterRecordIds(v)
```
visemet: Is there a test case which shows this logic is necessary? My understanding is `HandleNonTxnOp()` already unrolls top-level applyOps oplog entries. The oplog will not contain a nested applyOps oplog entry in any server version an "rid" field may be present.

I would have expected omitting the RecordId field from the Oplog struct to work as desired automatically. What am I missing?

https://github.com/mongodb/mongo-tools/blob/8279a5a59becfb58763f12305323373ae3c1ef28/mongorestore/oplog.go#L307-L324

[note] Prior to [SERVER-45033](https://jira.mongodb.org/browse/SERVER-45033) (7.0), the server would record the applyOps command the client had specified which means that the applyOps oplog entry could have another applyOps oplog entry within it. There was a finite limit imposed by the server of [10 nested applyOps](https://github.com/mongodb/mongo/blob/r4.4.29/src/mongo/db/commands/apply_ops_cmd.cpp#L88).

mongorestore/metadata.go
```diff
@@ -299,12 +299,33 @@ func (restore *MongoRestore) UpdateAutoIndexId(options bson.D) bson.D {
 	return options
 }
 
+// filterRecordIdsReplicatedOption removes the `recordIdsReplicated` option from the collection
+// options if the target server doesn't have the `featureFlagRecordIdsReplicated` feature enabled.
+func (restore *MongoRestore) filterRecordIdsReplicatedOption(options bson.D) bson.D {
+	if restore.recordIdsReplicatedEnabled {
```
visemet: If featureFlagRecordIdsReplicated is enabled by default on the target system, then the create command is going to create a collection which has all replica set members using the same mapping of record ID → BSON document. The target system will do so even if `recordIdsReplicated: true` is omitted from the create command request.

There is zero interest in preserving the absence of `recordIdsReplicated` in the collection metadata as `recordIdsReplicated: false` on a target system with featureFlagRecordIdsReplicated enabled by default. Why not always remove the `recordIdsReplicated` collection option and let the target system decide based on its own default? This means there would be no need to run the getParameter command earlier.

visemet (changes requested):
Adding my own thoughts here.

mongorestore/metadata.go
visemet: @henrikedin @erreina would you please confirm that following [SERVER-85589](https://jira.mongodb.org/browse/SERVER-85589), the `recordIdsReplicated` option will no longer appear in op=c create collection oplog entries? If not, then there would be another place in mongorestore to always remove the `recordIdsReplicated` option from.

https://github.com/mongodb/mongo-tools/blob/8279a5a59becfb58763f12305323373ae3c1ef28/mongorestore/oplog.go#L363-L364

erreina: It will not be on the `options` in the create collection oplog entries, but it will now be on the `o2` field.

Always removing the `recordIdsReplicated` field from the collection metadata and let the target decide seems reasonable to me.

common/db/oplog.go
FGasper: Suggest: `*bson.RawValue`

mongorestore/metadata.go
FGasper: Or, lo.Filter.

common/db/command.go
autarch: Done.

common/db/oplog.go
autarch: Done.

mongorestore/oplog.go
autarch: This is identical to the logic we use to filter out `ui` (UUID) entries from ops. You're right that we _also_ unroll the nested ops later in the `HandleNonTxnOp` code. But I'd just as soon leave this like it is so that the logic for UUID and record ID filtering is the same.  
  
As for test cases, there's the manual tests that were added in a previous PR.

mongorestore/metadata.go
autarch: Done.

mongorestore/metadata.go
erreina: Sorry, there was an small change on direction. The `recordIdsReplicated` field will not be on the `o2` field as I mentioned above, it will be in the `object` (`o`) field similar to what we have today, except it will not be a collection option and it will need to be filter out when writing a create oplog to a system that does not have the feature flag enabled.

autarch: There was a discussion in the Eng Proposal doc for this work where we agreed that we will simply not allow oplog replay between the two types of clusters. So I've updated this branch to reflect that.

Note that we _will_ support dump/restore between the cluster types as long as oplog replay isn't being used.

mongorestore/oplog.go
visemet: Yes I am familiar with [the logic](https://github.com/mongodb/mongo-tools/blob/3a9ea8cc31c1ef8ca23c7cf4b1e65a6c28e883de/mongorestore/oplog.go#L700-L722) to filter out the `ui` field. Oplog entries which contain the `ui` field include MongoDB versions prior to 7.0 (starting from MongoDB version 3.6). However, oplog entries which contain the `rid` field are only 8.3 and newer. This means nested applyOps is only a concern for the former and not for the latter. We ought to be able to simplify the latter.

> As for test cases, there's the manual tests that were added in a previous PR.

Having manual tests for an end-to-end scenario between ASC and DSC due to complications in setting up DSC clusters in an automated manner is something ok to accept for now. It is an entirely separate matter as an engineering practice to rely _only_ on manual testing for code changes.

From the end-to-end scenario which was manually run it is possible to extract the oplog entries which contain the `rid` field and add that to the mongorestore/testdata/ directory. This snapshot can then be restored against an ASC cluster like other test cases do.

https://github.com/mongodb/mongo-tools/blob/3a9ea8cc31c1ef8ca23c7cf4b1e65a6c28e883de/mongorestore/oplog_test.go#L796-L812

mongorestore/oplog.go
```diff
@@ -208,6 +208,20 @@ func (restore *MongoRestore) HandleNonTxnOp(oplogCtx *oplogContext, op db.Oplog)
 		return fmt.Errorf("error filtering UUIDs from oplog: %v", err)
 	}
 
+	// These ops are seen when replicated record IDs are enabled. They are related to internal
+	// bookkeeping for the cluster and do not need to be replicated on restore.
+	if op.Operation == "cd" || op.Operation == "ci" || op.Operation == "km" {
```
visemet: Is there a significance to filtering the `'cd'`, `'ci'`, and `'km'` oplog entry types here rather than in MongoRestore.HandleOp()? No-op oplog entries and namespace filtering occurs in HandleOp().

`oplogCtx.totalOps` appears informational rather than functional that perhaps the difference of where to do the filtering is minor.

https://github.com/mongodb/mongo-tools/blob/3a9ea8cc31c1ef8ca23c7cf4b1e65a6c28e883de/mongorestore/oplog.go#L156-L163

mongorestore/oplog.go
```diff
@@ -540,6 +554,45 @@ func (restore *MongoRestore) filterUUIDs(op db.Oplog) (db.Oplog, error) {
...
+// collection option from command objects, including nested `applyOps` ops.
+func (restore *MongoRestore) filterRecordIds(op db.Oplog) (db.Oplog, error) {
+	// Remove recordId from oplog entries
+	op.RecordId = nil
+
+	// Remove recordIdsReplicated option from command objects (e.g. create commands).
+	// This allows oplog replay across cluster types regardless of whether the source
+	// cluster had the feature enabled.
+	if op.Operation == "c" {
+		_, _ = bsonutil.RemoveKey("recordIdsReplicated", &op.Object)
```
visemet, line 558: [note to self] This will also remove `recordIdsReplicated` from a collMod oplog entry in addition to a create collection oplog entry. While we don't expect such a collMod oplog entry to exist in any DSC cluster's oplog it may exist in the future for some ASC's cluster oplog as part of how we eventually set `recordIdsReplicated: true` on existing ASC clusters. The applyOps command of a collMod oplog entry will succeed even if there are no additional arguments.

```javascript
MongoDB Enterprise foo:PRIMARY> db.adminCommand({applyOps: [{op: "c", ns: "test.$cmd", o: {collMod: "mycoll", recordIdsReplicated: true}}]})
{
	"applied" : 1,
	"code" : 72,
	"codeName" : "InvalidOptions",
	"errmsg" : "Cannot set recordIdsReplicated to true",
	"results" : [
		false
	],
```

```javascript
MongoDB Enterprise foo:PRIMARY> db.adminCommand({applyOps: [{op: "c", ns: "test.$cmd", o: {collMod: "mycoll"}}]})
{
	"applied" : 1,
	"results" : [
		true
	],
	"ok" : 1,
```

visemet (changes requested):

mongorestore/oplog.go
autarch: I just didn't notice we were already doing some op filtering there. I moved this.

mongorestore/oplog.go
autarch: For the filtering logic, I removed the unrolling bit.

For the testing, I am going to be working on adding automated testing across cluster types as an immediate follow-up to this work. But I would really like to merge this and get a new tools release out sooner rather than later. This was covered in the EP for this project.

autarch: Unrolling removed.

common/db/oplog.go
```diff
@@ -18,4 +18,5 @@ type Oplog struct {
 	TxnNumber   *int64         `bson:"txnNumber,omitempty"`
 	PrevOpTime  bson.Raw       `bson:"prevOpTime,omitempty"`
 	MultiOpType *int           `bson:"multiOpType,omitempty"`
+	RecordId    *bson.RawValue `bson:"rid,omitempty"`
```
visemet: Can the `RecordId` field be removed now? We only ever assign it to `nil` and would get the same effect by its omission.

https://github.com/mongodb/mongo-tools/blob/18e52c2a078bd5b8bdafab5aea7798db53dfa484/mongorestore/oplog.go#L554-L557

visemet (changes requested):
> > From the end-to-end scenario which was manually run it is possible to extract the oplog entries which contain the `rid` field and add that to the mongorestore/testdata/ directory. This snapshot can then be restored against an ASC cluster like other test cases do.
> > 
> > https://github.com/mongodb/mongo-tools/blob/3a9ea8cc31c1ef8ca23c7cf4b1e65a6c28e883de/mongorestore/oplog_test.go#L796-L812
> > 
> > https://github.com/mongodb/mongo-tools/pull/875#discussion_r2891500366
> 
> For the testing, I am going to be working on adding automated testing across cluster types as an immediate follow-up to this work. But I would really like to merge this and get a new tools release out sooner rather than later. This was covered in the EP for this project.
> 
> https://github.com/mongodb/mongo-tools/pull/875#discussion_r2892135632

Spend 1 hour between you and Claude of running the manual test again to generate [the oplog.bson file](https://github.com/mongodb/mongo-tools/blob/18e52c2a078bd5b8bdafab5aea7798db53dfa484/manual-tests/record-ids-replicated/main.go#L406-L407) + other files from mongodump and adding them to the mongorestore/testdata/ directory under a new mongorestore/oplog_test.go test case. Doing so will increase our confidence in the upcoming tools release. The urgency for the DSC → ASC direction is not so dire that you skip this.

common/db/oplog.go
autarch: Yes, good catch. This served a purpose in an earlier version of the code (I think), but it's not longer needed.

autarch: Removed.

common/db/command.go
```diff
@@ -151,11 +151,11 @@ func (sp *SessionProvider) GetNodeType() (NodeType, error) {
 		&bson.M{"ismaster": 1},
 	)
 	if result.Err() != nil {
-		return Unknown, result.Err()
+		return Unknown, fmt.Errorf("error running `ismaster` command: %w", result.Err())
```
FGasper: nit: “error” is redundant here

manual-tests/record-ids-replicated/main.go
FGasper, line 1: The changes to this script aren’t described in the PR description. Do you mind adding that explanation?

mongorestore/oplog.go
FGasper: Should we assert `op.Object[0].Key == "create"` here to constrain scope here?

FGasper (approved):
This looks OK, though I do think constraining the filtering to just what we expect would make sense, and some reference to the manual-test script’s changes would also be good.

common/db/command.go
autarch: Fixed, along with the others in this diff.

manual-tests/record-ids-replicated/main.go
autarch: Added.

mongorestore/oplog.go
autarch: I added this and re-ran the end-to-end test to confirm that this still works as expected.

mongorestore/oplog.go
autarch: We discussed the testing offline and agreed we could merge this PR as-is. Automated end-to-end testing will be added in a follow-up.

PR #876, opened 2026-02-04
TOOLS-4073 Simplify `MongoRestore.ApplyOps` method

autarch:
This method was accepting ops as a `[]any` slice, even though it was only ever called with a single value, and that single value can (and should) always be a `db.Oplog` struct. I changed the code to work this way, which simplifies adding some new logging in this method.


FGasper (approved):
The fact that we only apply 1 op at a time is, I suspect, a major performance bottleneck for mongorestore. That wasn’t very obvious to me at first precisely because of the complexity you’ve removed here. Thank you!

PR #883, opened 2026-02-11
TOOLS-4071 Fix mongorestore collMod an index lacking options

tdq45gj:
This commit fixes a bug that mongorestore only recognizes `hidden` and `expireAfterSeconds` options in an index collMod oplog. The change allows mongorestore to pass any collMod index options to the restored index.


common/idx/index_catalog.go
```diff
@@ -337,39 +325,11 @@ func (i *IndexCatalog) collMod(database, collection string, indexModValue any) e
...
-	if expireKeyError == nil {
-		newExpire, ok := expireValue.(int64)
-		if !ok {
-			return errors.Errorf(
-				"expireAfterSeconds must be a number (found %v of type %T): %v",
-				expireValue,
-				expireValue,
-				indexMod,
-			)
+	for k, v := range bsonutil.ToMap(indexMod) {
```
FGasper: Nit: why ToMap here? It seems almost as easy to iterate the bson.D.

FGasper (changes requested):
This sort of “deny-list” approach seems out of step with our usual stance of vetting anything we send to the destination.

Discussed over DM that it should be an allow-list instead (like mongosync).

common/idx/index_catalog.go
```diff
@@ -337,39 +325,18 @@ func (i *IndexCatalog) collMod(database, collection string, indexModValue any) e
...
-		}
-		updateHidden(matchingIndex, newHidden)
-	}
+		if k == "expireAfterSeconds" || k == "hidden" || k == "prepareUnique" || k == "unique" {
+			matchingIndex.Options[k] = element.Value
 
-	if expireKeyError != nil && hiddenKeyError != nil {
-		return errors.Errorf("must specify expireAfterSeconds or hidden: %v", indexMod)
+		} else {
+			return errors.Errorf("unknown index option: %v", k)
```
FGasper, line 338: How does this logic square with the `validIndexOptions` in common/bsonutil/indexes.go? It seems like there are many more attributes than are listed here.

FGasper (changes requested):
It looks like there’s an internal discrepancy here … ?

common/idx/index_catalog.go
tdq45gj: I'm following the collMod index options in the [documentation](https://www.mongodb.com/docs/manual/reference/command/collMod/#mongodb-collflag-index). I think this differs from `validIndexOptions` because collMod does not support every valid index option. I tried to collMod `2dsphereIndexVersion` and got an IDLUnknownField error.

FGasper (approved):
LGTM. Thank you!

PR #884, opened 2026-02-12
TOOLS-4087 / TOOLS-4088 Disable unnecessary tasks in waterfall

tdq45gj:
This commit disables task `deliberate_fail_to_test_build_baron_context_rules` and `run_manual_generated_tasks` in waterfall.

### DIFF
--- a/mongodump_passthrough/tasks.yml
+++ b/mongodump_passthrough/tasks.yml
@@ -30,6 +30,7 @@ tasks:
             echo "Deliberate fail test should give BF on mongodump_passthrough context"
             echo "exit 1"
             exit 1
+    patch_only: true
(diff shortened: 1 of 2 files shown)


mongodump_passthrough/tasks.yml
mankawal, line 63: Do we need to update the code comment here?

mankawal (approved):
LGTM, with a trivial nit.

mongodump_passthrough/tasks.yml
tdq45gj: I think the comment is still correct. This patch just restricts the task to be patch-only, and shouldn't break the workflow for the generator.

PR #885, opened 2026-02-12
TOOLS-4059 / TOOLS-4068 Fix qa-tests-7.0 failure

tdq45gj:
[SERVER-96103](https://jira.mongodb.org/browse/SERVER-96103) is marked as fixed for 7.0.31. This commit fixes a qa BF that happens on 7.0 due to a mismatch in fix version.


FGasper (approved):
LGTM. (I’m assuming you’ve actually verified in the waterfall that this works as intended.)

PR #889, opened 2026-02-19
TOOLS-4070 Mongorestore pass bypassDocumentValidation to applyOps

tdq45gj:
This commit fixes a bug where `bypassDocumentValidation` is not passed to oplog restore commands even if the mongorestore is started with CLI flag `--bypassDocumentValidation`.


mongodump/oplog_dump_test.go
```diff
@@ -276,3 +277,134 @@ func createIndexesAndRunCollModPrepareUnique(ctx context.Context) error {
 
 	return nil
 }
+
+func TestOplogDumpBypassDocumentValidation(t *testing.T) {
```
mmcclimon, line 283: I'm not totally sure what this test is doing here; I think just to create the oplog BSON to use for the restore test. Do we usually keep these tests around? (I can imagine yes, so that we can regenerate them if we want, but I could also imagine no, because they're not really testing mongodump.)

mmcclimon (approved):
LGTM % one question about the dump test; thanks! (Also, thanks for making this PR super easy to review.)

mongodump/oplog_dump_test.go
tdq45gj: I'm inclined to keep these tests so that we can regenerate the test files for mongorestore. Despite not the primary reason, sometimes they might also add test coverage for mongodump.

PR #891, opened 2026-02-23
TOOLS-4075 Add integration test for restoring TTL `collMod` oplog

tdq45gj:
This commit adds integration tests for verifying that mongodump+mongorestore handles a `collMod` oplog that converts a non-TTL index into TTL correctly. This was a bug fixed in [TOOLS-4071](https://jira.mongodb.org/browse/TOOLS-4071).


mongodump/oplog_dump_test.go
```diff
@@ -410,3 +411,122 @@ func createCollectionWithValidatorAndInsertBypassValidation(ctx context.Context,
...
+			t.Errorf("error getting FCV: %v", err)
+		}
+		t.Skipf("Requires server with FCV 6.0 or later; found %v", fcv)
+	}
+
+	testCollName := testCollectionNames[0]
+	err = session.Database(testDB).Collection(testCollName).Drop(ctx)
+	require.NoError(t, err)
+
+	err = session.Database(testDB).CreateCollection(ctx, testCollName)
```
nickweinberger, line 432: Why is it necessary to drop and create the collection?

mongodump/oplog_dump_test.go
```diff
@@ -410,3 +411,122 @@ func createCollectionWithValidatorAndInsertBypassValidation(ctx context.Context,
...
+	md.OutputOptions.Out = "collMod_ttl_index"
+
+	require.NoError(t, md.Init())
+
+	// Enable a failpoint so that the test can create oplogs during dump.
+	failpoint.ParseFailpoints(failpoint.PauseBeforeDumping)
+	defer failpoint.Reset()
+
+	go func() {
+		convertIndexToTTL(ctx, t)
```
nickweinberger, line 466: Know you're following existing patterns here but I think calling testify's `require` (which `convertIndexToTTL` does) within a goroutine is bad practice in go. I think on failure it only exits the goroutine but not the whole test, so the test keeps running and it can cause weird panics that are harder to debug. Up to you whether you want to address or just note as tech debt.

mongodump/oplog_dump_test.go
```diff
@@ -410,3 +411,122 @@ func createCollectionWithValidatorAndInsertBypassValidation(ctx context.Context,
...
+
+	session, err := testutil.GetBareSession()
+	require.NoError(t, err)
+
+	fcv := testutil.GetFCV(session)
+	if cmp, err := testutil.CompareFCV(fcv, "6.0"); err != nil || cmp < 0 {
+		if err != nil {
+			t.Errorf("error getting FCV: %v", err)
+		}
+		t.Skipf("Requires server with FCV 6.0 or later; found %v", fcv)
```
nickweinberger: Why exactly 6.0+ required? Maybe useful to add into the error.

nickweinberger: Would also ideally put this in a helper for reusability, but I haven't worked much w/ the tools testing so don't know if that's worth it.

mongodump/oplog_dump_test.go
```diff
@@ -410,3 +411,122 @@ func createCollectionWithValidatorAndInsertBypassValidation(ctx context.Context,
 		require.Error(t, err)
 	}
 }
+
+func TestOplogDumpCollModTTL(t *testing.T) {
```
nickweinberger, line 417: Is this test intended as a generator for the restore test's testdata fixtures (like `TestOplogDumpBypassDocumentValidation`) ? If so, might benefit from a comment. Otherwise, mind explaining just what this test is doing?

mongorestore/oplog_test.go
```diff
@@ -880,3 +880,62 @@ func TestOplogRestoreBypassDocumentValidation(t *testing.T) {
...
+	//nolint:errcheck
+	defer session.Disconnect(ctx)
+
+	fcv := testutil.GetFCV(session)
+	if cmp, err := testutil.CompareFCV(fcv, "6.0"); err != nil || cmp < 0 {
+		if err != nil {
+			t.Errorf("error getting FCV: %v", err)
+		}
+		t.Skipf("Requires server with FCV 6.0 or later; found %v", fcv)
+	}
```
nickweinberger: Same comment about helper.

nickweinberger (approved):
Core logic from the restore test looks good, left a couple questions partly for my own understanding and partly as suggestions. Approving % addressing these questions at your discretion.

mongodump/oplog_dump_test.go
tdq45gj: Dropping the collection because the collection might have been created by another test. Unlike mongosync, the tools tests don't clean up between runs.

mongodump/oplog_dump_test.go
tdq45gj: Ideally `convertIndexToTTL` would run as a callback function in the failpoint. But the failpoints in the tools are simply calling `sleep`.

mongodump/oplog_dump_test.go
tdq45gj: It's to generate test files for the corresponding mongorestore test. Added a comment.

mongodump/oplog_dump_test.go
tdq45gj: 6.0+ is required because collMod ttl index is not supported before 6.0. Added the helper function.

PR #892, opened 2026-02-23
Upload passthrough archives

tdq45gj:
This commit makes the passthrough tests upload archive files that are saved in designated directory.

### DIFF
--- a/mongodump_passthrough/functions.yml
+++ b/mongodump_passthrough/functions.yml
@@ -27,7 +27,7 @@ variables:
   # And when you change the _mongosync_pin, you also need to update other pins
   # which get passed as environment variables to scripts in mongosync/evergreen/scripts.
   #
-  _mongosync_pin: &_mongosync_pin "f259dbd3841f16c52f757bf16573a08e2ab06c57"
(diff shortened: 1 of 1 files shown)


mongodump_passthrough/functions.yml
```diff
@@ -604,6 +604,23 @@ functions:
...
+        target: tools_output.tgz
+        source_dir: "src/resmoke/tools_output"
+        include:
+          - "**"
+    - command: s3.put
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file: tools_output.tgz
+        remote_file: mongosync/${build_variant}/${revision}/tools_output/${task_name}/${build_id}/tools_output.tgz
```
autarch: Should this be `mongo-tools` instead of `mongosync` for the first element in the path?

autarch (approved):
LGTM % one question

mongodump_passthrough/functions.yml
tdq45gj: Yes, tools should just upload to `mongo-tools`. Corrected.

PR #895, opened 2026-02-25
TOOLS-4102 Add a `mise.toml` with all of our tools

autarch:
This adds a `mise.toml`, as well as some infrastructure for installing mise itself. However, nothing uses this yet. That will come in follow-up PRs.


etc/mise.run.sh
autarch, line 1: This is copied from https://mise.run. I just don't want to do the `curl ... | sh` thing in the tools CI.

etc/mise.run.sh
```diff
@@ -0,0 +1,314 @@
...
+  else
+    echo "tar.gz"
+  fi
+}
+
+tar_supports_zstd() {
+  # tar is bsdtar or version is >= 1.31
+  if tar --version | grep -q 'bsdtar' && command -v zstd >/dev/null 2>&1; then
+    true
+  elif tar --version | grep -q '1\.(3[1-9]|[4-9][0-9]'; then
```
nickweinberger, line 89: regex is missing a closing ) right, so won't this always fall back?

nickweinberger: Know you just copied this but that seems off.

nickweinberger (approved):
LGTM since it's not being used yet, although I think I did notice one problem in mise.run.sh.

etc/mise.run.sh
autarch: It's actually more broken than just missing a closing paren - https://github.com/jdx/mise/pull/8430

PR #899, opened 2026-02-26
TOOLS-4109 Update manual test for replicated record ID clusters to expect failure when oplog replay is enabled

autarch:
We will not support dump/restore with oplog replay between a cluster with replicated record IDs and one with out this.

This PR updates our manual test code to expect mongorestore to _fail_ in a such scenario. However, we still support dump/restore without oplog replay in this case, so we test that as well.

### DIFF
--- a/manual-tests/record-ids-replicated/main.go
+++ b/manual-tests/record-ids-replicated/main.go
@@ -1,7 +1,10 @@
 // This tool tests the ability of mongodump and mongorestore to move data between clusters with and
 // without replicated record IDs enabled.
 //
-// It performs a full dump/restore cycle between two clusters and verifies the data matches.
+// It performs a full dump/restore cycle between two clusters. When both clusters are of the same
+// type (both with or both without recordIdsReplicated), it verifies the data matches after
+// restore. When the clusters differ (one with and one without recordIdsReplicated), it verifies
+// that mongorestore fails when --oplogReplay is used.
 //
 // Usage:
 //
@@ -17,6 +20,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -45,13 +49,16 @@ type collectionSpec struct {
 
 // testContext holds the state needed throughout the test.
 type testContext struct {
-	srcURI      string
-	dstURI      string
-	keepDump    bool
-	srcClient   *mongo.Client
-	dstClient   *mongo.Client
-	dumpDir     string
-	barrierFile string
+	srcURI                        string
+	dstURI                        string
+	keepDump                      bool
+	srcClient                     *mongo.Client
+	dstClient                     *mongo.Client
+	dumpDir                       string
+	simpleDumpDir                 string
+	barrierFile                   string
+	srcRecordIdsReplicatedEnabled bool
+	dstRecordIdsReplicatedEnabled bool
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


manual-tests/record-ids-replicated/main.go
```diff
@@ -165,9 +202,55 @@ func (tc *testContext) connectToClusters(ctx context.Context) error {
 	tc.dstClient = dstClient
 	log.Println("✅ Connected to destination cluster")
 
+	tc.srcRecordIdsReplicatedEnabled, err = hasRecordIdsReplicated(ctx, srcClient)
+	if err != nil {
+		return fmt.Errorf("failed to check recordIdsReplicated on source: %w", err)
```
FGasper: nit:
```suggestion
		return fmt.Errorf("checking recordIdsReplicated on source: %w", err)
```

FGasper: … and so forth elsewhere.

manual-tests/record-ids-replicated/main.go
```diff
@@ -165,9 +202,55 @@ func (tc *testContext) connectToClusters(ctx context.Context) error {
...
+	}
+	log.Printf("Source cluster recordIdsReplicated: %v", tc.srcRecordIdsReplicatedEnabled)
```
FGasper:
```suggestion
	log.Printf("Source cluster recordIdsReplicated: %t\n", tc.srcRecordIdsReplicatedEnabled)
```

manual-tests/record-ids-replicated/main.go
```diff
@@ -165,9 +202,55 @@ func (tc *testContext) connectToClusters(ctx context.Context) error {
...
+
+	tc.dstRecordIdsReplicatedEnabled, err = hasRecordIdsReplicated(ctx, dstClient)
+	if err != nil {
+		return fmt.Errorf("failed to check recordIdsReplicated on destination: %w", err)
+	}
+	log.Printf("Destination cluster recordIdsReplicated: %v", tc.dstRecordIdsReplicatedEnabled)
```
FGasper:
```suggestion
	log.Printf("Destination cluster recordIdsReplicated: %t\n", tc.dstRecordIdsReplicatedEnabled)
```

manual-tests/record-ids-replicated/main.go
```diff
@@ -165,9 +202,55 @@ func (tc *testContext) connectToClusters(ctx context.Context) error {
...
+		ctx,
+		bson.D{
+			{"getParameter", 1},
+			{"featureFlagRecordIdsReplicated", 1},
+		},
+	)
+	if result.Err() != nil {
+		// This is the "InvalidOptions" error code, which is what we get when the parameter name
+		// isn't recognized by the Server.
+		if slices.Contains(mongo.ErrorCodes(result.Err()), 72) {
```
FGasper: nit: `const ErrCodeInvalidOptions = 72`

manual-tests/record-ids-replicated/main.go
```diff
@@ -448,6 +599,31 @@ func (tc *testContext) runRestore() error {
...
+	restoreCmd.Dir = repoRoot()
+	restoreCmd.Stdout = os.Stdout
+	restoreCmd.Stderr = os.Stderr
+
+	if err := restoreCmd.Run(); err == nil {
+		return fmt.Errorf(
+			"mongorestore succeeded but should have failed: clusters differ in recordIdsReplicated support",
+		)
+	}
+	log.Println("✅ mongorestore failed as expected")
```
FGasper, line 623: Can we check somehow that the failure is the actual expected failure? Maybe tee the log to a buffer & do a regexp check on that?

FGasper (changes requested):
A few nits, and a request for a more detailed check of the expected failure case.

Thank you!

manual-tests/record-ids-replicated/main.go
autarch: Done

manual-tests/record-ids-replicated/main.go
autarch: Yes, but the error message is added in the PR above this in the stack. I will change the test code in that branch.

autarch: https://github.com/mongodb/mongo-tools/pull/875/changes#diff-25bb8969f1ee4c6d98b677fc91e25a1d4fb706c47efcd0e0cd23f296e0ded393R625

FGasper (approved):
LGTM. Thank you!

PR #900, opened 2026-02-27
Use a new S3 bucket instead of `mciuploads`

autarch:
This is being done for two reasons.

First, we want to move projects away from all using the shared `mciuploads` bucket.

Second, the credentials to access `mciuploads` are only available for the DB Tools Evergreen project for the `master` branch. When we make a PR off a different branch, we cannot see those credentials. That means those branches cannot interact with S3 in CI, which leads to lots of CI failures. Switching to a new bucket and using `role_arn` to access that bucket means this works on any branch.

This also changes all our S3 uploads to be private, but I don't think there was any reason for these to be publicly readable. People outside MongoDB can't see Evergreen logs, so they'd have no way to find the relevant URLs anyway.

### DIFF
--- a/mongodump_passthrough/functions.yml
+++ b/mongodump_passthrough/functions.yml
@@ -67,8 +67,7 @@ variables:
   f_resmoke_with_binaries_fetch: &f_resmoke_with_binaries_fetch
     command: s3.get
     params:
-      aws_key: ${aws_key}
-      aws_secret: ${aws_secret}
+      role_arn: arn:aws:iam::391144487543:role/evergreen-project-mongo-tools
       # Changed for mongodump_passthrough
       remote_file: mongo-tools/mongodump_passthrough/${build_variant}/${revision}/resmoke-with-binaries/${build_id}.tgz
       bucket: mciuploads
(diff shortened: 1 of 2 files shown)


mongodump_passthrough/functions.yml
```diff
@@ -610,14 +604,14 @@ functions:
...
       params:
-        aws_key: ${aws_key}
-        aws_secret: ${aws_secret}
+        role_arn: arn:aws:iam::391144487543:role/evergreen-project-mongo-tools
         local_file: tools_output.tgz
         remote_file: mongo-tools/${build_variant}/${revision}/tools_output/${task_name}/${build_id}/tools_output.tgz
-        bucket: mciuploads
-        permissions: public-read
+        bucket: evergreen-project-mongo-tools-i6qg5nn6nbm
+        permissions: private
```
tdq45gj, line 614: Will the uploaded file be available for download from Evergreen UI?

tdq45gj: Seems that the default visibility is "public" according to the [doc](https://github.com/evergreen-ci/evergreen/blob/main/docs/Project-Configuration/Project-Commands.md), so it should be good.

tdq45gj (approved):
LGTM

mongodump_passthrough/functions.yml
autarch: Actually, I needed to add `visibility: signed` to make this work. The links showed up in Evergreen, but they resulted in permissions error without setting `visibility`.

PR #912, opened 2026-03-11
TOOLS-4120 Ignore the "cu" oplog type when restoring from an oplog dump

autarch:
This op type is being added as part of ongoing work related to the `recordIdsReplicated` feature.


tdq45gj (approved):
LGTM

PR #913, opened 2026-03-11
Update the script to create clusters to allow creating the default type of cluster too

autarch:
This makes it easy to quickly create two clusters in our dev env, one with `recordIdsReplicated` enabled and one without.

### DIFF
--- a/scripts/create-atlas-dev-cluster.sh
+++ b/scripts/create-atlas-dev-cluster.sh
@@ -29,10 +29,14 @@ GROUP_ID="673e58327d4f1a7610a14faf"
 NODE_COUNT="${3:-2}"
 API_CREDENTIALS="$ATLAS_PUBLIC_KEY:$ATLAS_PRIVATE_KEY"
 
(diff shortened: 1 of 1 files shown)


scripts/create-atlas-dev-cluster.sh
```diff
@@ -29,10 +29,14 @@ GROUP_ID="673e58327d4f1a7610a14faf"
 NODE_COUNT="${3:-2}"
 API_CREDENTIALS="$ATLAS_PUBLIC_KEY:$ATLAS_PRIVATE_KEY"
 
+if [ -n "$CLUSTER_TYPE" ]; then
```
nickweinberger, line 32: Would you need to pass "" for default cluster type?

nickweinberger (approved):
LGTM just the one question.

scripts/create-atlas-dev-cluster.sh
autarch: Yes. It's a bit clunky, but once we have automated testing for this with local clusters, this will all go away.

PR #922, opened 2026-03-19
TOOLS-4148 Annotate JS tests with SKIP, EXTEND, or NEW after auditing to figure out which ones need to be converted to Go

autarch:
Adds a // MIGRATION: comment to the first line of every JS test file under test/qa-tests/ and test/legacy42/ indicating its migration status:

- SKIP (27 files): genuinely covered by existing Go tests, broken pipe   tests (OS-signal-level, not testable in Go), or already disabled
- NEW (111 files): no Go test coverage exists for this scenario
- EXTEND (41 files): partial Go coverage exists but specific sub-scenarios   are missing

Each annotation includes the justification and the target Go test file.

### DIFF
--- a/test/qa-tests/jstests/bson/bsondump_broken_pipe.js
+++ b/test/qa-tests/jstests/bson/bsondump_broken_pipe.js
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — broken pipe is OS-signal-level, not testable in Go
 (function() {
   var bsondumpArgs = ['bsondump', '--failpoints', 'SlowBSONDump', '--type=json', 'jstests/bson/testdata/all_types.bson'];
   var ddArgs = ['dd', 'count=1000000', 'bs=1024', 'of=/dev/null'];
--- a/test/qa-tests/jstests/dump/dump_broken_pipe.js
+++ b/test/qa-tests/jstests/dump/dump_broken_pipe.js
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — broken pipe is OS-signal-level, not testable in Go
 (function() {
   if (typeof getToolTest === 'undefined') {
     load('jstests/configs/plain_28.config.js');
--- a/test/qa-tests/jstests/dump/version_test.js
+++ b/test/qa-tests/jstests/dump/version_test.js
@@ -1,3 +1,4 @@
+// MIGRATION: NEW — no Go test verifies that mongorestore with -vvv on an archive dump prints archive format version, server version, and tool version; goes in mongorestore/mongorestore_qa_test.go
 (function() {
   if (typeof getToolTest === 'undefined') {
     load('jstests/configs/plain_28.config.js');
--- a/test/qa-tests/jstests/export/export_broken_pipe.js
+++ b/test/qa-tests/jstests/export/export_broken_pipe.js
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — broken pipe is OS-signal-level, not testable in Go
 (function() {
   if (typeof getToolTest === 'undefined') {
     load('jstests/configs/plain_28.config.js');
--- a/test/qa-tests/jstests/files/mongofiles_version.js
+++ b/test/qa-tests/jstests/files/mongofiles_version.js
@@ -1,3 +1,4 @@
+// MIGRATION: NEW — no Go test verifies --version output; goes in common/options/options_test.go (covers all tools)
 // mongofiles_version.js; ensure that getting the version works without error
 var testName = 'mongofiles_version';
 load('jstests/files/util/mongofiles_common.js');
--- a/test/qa-tests/jstests/restore/different_collection.js
+++ b/test/qa-tests/jstests/restore/different_collection.js
@@ -1,3 +1,4 @@
+// MIGRATION: EXTEND — TestMongorestore covers nsFrom/nsTo same-DB rename but not: (1) cross-database restore with --db/--collection, (2) different-DB different-collection, (3) --db/--collection with a direct .bson file path; goes in mongorestore/mongorestore_qa_test.go
 (function() {
   if (typeof getToolTest === 'undefined') {
     load('jstests/configs/plain_28.config.js');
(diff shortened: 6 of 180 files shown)


test/qa-tests/jstests/export/export_broken_pipe.js
```diff
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — broken pipe is OS-signal-level, not testable in Go
```
FGasper, line 1: I disagree … ?

test/qa-tests/jstests/dump/version_test.js
```diff
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — standard CLI --version flag, not mongo-specific
```
FGasper: huh? The args given to mongodump here seem more interesting than this.

test/qa-tests/jstests/dump/dump_broken_pipe.js
```diff
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — broken pipe is OS-signal-level, not testable in Go
```
FGasper, line 1: can def be tested in Go

test/qa-tests/jstests/bson/bsondump_broken_pipe.js
```diff
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — broken pipe is OS-signal-level, not testable in Go
```
FGasper, line 1: testable in Go

test/qa-tests/jstests/files/mongofiles_version.js
```diff
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — standard --version flag, not mongo-specific
```
FGasper: This seems imprudently sensitive to implementation. It’s not a slow test or anything, so ISTM we might as well preserve this test.

test/qa-tests/jstests/restore/different_collection.js
```diff
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — rename collection on restore covered by mongorestore_test.go
```
FGasper: I’m a bit suspicious of these SKIPs. Just because something else tests renames doesn’t mean that that existing test flexes exactly the same code paths & does the same assertions.

test/qa-tests/jstests/stat/stat_broken_pipe.js
```diff
@@ -1,3 +1,4 @@
+// MIGRATION: SKIP — broken pipe is OS-signal-level, not testable in Go
```
FGasper, line 1: can, too

FGasper (changes requested):
- I think we should preserve the SIGPIPE tests.
- I’ll defer to you on whether existing Go tests do indeed obviate specific JS tests, but I’m suspicious.

test/qa-tests/jstests/bson/bsondump_broken_pipe.js
autarch: What exactly are we testing here? I did notice that we suppress `SIGPIPE` in the signal handling under `common`. The comment there says something about the tools checking the results of their write calls. This dates back to [https://jira.mongodb.org/browse/TOOLS-1366](https://jira.mongodb.org/browse/TOOLS-1366), which says that sometimes network write errors would manifest as a `SIGPIPE`.

I've never heard of this, so I asked both Glean and Gemini. Both seemed to confirm that with modern Go versions, the Go runtime suppresses this behavior except when writing to stdout/stderr, so I don't think we can get a `SIGPIPE` error because of a network issue in the tools code.

Apparently, the Go runtime will convert the `SIGPIPE` from the kernel into an error that it returns from the network read or write call. However, it still _does_ allow a `SIGPIPE` through when writing to stdout/stderr.

So given all this, I think we should revert the change to have the tools ignore the `SIGPIPE` signal. This doesn't achieve anything. And it _never_ made sense for things like `bsondump` which don't even talk to the network.

Here's what it looks like if I run `bsondump` with the code to ignore `SIGPIPE` commented out:

```
go run ./bsondump/main/ --type=json ./test/qa-tests/jstests/bson/testdata/all_types.bson | dd count=0 bs=1 of=/dev/null
0+0 records in
0+0 records out
0 bytes copied, 3.1766e-05 s, 0.0 kB/s
signal: broken pipe
```

And the current behavior, with the signal ignored:

```
go run ./bsondump/main/ --type=json ./test/qa-tests/jstests/bson/testdata/all_types.bson | dd count=0 bs=1 of=/dev/null
0+0 records in
0+0 records out
0 bytes copied, 2.3409e-05 s, 0.0 kB/s
2026-03-24T16:23:48.254-0500	0 objects found
2026-03-24T16:23:48.254-0500	write /dev/stdout: broken pipe
exit status 1
```

The first seems _more_ correct to me.

What do you think?

test/qa-tests/jstests/dump/version_test.js
autarch: Yeah, the AI got this wrong. I already had it revisit all its SKIP markings earlier, and I thought I caught all the bogus ones, but apparently not. I will push an update to the plan that includes this as one to be converted.

test/qa-tests/jstests/files/mongofiles_version.js
autarch: Yeah, the AI hallucinated this too. That said, I really question some of these tests. It runs `mongofiles --version` with three different cluster types to test that it prints its version?! And why do we only test this for `mongfiles` but not any other tools.

I'll tell the AI to add a test to the code in `common` that generates this output.

test/qa-tests/jstests/restore/different_collection.js
autarch: Yeah, some of this isn't tested by the existing code. I'll update the plan.

test/qa-tests/jstests/bson/bsondump_broken_pipe.js
FGasper: Given SIGPIPE’s history (a kludge for when folks “couldn’t be bothered” to check for failure), I actually favor ignoring it as long as we have linting in place that enforces error checking.

autarch: We do have `errcheck` enabled in this project's `golangci-lint` config.

I still don't see the point of leaving this as-is. But that's fine. I'll just add a single test in `common` for this behavior, since that's where the tools get this behavior.

autarch: Alright, see https://github.com/mongodb/mongo-tools/pull/956

PR #923, opened 2026-03-19
TOOLS-4148 Rewrite all the bsondump tests in Go and delete the JS tests for this

autarch:
A few of the JS tests were not ported, per the plan in docs/superpowers


bsondump/bsondump_test.go
```diff
@@ -260,6 +264,184 @@ func testFromFileWithPositionalArgumentToFile(t *testing.T) {
...
+		"minkey",
+		"null",
+		"oid",
+		"regex",
+		"string",
+		"symbol",
+		"ts",
+		"undefined",
+	} {
+		assert.Contains(t, result, key, "expected field %q in JSON output", key)
```
FGasper: Hrm? The original test does a full document-equality check; this just seems to check for field equality.

bsondump/bsondump_test.go
```diff
@@ -260,6 +264,184 @@ func testFromFileWithPositionalArgumentToFile(t *testing.T) {
...
+func TestBsondumpBadFiles(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	badFiles := []string{
+		"testdata/bad_cstring.bson",
+		"testdata/bad_type.bson",
+		"testdata/invalid_field_name.bson",
+		"testdata/partial_file.bson",
+		"testdata/random_bytes.bson",
+	}
```
FGasper, line 375: The refactoring of these tests makes it kind of hard to compare with the originals.

It might be nice if the AI privileged line-for-line translation rather than trying to refactor.

test/qa-tests/jstests/bson/bsondump_broken_pipe.js
```diff
@@ -1,19 +0,0 @@
-// MIGRATION: SKIP — broken pipe is OS-signal-level, not testable in Go
```
FGasper: ??? We can definitely test this in Go …

FGasper (changes requested):
Looks like the refactor reduced coverage in one test and imprudently skipped another.

bsondump/bsondump_test.go
autarch: It can do that, but do we really want that? If a human did this work I'm sure they'd do this same refactor. The way the JS test was written is just wrong. It should have used a loop!

test/qa-tests/jstests/bson/bsondump_broken_pipe.js
autarch: I think it said this because all the Go tests work with the Go code directly and construct the struct for the code in question directly, rather than compiling the program and testing it that way.

But what is the point of these tests? They seem to be testing bash, not the Go programs. I don't see anything to handle closed pipes in the Go code.

bsondump/bsondump_test.go
autarch: I had it rewrite this to test the full document.

FGasper (changes requested):
There are still unaddressed comments.

bsondump/bsondump_test.go
FGasper: I agree that the original test’s copy-paste-ness was excessive.

The refactor, though, makes it hard to ensure that the test was ported faithfully. I found one spot in this PR where Claude dropped something earlier; verifying that there isn’t something else is easier when the Go code resembles the JS.

test/qa-tests/jstests/bson/bsondump_broken_pipe.js
FGasper: It looks to me like this test intends to test `bsondump`’s behavior when STDOUT closes.

It’s a strange test, but that probably derives more from mongo shell limitations than the test author’s desire to use `bash`.

Since we have Go it should be easy to check the status code & verify that SIGPIPE was the “cause of death”.

bsondump/bsondump_test.go
FGasper: I’m not blocking on this, btw … but it might be a good parameter to give the AI in future translations.

FGasper (changes requested):
Can we have the AI create a SIGPIPE test for bsondump?

bsondump/bsondump_test.go
autarch: If you think the existing set of PRs are a lot to review, imagine effectively _doubling_ the number of PRs by breaking it up into "rewrite the JS test exactly into Go" and then "rewrite that into sane Go code". Is that a tradeoff we really want? I think the real problem with this PR is that I should've just done one test at a time.

test/qa-tests/jstests/bson/bsondump_broken_pipe.js
autarch: I talked about this in the other PR, but in fact this test is doing the opposite of checking that `SIGPIPE` was the cause of death. It's checking that we _ignore_ `SIGPIPE`, for historical reasons, I think.

autarch: This is now addressed in the previous PR in the stack.

FGasper (approved):
LGTM % the Evergreen failure.

Thank you!

PR #924, opened 2026-03-20
Add an `AGENTS.md` file for this repo

autarch:


AGENTS.md
```diff
@@ -0,0 +1,115 @@
...
+we should ensure that data is preserved via a dump and restore round trip.
+
+## Project Overview
+
+This repo contains the official MongoDB command-line tools: `bsondump`, `mongodump`, `mongorestore`,
+`mongoimport`, `mongoexport`, `mongostat`, `mongotop`, and `mongofiles`.
+
+Each tool lives in its own top-level directory (e.g., `/mongodump/`). Shared utilities are in
+`/common/`.
+
```
FGasper, line 28: Suggest adding that these tools are vendored in other (closed-source) tooling.

AGENTS.md
```diff
@@ -0,0 +1,115 @@
...
+## CI
+
+CI runs on Evergreen (config: `common.yml`). Do not edit `common.yml` by hand for test topology
+changes — consult the existing pattern for `integration-X.Y` and `integration-X.Y-sharded` task
+variants.
+
+## Minimal Change Principle
+
+Make minimal, incremental changes. Do not refactor surrounding code when fixing a bug or adding a
+feature. Do not add error handling for scenarios that cannot occur.
```
FGasper:
```suggestion
feature. Do not add error handling for scenarios that cannot occur; instead, add assertions around that behavior (e.g., via `panic` or `lo.Assert`).
```

FGasper (changes requested):
We should definitely mention the downstream tooling, IMO.

AGENTS.md
autarch: How would this affect the AI's decision making?

AGENTS.md
autarch: I updated this manually to just say `panic`. I don't think we should give it two options, and we don't use `lo.Assert` anywhere in this code base currently.

AGENTS.md
autarch: FWIW, I don't really object to adding something here, I just don't know exactly what you want added or what you want the AI to do given this addition.

FGasper: @autarch Did you see this?

FGasper (changes requested):

AGENTS.md
autarch: I had commented but it was pending submitting a review for some reason.

FGasper: > what you want the AI to do given this addition

The AI should take care not to break exported interfaces since they could be used downstream.

FGasper (approved):
LGTM % the bit about telling the AI not to break exported interfaces.

PR #925, opened 2026-03-20
TOOLS-4148 Convert `test/qa-tests/jstests/export/basic_data.js` to Go

autarch:
This PR adds a new  `TestRoundTripBasicData` test to `mongoimport/mongoimport_test.go`. The test inserts 50 docs, exports them as JSON, and then imports them to test a basic round trip scenario.


mongoimport/mongoimport_test.go
```diff
@@ -1533,3 +1537,76 @@ func TestImportMIOSOE(t *testing.T) {
...
+	require.NoError(t, err)
+	imported, _, err := mi.ImportDocuments()
+	require.NoError(t, err)
+	assert.EqualValues(t, 50, imported, "should import all 50 documents")
+
+	count, err := coll.CountDocuments(t.Context(), bson.D{})
+	require.NoError(t, err)
+	assert.EqualValues(t, 50, count, "collection should have all 50 documents after round-trip")
+
+	for i := range int32(50) {
```
mmcclimon: Your robot is obsessed with `int32`s, man.

mmcclimon (commented):
I also left this comment like 5 minutes ago on the DSC testing thing, but: I think it would be useful to have a conversation about where to put / how to do these round-trip tests. I recently added (at your prompting) an `integration/importexport` directory to put these kinds of tests, because I feel like it's sort of weird to put them in either `mongoimport/` or `mongoexport/`. 

Like this test is fine, and we can always move them later if we want, but it would also be nice if we didn't have to and could just rewrite them all _once_. The tests I wrote also used the testify suite, which I prefer to passing `t` around everywhere, and I'd imagined that the other e2e-style tests would do similar, and we probably don't want to rewrite them all _twice_ (once to Go, once to suite-style). We don't _have_ to use the suites, of course, but we do in mongosync and I would love to have one way to write tests instead of 7.

mongoimport/mongoimport_test.go
autarch: I _think_ that's because the existing test code is constantly doing this already, and it's trying to match the existing style. Why does the existing code do this? I dunno :shrug:

autarch: Fixed

mmcclimon (approved):
> I actually was thinking of moving them after doing a bunch of these test conversions. At this point, I'd like to just merge these as-is and do some cleanup after. I also agree that moving to using a testify suite would be good. But I'd also like to defer that until after mergint his mega-stack.

Cool, sounds good to me; I just wanted to make sure we were vaguely aligned on the direction we want to shove on these.

Change LGTM, thanks!

mongoimport/mongoimport_test.go
mmcclimon: I was wondering if it was like, "well, numbers in JS are all floats, so we might not always be able to fit a 64-bit int" or something.

PR #933, opened 2026-03-23
TOOLS-4148 Delete `test/qa-tests/jstests/export/force_table_scan.js` without converting

autarch:
The `--forceTableScan` option was deprecated when WiredTiger became the default in MongoDB 3.0 and mmapv1 was removed in 4.2. On 4.2+ the flag is a no-op and the profiling assertions in the JS test are trivially true.


mmcclimon (approved):
LGTM, thanks.

Also, a meta-comment: is there a ticket number for this work? It would be good to make one, even a catch-all, because (I think?) we use JIRA to generate the release changelogs.

PR #935, opened 2026-03-23
TOOLS-4148 Convert `test/qa-tests/jstests/export/limit.js` to Go

autarch:


mongoimport/mongoimport_test.go
```diff
@@ -2201,3 +2201,74 @@ func TestRoundTripJSONArray(t *testing.T) {
 		assert.EqualValues(t, 1, c, "document with _id %d should exist", i)
 	}
 }
+
+// TestRoundTripLimit verifies that mongoexport --limit restricts the number of
+// exported documents, and that the correct documents are restored (from limit.js).
```
tdq45gj: This comments still references limit.js, but it's removed. I think we should either clarify where to find the legacy file or just remove the reference.

autarch: Removed

tdq45gj (approved):
LGTM

PR #941, opened 2026-03-23
TOOLS-4148 Delete `test/qa-tests/jstests/export/slave_ok.js` without converting

autarch:
The `--slaveOk` flag is hidden and deprecated. It's the equivalent of setting ` readPreference=nearest` in the connection string.


mmcclimon (approved):
LGTM, thanks. (Also: oof.)

You probably don't know the answer to this, but: are we ever going to be able to delete all these deprecated options?

PR #942, opened 2026-03-23
TOOLS-4148 Convert `test/qa-tests/jstests/export/sort_and_skip.js` to Go

autarch:


mongoimport/mongoimport_test.go
```diff
@@ -2481,3 +2481,73 @@ func exportAndImportWithQuery(
...
+		if err := client.Database(dbName).Drop(context.Background()); err != nil {
+			t.Errorf("dropping test database: %v", err)
+		}
+	})
+
+	coll := client.Database(dbName).Collection(collName)
+	docs := make([]any, 50)
+	for i := range 50 {
+		docs[i] = bson.D{{"a", int32(i)}}
+	}
```
nickweinberger, line 2506: I think this kinda defeats the purpose of testing sort behavior right? The js test inserted in scrambled order, whereas this inserts in sorted order.

nickweinberger (changes requested):

mongoimport/mongoimport_test.go
autarch: well, actuallly ... the AI pointed out that WiredTiger doesn't guarantee an order for an unsorted query ... but I added a shuffle.

nickweinberger (approved):
LGTM thanks.

PR #946, opened 2026-03-23
TOOLS-4148 Remove testdata for export JS tests

autarch:
We're not converting the broken pipe test, and this test data is no longer needed now that all the other tests have been converted.


mmcclimon (commented):
I don't know that it's right to delete this test. It's not possible to test entirely in Go but do we really want to delete the test for this behavior entirely?

mmcclimon (approved):
LGTM % the title/commit message is wrong now since this only removes testdata; thanks!

PR #956, opened 2026-03-25
TOOLS-4148 Add tests for SIGPIPE handling, remove equivalent JS tests, and fix handling for other signals

autarch:
This test makes sure that we ignore SIGPIPE. Instead, writes to a broken pipe will cause the write call to return an error. The tools should always check errors on write calls, so these will be surfaced as normal "failure to write to an fd" errors, as opposed to the signal aborting the process.

This also fixes a bug in our handling of other channels that the AI found. In Go, we should always use a buffered channel for signal handling, or we can miss the signal entirely. See https://pkg.go.dev/os/signal for more on this.


FGasper (changes requested):
This replaces end-to-end tests of the actual applications with a unit test of a specific function. It’s useful but doesn’t retain the original test’s scope.

How feasible is it to build the application, pipe its stdout to a process that ends right away, and verify that the application exits nonzero but doesn’t die from SIGPIPE? (Probably also keep the check for the string `broken pipe`.)

common/signals/signals.go
```diff
@@ -21,24 +21,23 @@ func Handle() chan struct{} {
...
-// SIGKILL and explicitly ignores SIGPIPE. It calls the finalizer function when
-// the first signal is received and forcibly terminates the program after the
-// second. If a nil function is provided, the program will exit after the first
-// signal.
+// HandleWithInterrupt starts a goroutine which listens for SIGTERM, SIGINT, and SIGKILL. It also
+// calles signal.Ignore to explicitly ignore SIGPIPE. It calls the finalizer function when the first
+// signal is received and forcibly terminates the program after the second. If a nil function is
+// provided, the program will exit after the first signal.
 func HandleWithInterrupt(finalizer func()) chan struct{} {
+	// Ignore SIGPIPE synchronously so there is no race between this call and the first write: a
```
FGasper, line 29: Should the race-prevention logic here also be applied to the handling of TERM & INT?

FGasper: (Aside: I’m a bit surprised this compiles on Windows …)

common/testutil/broken_pipe_unix.go
```diff
@@ -0,0 +1,63 @@
...
+	require.NoError(t, err, "process should write at least some output before the pipe breaks")
+	require.NoError(t, pr.Close())
+
+	err = cmd.Wait()
+	if err == nil {
+		return
+	}
+
+	var exitErr *exec.ExitError
+	if errors.As(err, &exitErr) {
```
FGasper: This seems to accept exiting without error. Can we not ensure that the process _does_ exit in error? Or is that something the caller can do?

common/testutil/broken_pipe_unix.go
```diff
@@ -0,0 +1,63 @@
...
+
+		if status.Signaled() {
+			require.NotEqual(
+				t,
+				syscall.SIGPIPE,
+				status.Signal(),
+				"process should not be killed by SIGPIPE: broken pipe should surface as a write error",
+			)
+		}
+	}
```
FGasper, line 59: Can we retain the check for `broken pipe` in the program’s stdout/stderr? That’ll ensure that the pipe is what caused the failure, and that we handled it cleanly.

common/testutil/broken_pipe_unix.go
```diff
@@ -0,0 +1,63 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// AssertBrokenPipeHandled starts cmd with its stdout connected to a pipe,
+// reads a small amount of output to confirm the process has started writing,
+// then breaks the pipe by closing the read end. It asserts the process was
+// not killed by SIGPIPE — a broken pipe must surface as a write error that
+// the process can handle cleanly.
```
FGasper, line 24: This description seems a _bit_ misleading with the function written as it is because—to me, at least—it implies that we expect a failure due to write error (albeit one that the program handles cleanly).

(I can also read it as _not_ implying that, but it’s not the first impression I got when reading this paragraph.)

FGasper (approved):
I still think this regresses coverage a bit (e.g., not checking for `broken pipe` in the output), but I also think more back & forth here isn’t warranted.

I’ve made a few comments; handle them as you will.

Thank you for this, though. It definitely simplifies this project’s maintenance.

common/signals/signals.go
autarch: That's a good question re: race. I have no idea why the code is written this way. I _think_ that it's not really a big deal. The worst case is that we get a TERM or INT signal before our handlers are installed, but I don't think that would cause any breakage.

autarch: As for Windows, I think the Go stdlib & runtime handles the platform-specific details to make this work.

common/testutil/broken_pipe_unix.go
autarch: Yes, we expect the command we run to fail due to a write error. That's what the code looks for. It checks that the process exited but _not_ via a `SIGPIPE`.

I wondered why we have the conditional on whether we got an exit error, but I think that makes sense. It's possible that a tool could handle a write error and exit 0. Maybe they _shouldn't_ do that, but that's somewhat out of scope for these tests, so I think having the test pass with exit 0 is reasonable.

common/testutil/broken_pipe_unix.go
autarch: See above.

common/testutil/broken_pipe_unix.go
autarch: I'm not sure we should. In the JS code, this had been commented out for `mongodump` with this comment: TODO: TOOLS-1883 Following tests are flaky.


I don't want to try to figure that out right now, especially since I think this is all pointless and we don't need to suppress SIGPIPE any more (this is a legacy of _very_ early Go versions being weird with `SIGPIPE`).

PR #960, opened 2026-03-26
TOOLS-4148 Convert `test/qa-tests/jstests/import/mongoimport mode.js` to Go

autarch:
Adds TestImportModes covering all --mode / --upsertFields combinations
from jstests/import/mode.js.

Notable implementation details:
- Table-driven with importModeCase struct; wantErrContains checks
  specific error substrings via require.ErrorContains rather than just
  require.Error
- runImportOpts helper returns errors from New() (unlike
  importWithIngestOpts which uses require.NoError), enabling error-path
  test cases for invalid option combinations
- writeJSONLinesFile marshals []map[string]any to newline-separated JSON
  instead of hard-coding string literals
- map[string]any decode results must be reset to map[string]any{} between
  FindOne/Decode calls; stale keys from a previous decode persist
  otherwise (bug found via test failure)


mmcclimon (commented):
I dunno man; this test was pretty difficult to follow before and I think the translation to Go has made that even worse. This whole file is a maze of if statements and magic strings, and I find it really difficult to tell what's _actually_ being tested here.

To take a concrete example: in `checkByIDImport`, there's an assert that for `_id: "one"`, the `b` field should be `"zebras"`. To figure out why, you need to go back to find the caller, determine that it's either `checkByIDUpsert` or `checkByIDMerge`, find the entries in test table that have either of those in the `check` field, see that they are using the `upsert1File` (that actually contains the string "zebras"), then follow the code back out through `runImportOpts` to see what in the world the test is actually _doing_.

mongoimport/mongoimport_test.go
```diff
@@ -3158,3 +3158,274 @@ func importWithIngestOpts(
...
+		{"deprecated --upsert", IngestOptions{Upsert: true, UpsertFields: "a,c"}},
+		{"mode=upsert", IngestOptions{Mode: modeUpsert, UpsertFields: "a,c"}},
+	} {
+		t.Run(tc.name+" replaces matched docs", func(t *testing.T) {
+			setupUpsertFieldsDocs(t, coll)
+			require.NoError(t, runImportOpts(t, ns, importFile, tc.opts))
+
+			var doc map[string]any
+			require.NoError(t, coll.FindOne(t.Context(), bson.D{{"b", "blah"}}).Decode(&doc))
+			assert.Nil(t, doc["x"], "upsert replaces doc; original x field gone")
```
mmcclimon, line 3239: I think this version is better but it's still pretty hard to follow, imo. In particular, when I get to this line I was confused because I was like "wait, there wasn't an `x` field at all in the docs above!" That's because the `x` field is _actually_ inserted in `setupUpsertFieldsDocs`, which is 150 lines further down in this file. 

I think both of these would be easier to read if the original docs were defined in the same place as the upsert lines, like:

```go
docs := []any{
    bson.D{...},
    ...,
}

importFile := writeJSONLines([]map[string]any{
    {...},
    ....
}

// later on
setupDocs(t, coll, docs)
```

(this also means you don't need two separate methods for `setupUpsertFieldsDocs` and `setupImportDocs`, because they're exactly the same except for the magic docs they insert.)

mmcclimon (changes requested):
This is better than it was but I think still confusing; thanks.

mongoimport/mongoimport_test.go
autarch: Ok, I did some more refactoring. Now the docs are defined right above the subtests, which should make this a bit easier to understand.

mmcclimon (approved):
LGTM, thanks for the tweaks here!

PR #963, opened 2026-03-26
Update every embedded script in the Evergreen config to start with the same `set` directives

autarch:
Now most scripts start with:

```
set -o errexit
set -o pipefail
set -o verbose
```

There were a few spots where error checking needs to be off because the commands we run may fail, but failure is okay.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -63,11 +63,13 @@ functions:
     command: shell.exec
     type: test
     params:
+      shell: bash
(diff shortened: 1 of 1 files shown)


mmcclimon (changes requested):
I'm fine with this in theory but this not actually seem to work, according to Evergreen.

mmcclimon (approved):
LGTM, thanks!

PR #966, opened 2026-03-27
Fix evergreen-validate failures in CI

autarch:
This also rewrites the code that ignores certain warnings to be a bit clearer.

### DIFF
--- a/buildscript/sa.go
+++ b/buildscript/sa.go
@@ -403,16 +403,21 @@ func SAEvergreenValidate(ctx *task.Context) error {
 		return fmt.Errorf("error from `evergreen validate`: %s: %w", output, err)
 	}
 
-	// TODO: change this if-block in TOOLS-2840.
-	// This check ignores any YAML warnings related to duplicate keys in YAML maps.
-	// See ticket for more details.
 	if strings.HasSuffix(output, "is valid with warnings") {
 		for _, line := range strings.Split(output, "\n") {
-			if !strings.HasSuffix(line, "unmarshal errors:") &&
-				!strings.HasSuffix(line, "already set in map") &&
-				!strings.HasSuffix(line, "is valid with warnings") {
-				return fmt.Errorf("error from `evergreen validate`: %s", output)
+			if strings.HasPrefix(line, "WARNING: ") &&
+				strings.HasSuffix(
+					line,
+					"defined but not used by any variants; consider using or disabling",
+				) {
+				continue
 			}
+
+			if strings.HasSuffix(line, "is valid with warnings") {
+				continue
+			}
+
+			return fmt.Errorf("error from `evergreen validate`: %s", output)
 		}
 	}
 
--- a/mongodump_passthrough/tasks.yml
+++ b/mongodump_passthrough/tasks.yml
@@ -70,7 +70,7 @@ tasks:
         params:
           files:
             - src/github.com/mongodb/mongo-tools/manual-tasks.json
-    patch_only: true
+    allowed_requesters: ["patch"]
 
   # Task dependencies have been changed slightly here, to allow this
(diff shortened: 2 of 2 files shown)


buildscript/sa.go
```diff
@@ -403,16 +403,26 @@ func SAEvergreenValidate(ctx *task.Context) error {
...
-				!strings.HasSuffix(line, "already set in map") &&
-				!strings.HasSuffix(line, "is valid with warnings") {
-				return fmt.Errorf("error from `evergreen validate`: %s", output)
+			if strings.HasPrefix(line, "WARNING: ") {
+				if strings.HasSuffix(line, "unmarshal errors:") ||
+					strings.HasSuffix(line, "already set in map") ||
+					strings.HasSuffix(line, "but allowed_requesters is always higher precedence") ||
+					strings.HasSuffix(
+						line,
+						"defined but not used by any variants; consider using or disabling",
```
autarch: The real change is adding these two things to the ignore list and removing the "already set in map" check, because that is no longer a warning we see.

buildscript/sa.go
```diff
@@ -403,16 +403,26 @@ func SAEvergreenValidate(ctx *task.Context) error {
...
 	if strings.HasSuffix(output, "is valid with warnings") {
 		for _, line := range strings.Split(output, "\n") {
-			if !strings.HasSuffix(line, "unmarshal errors:") &&
-				!strings.HasSuffix(line, "already set in map") &&
-				!strings.HasSuffix(line, "is valid with warnings") {
-				return fmt.Errorf("error from `evergreen validate`: %s", output)
+			if strings.HasPrefix(line, "WARNING: ") {
+				if strings.HasSuffix(line, "unmarshal errors:") ||
+					strings.HasSuffix(line, "already set in map") ||
+					strings.HasSuffix(line, "but allowed_requesters is always higher precedence") ||
```
mmcclimon: I think a better fix here would just be to fix the task definition (I meant to do this the other day and then got sidetracked and forgot). The fix is just in the `run_manual_generated_tasks` task:

```diff
diff --git a/mongodump_passthrough/tasks.yml b/mongodump_passthrough/tasks.yml
index e6ea4fde..722f258a 100644
--- a/mongodump_passthrough/tasks.yml
+++ b/mongodump_passthrough/tasks.yml
@@ -70,7 +70,7 @@ tasks:
         params:
           files:
             - src/github.com/mongodb/mongo-tools/manual-tasks.json
-    patch_only: true
+    allowed_requesters: ["patch"]

   # Task dependencies have been changed slightly here, to allow this
   # task to execute concurrently with "compile_coverage".
```

(And then with that, we don't need this suffix check.)

buildscript/sa.go
mmcclimon: [narrator] But he did not, in fact, remove the "already set in map" check.

mmcclimon (approved):
I would tweak a couple things here but trust you to make these changes; I don't need to review this again unless you just really want me to, thanks!

buildscript/sa.go
autarch: Now I did. I also removed most of the other checks, because we're no longer getting them. It's just the unused variants one.

buildscript/sa.go
autarch: Fixed.

autarch:
### Merge activity

* **Mar 27, 9:14 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/mongodb/mongo-tools/966).
* **Mar 27, 9:15 PM UTC**: @autarch merged this pull request with [Graphite](https://app.graphite.com/github/pr/mongodb/mongo-tools/966).

PR #967, opened 2026-03-27
TOOLS-4154 Fix handling empty `partialFilterExpression` option 

tdq45gj:
This commit makes the `PartialFilterExpression` field a pointer so that it's not omitted when set to a valid empty document.


mmcclimon (approved):
LGTM, thank you! (And nice work on the diagnosis + fix.)

Do we also want/need a test on the mongodump side to make sure we dump these correctly? I suspect probably we _do_, if it was failing in production, but it might be good to double-check.

PR #969, opened 2026-03-30
Fix the `evergreen validate` check after changes to `evergreen` CLI output

autarch:


mmcclimon (approved):
This is fine as it goes, so I will approve, but also: I don't see this change locally, and allegedly I am up to date: 

```
[..ongo/mongo-tools on main] $ evergreen update
[message='Binary is already up to date - not updating.' revision='2026-03-28']

[..ongo/mongo-tools on main] $ evergreen validate common.yml
WARNING: task 'commit-queue-workaround' defined but not used by any variants; consider using or disabling
WARNING: task 'deliberate_fail_to_test_build_baron_context_rules' defined but not used by any variants; consider using or disabling
WARNING: task 't_resmoke_setup' defined but not used by any variants; consider using or disabling
WARNING: task 'generate_mongodump_fuzz_tasks' defined but not used by any variants; consider using or disabling
common.yml is valid with warnings
```

PR #970, opened 2026-03-30
Add a name to our `package.json` file

autarch:
If this isn't set, then every time we run `npm`, it updates the name in the `package-lock.json` file to match the current directory. I discovered this when working in a second copy of the tools repo in a directory named `mongo-tools2`.


tdq45gj (approved):
LGTM

PR #972, opened 2026-03-30
Move all existing round trip Go tests to the `integration` dir

autarch:
The test code has been moved as verbatim as possible. There will be some cleanups in follow-up PRs. The only changes made in this PR are those that were necessary for linting to pass, as well as adding a few `t.Cleanup` calls to make sure that test data from one tests didn't break another test (this was causing CI failures).

The tests were moved as follows:

mongoimport/mongoimport_test.go → integration/importexport/:

- csv_test.go
    - TestRoundTripFieldFile
    - TestRoundTripFieldsCSV
    - TestRoundTripNestedFieldsCSV
- data_test.go
    - TestRoundTripBasicData
    - TestRoundTripDataTypes
    - TestRoundTripDecimal128
    - TestRoundTripViewExport
- import_validation_test.go
    - TestImportDocumentValidation
- import_modes_test.go
    - TestImportModeUpsertIDSubdoc
- json_test.go
    - TestRoundTripFieldsJSON
    - TestRoundTripJSONArray
- query_test.go
    - TestRoundTripLimit
    - TestRoundTripQuery
    - TestRoundTripSortAndSkip

mongorestore/mongorestore_test.go → integration/dumprestore/roundtrip_test.go:

- TestPipedDumpRestore
- TestDumpAndRestoreConfigDB
- TestRestoreZeroTimestamp
- TestRestoreZeroTimestamp_NonClobber
- TestRestoreMultipleIDIndexes


mongorestore/oplog_test.go
```diff
@@ -896,6 +899,7 @@ func TestOplogRestoreCollModTTLIndex(t *testing.T) {
 
 	// Prepare the test by creating the necessary collection.
 	require.NoError(t, session.Database("mongodump_test_db").Drop(ctx))
+	t.Cleanup(func() { _ = session.Database("mongodump_test_db").Drop(t.Context()) })
```
mmcclimon, line 902: For some reason I thought I had tried to do this in the past (drop things with t.Cleanup using t.Context) and it didn't work because t.Context() is canceled before Cleanup runs. That might still actually be the case here, but we're ignoring the errors so maybe we don't notice?

Most of the other code you've added recently has used context.Background() in the Cleanup functions, I assume for this reason. I have also written some code that uses normal deferred funcs with t.Context. I don't know that ultimately it matters a whole lot but it would be nice if we could move toward the same way of doing this.

(And maybe what we _actually_ want is to move toward suites where we have a BeforeTest method that drops everything so just don't have to worry about it at all.)

mmcclimon (approved):
LGTM; I didn't read this carefully because I'm trusting you that it's just moving things between files. I left one comment but it's not blocking here, just something to think about going forward.

mongorestore/oplog_test.go
autarch: I forgot to address this in this PR. I'll make a follow-up PR to add a `BeforeTest` that drops everything. That's much more future proof.

PR #974, opened 2026-03-31
TOOLS-4158 Properly escape DB and collection names in tool output

autarch:
Previously, we used `%v` or `%s` to print these names. But they can contain things like newlines or tabs, which make a huge mess when printed. This changes the code to `%#q` when printing these names.

Along the way, it also changes how we print other strings like filenames and such to use this as well.

This only covers mongodump, mongorestore, mongoexport, and mongoimport for now. I will make some follow-up PRs with more escaping.

### DIFF
--- a/mongoimport/mongoimport.go
+++ b/mongoimport/mongoimport.go
@@ -241,7 +241,7 @@ func (imp *MongoImport) validateSettings() error {
 
 	if imp.IngestOptions.Mode != modeInsert {
 		imp.IngestOptions.MaintainInsertionOrder = true
-		log.Logvf(log.Info, "using upsert fields: %v", imp.upsertFields)
+		log.Logvf(log.Info, "using upsert fields: %v", util.QuoteAndJoin(imp.upsertFields, ","))
 	}
 
 	if imp.IngestOptions.MaintainInsertionOrder {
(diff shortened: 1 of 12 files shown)


mongoimport/mongoimport.go
```diff
@@ -372,9 +372,7 @@ func (imp *MongoImport) importDocuments(inputReader InputReader) (uint64, uint64
 		util.SanitizeURI(imp.ToolOptions.ConnectionString),
 	)
 
-	log.Logvf(log.Info, "ns: %v.%v",
-		imp.ToolOptions.DB,
-		imp.ToolOptions.Collection)
+	log.Logvf(log.Info, "ns: %#q", imp.ToolOptions.DB+"."+imp.ToolOptions.Collection)
```
mankawal, line 375: I'm curious about the reason behind moving the concatenation to arguments instead of using the string formatter (as in the original code) was intentional.

mankawal (approved):

mongoimport/mongoimport.go
autarch: Yes, if we did `%q.%q` we'd end up with \`foo\`.\`bar\` in the output. This way we get \`foo.bar\`

PR #975, opened 2026-03-31
TOOLS-4158 Use `%#q` for all strings and runes in output

autarch:
There were many places that could benefit from escaping, not just for DB and collection names


nickweinberger (approved):
LGTM

PR #981, opened 2026-04-03
fix: increase OP_MSG overhead margin in BufferedBulkInserter

jslay-webflow:
`BufferedBulkInserter` subtracts 100 bytes from `maxMessageSizeBytes` (48000000) to leave room for the OP_MSG wire overhead. this margin is too small.

we hit this doing mongorestore on a ~27GB mongo 7.0 instance. two collections consistently produce wire messages of exactly 48000149 bytes -- 149 bytes over the limit. the server rejects them:

```
recv(): message msgLen 48000149 is invalid. Min 16 Max: 48000000
```

the `byteCount` in `addModel` tracks raw BSON document sizes, but the actual wire message includes the OP_MSG header (16 bytes), flagBits (4), the insert command body with collection name and options, and document sequence section headers. for collections with longer namespaces or write concern options, this overhead easily exceeds 100 bytes.

bumped the margin from 100 to 1024 bytes. this is conservative but still allows batches very close to the theoretical max. the alternative is tracking the exact overhead per-message, but that would require awareness of the collection namespace length and write concern at the `BufferedBulkInserter` level, which seems like overkill for a safety margin.

found this while setting up nightly mongodump/mongorestore backups for a dev cluster. both archive-format and directory-format restores failed reliably on certain collections with mongo 7.0 tools. mongo 5.0 tools worked against the same dump, presumably because of smaller default batch sizes that never got close enough to the limit. the 100-byte margin has probably been insufficient for a while but only surfaces when you have enough large documents to fill a batch right up to the limit.

### how to reproduce

any collection at a namespace like `testdb.large_documents` (total namespace > ~11 chars) with enough ~48KB documents to fill a batch to the byte limit will trigger this. the OP_MSG overhead for that namespace is ~107 bytes, exceeding the 100-byte margin.

### tests

added unit tests that verify the 1024-byte margin covers the OP_MSG overhead for namespaces of various lengths, and that the old 100-byte margin was insufficient for realistic namespaces.

### DIFF
--- a/common/db/buffered_bulk.go
+++ b/common/db/buffered_bulk.go
@@ -75,10 +75,12 @@ func newBufferedBulkInserter(
 		collection:    collection,
 		bulkWriteOpts: bulkOpts,
 		docLimit:      docLimit,
-		// We set the byte limit to be slightly lower than maxMessageSizeBytes so it can fit in one OP_MSG.
-		// This may not always be perfect, e.g. we don't count update selectors in byte totals, but it should
-		// be good enough to keep memory consumption in check.
-		byteLimit:          MAX_MESSAGE_SIZE_BYTES - 100,
+		// Reserve space for the OP_MSG header (16 bytes), flagBits (4 bytes),
+		// command body section (insert command + collection name + options ~200 bytes),
+		// and document sequence section header (4 + 4 + identifier bytes).
+		// The previous margin of 100 bytes was insufficient and caused
+		// "message msgLen is invalid" errors on collections with large documents.
+		byteLimit:          MAX_MESSAGE_SIZE_BYTES - 1024,
 		writeModels:        make([]mongo.WriteModel, 0, docLimit),
 		canDoZeroTimestamp: zeroTimestampOk,
 	}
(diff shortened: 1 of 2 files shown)


mmcclimon:
Hi @jslay-webflow, thanks again for this! Before I see about merging it, can you confirm what version of the tools you were testing with? I ask because we released 100.16.0 last week, which had a fix for TOOLS-4145 in it; I wonder if that fixes the problem you were seeing.

mmcclimon (commented):
(Waiting to do anything more on this because I think this will have been fixed by #952.)

tdq45gj:
@jslay-webflow Thanks for the PR! I think the size of the overhead depends on length of the namespace (255 bytes at max) and number of documents in the buffer. In case there are many documents in the buffer, a margin of 1024 bytes might not be enough. I would suggest increasing the margin more aggressively to 1MB since it's not easy to predict the message size in the tools.

jslay-webflow:
Will get back to you guys on this on Monday.

tdq45gj:
@jslay-webflow Thanks again for this PR. I made some changes based on your branch and merged it in #992. I'm closing this one now.

jslay-webflow:
Thanks. Sorry it has taken me so long to get back to this.

PR #989, opened 2026-04-07
TOOLS-4142 Estimate 1 MB for message overhead

tdq45gj:
A previous fix to cap the total size of inserted documents to 48MB does not seem to fix the issue for some users. This commit adds another fix in the `BufferedBulkInserter` to always reserve 1 MB for message overhead.


mmcclimon:
@tdq45gj See also #981, from an external person, which is a similar fix (but much smaller than 1MB).

tdq45gj:
Thanks for the heads up. I posted a comment in their PR and converted this one to draft for now.

PR #991, opened 2026-04-16
Increase OP_MSG overhead margin from 100 bytes to 4 MiB

github-benoit-dav:
## Summary

- Increase the byte limit margin from 100 bytes to 4 MiB (`MSG_OVERHEAD_BYTES`) to account for wire message overhead not tracked by `BufferedBulkInserter.byteCount`
- The Go driver v2 `Batches.AppendBatchSequence` checks only cumulative document size against `MaxMessageSize` (48 MB) without subtracting the already-written command header, session metadata, and framing — so the tool-side limit must reserve enough room
- The previous 100-byte margin was sufficient with driver v1 (which split batches at 16 MB), but caused `ProtocolError: msgLen is invalid` after the driver v2 upgrade (TOOLS-3968)

## Test plan

- [ ] Restore a backup with collections large enough to produce batches near the 48 MB limit — previously failing restores should now succeed
- [ ] Verify no regression with small backups that fit in a single batch
- [ ] Run existing `TestBufferedBulkInserterInserts` integration tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/common/db/buffered_bulk.go
+++ b/common/db/buffered_bulk.go
@@ -16,9 +16,18 @@ import (
 	"go.mongodb.org/mongo-driver/v2/x/mongo/driver/xoptions"
 )
 
-// The default value of maxMessageSizeBytes
+// MAX_MESSAGE_SIZE_BYTES default value of maxMessageSizeBytes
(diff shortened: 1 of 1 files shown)


autarch:
Hi @github-benoit-dav, thanks for this PR. We already have a PR in flight for this issue, #992. So I'm closing this one.

PR #992, opened 2026-04-16
Fix OP_MSG size overflow in BufferedBulkInserter

tdq45gj:
This PR is based on PR #981, with the overhead buffer size in BufferedBulkInserter increased to 1MB.

### DIFF
--- a/common/db/buffered_bulk.go
+++ b/common/db/buffered_bulk.go
@@ -75,10 +75,12 @@ func newBufferedBulkInserter(
 		collection:    collection,
 		bulkWriteOpts: bulkOpts,
 		docLimit:      docLimit,
-		// We set the byte limit to be slightly lower than maxMessageSizeBytes so it can fit in one OP_MSG.
-		// This may not always be perfect, e.g. we don't count update selectors in byte totals, but it should
-		// be good enough to keep memory consumption in check.
-		byteLimit:          MAX_MESSAGE_SIZE_BYTES - 100,
+		// Reserve space for the OP_MSG header (16 bytes), flagBits (4 bytes),
+		// command body section (insert command + collection name + options ~200 bytes),
+		// and document sequence section header (4 + 4 + identifier bytes).
+		// The previous margin of 100 bytes was insufficient and caused
+		// "message msgLen is invalid" errors on collections with large documents.
+		byteLimit:          MAX_MESSAGE_SIZE_BYTES - 1024,
 		writeModels:        make([]mongo.WriteModel, 0, docLimit),
 		canDoZeroTimestamp: zeroTimestampOk,
 	}
--- a/common/db/buffered_bulk.go
+++ b/common/db/buffered_bulk.go
@@ -76,6 +76,8 @@ func newBufferedBulkInserter(
 		bulkWriteOpts: bulkOpts,
 		docLimit:      docLimit,
 		// Leave room for OP_MSG wire overhead (header, command body, namespace, etc).
+		// The overhead depends on length of the namespace (max 255 bytes) and the number of
+		// documents. The estimation is set to 1MB to be conservative.
 		byteLimit:          MAX_MESSAGE_SIZE_BYTES - 1024*1024,
 		writeModels:        make([]mongo.WriteModel, 0, docLimit),
 		canDoZeroTimestamp: zeroTimestampOk,
--- a/common/db/buffered_bulk.go
+++ b/common/db/buffered_bulk.go
@@ -76,10 +76,7 @@ func newBufferedBulkInserter(
 		bulkWriteOpts: bulkOpts,
 		docLimit:      docLimit,
 		// Leave room for OP_MSG wire overhead (header, command body, namespace, etc).
-		// The previous margin of 100 was too small -- namespaces longer than ~11 chars
-		// total pushed the actual message past maxMessageSizeBytes. 1024 is conservative
-		// but safe for any realistic namespace.
-		byteLimit:          MAX_MESSAGE_SIZE_BYTES - 1024,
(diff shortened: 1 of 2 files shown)


utafrali (commented):
The core fix is correct and addresses a real production bug where the 100-byte OP_MSG overhead margin was insufficient for most namespace lengths. However, the new tests have a stale bound (`1024` instead of `1024*1024`) in `TestOldByteLimitMarginWasInsufficient`, hardcode the margin value rather than referencing the production constant, and never actually instantiate a `BufferedBulkInserter` to verify the field is set correctly.

common/db/buffered_bulk_test.go
```diff
@@ -16,6 +16,54 @@ import (
...
+// bytes of overhead, which is why mongorestore would fail with
+// "message msgLen is invalid" on collections with many large documents.
+func TestOldByteLimitMarginWasInsufficient(t *testing.T) {
+	testtype.SkipUnlessTestType(t, testtype.UnitTestType)
+
+	oldMargin := 100
+	overhead := 82 + len("testdb") + len("large_documents")
+
+	assert.Greater(t, overhead, oldMargin, "old margin was too small")
+	assert.Less(t, overhead, 1024, "new margin handles it fine")
```
utafrali: This bound is stale. The new margin in production code is `1024*1024` (1MB), but this assertion checks `overhead < 1024` (1KB), which was the intermediate value from an earlier commit before being bumped to 1MB. The test still passes because 103 < 1024, but the message "new margin handles it fine" is misleading — it's validating against the wrong threshold. Fix:
```go
assert.Less(t, overhead, 1024*1024, "new margin of 1MB handles it fine")
```

common/db/buffered_bulk_test.go
```diff
@@ -16,6 +16,54 @@ import (
...
+			"a_database_with_a_very_long_name_for_testing",
+			"a_collection_name_that_is_also_extremely_long_to_test_the_overhead_boundaries",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			fixedOverhead := 82
+			totalOverhead := fixedOverhead + len(tc.db) + len(tc.collection)
+			margin := 1024 * 1024 // 1MB
```
utafrali: The margin is hardcoded as `1024 * 1024` in both test functions rather than being derived from the production constant. If someone changes the margin in `buffered_bulk.go`, these tests will still pass without flagging the discrepancy. Consider exporting the constant:
```go
// In buffered_bulk.go
const OPMSGOverheadMargin = 1024 * 1024

// byteLimit: MAX_MESSAGE_SIZE_BYTES - OPMSGOverheadMargin
```
Then reference `OPMSGOverheadMargin` in the test so a change to the production value is immediately reflected.

common/db/buffered_bulk_test.go
```diff
@@ -16,6 +16,54 @@ import (
...
+		{
+			"very long namespace",
+			"a_database_with_a_very_long_name_for_testing",
+			"a_collection_name_that_is_also_extremely_long_to_test_the_overhead_boundaries",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			fixedOverhead := 82
```
utafrali: The `fixedOverhead := 82` figure is an unverified estimate with no derivation shown. The tests assert that the math holds, but they never actually construct a `BufferedBulkInserter` and check that its `byteLimit` field is set correctly. A test like:
```go
bb := newBufferedBulkInserter(coll, 100, version, false)
assert.Equal(t, MAX_MESSAGE_SIZE_BYTES - OPMSGOverheadMargin, bb.byteLimit)
```
would directly guard against the production value drifting.

common/db/buffered_bulk.go
```diff
@@ -75,10 +75,8 @@ func newBufferedBulkInserter(
 		collection:    collection,
 		bulkWriteOpts: bulkOpts,
 		docLimit:      docLimit,
-		// We set the byte limit to be slightly lower than maxMessageSizeBytes so it can fit in one OP_MSG.
-		// This may not always be perfect, e.g. we don't count update selectors in byte totals, but it should
-		// be good enough to keep memory consumption in check.
-		byteLimit:          MAX_MESSAGE_SIZE_BYTES - 100,
+		// Leave room for OP_MSG wire overhead (header, command body, namespace, etc).
+		byteLimit:          MAX_MESSAGE_SIZE_BYTES - 1024*1024,
```
utafrali, line 81: The 1MB margin is not quantitatively justified in the comment. The actual maximum OP_MSG overhead is ~82 bytes fixed + MongoDB's 120-char namespace limit = ~202 bytes. 1MB is a 5000x safety factor over the real overhead. The original PR #981 used 1KB which already provided ample headroom. If the intent is to be conservative, the comment should say so explicitly and explain the reasoning, for example:
```go
// Leave room for OP_MSG wire overhead (header, command body, namespace, etc).
// Real overhead is at most ~200 bytes, but we use 1MB for a wide safety margin
// since we also do not count update selectors toward byteCount.
byteLimit: MAX_MESSAGE_SIZE_BYTES - OPMSGOverheadMargin,
```
Note that the original comment's caveat about update selectors not being counted in byte totals was useful documentation and should be preserved.

common/db/buffered_bulk_test.go
tdq45gj: Thanks for your reviews! I replaced the unit tests are making assumptions on message constructions to have a fixed overhead of 82 bytes. I replaced those with an integration test.

common/db/buffered_bulk.go
tdq45gj: Added a comment.

autarch (approved):
LGTM!

PR #993, opened 2026-04-17
Fix test failure on v4.2

tdq45gj:
This commit fixes `TestBufferedBulkInserterInserts/full_buffer_with_max-length_namespace` failure on v4.2. The fix checks server version and set namespace length to 120 for v4.2.


autarch (approved):
LGTM!

PR #994, opened 2026-04-21
TOOLS-4189 release: fix artifact download for 8.2

mmcclimon:
For whatever reason, there are two git hashes for the commit "add release config for 8.2.7"; this means that we try to construct an evergreen URL with two hashes joined by a newline. This causes url.Parse to error with "invalid control character in URL," meaning that all the 8.2 tests are failing on the waterfall.


autarch (approved):
LGTM!

PR #995, opened 2026-04-22
Update expired test certs

mmcclimon:
The QA tests were all failing to start mongod because the certs expired 2026-04-18T16:11:37.000+00:00, which is now in the past! Fix this by regenerating the certs that were expired.

For posterity, this was done by running, in the server repo:

```
poetry run python jstests/ssl/x509/mkcert.py --config jstests/ssl/x509/certs.yml
```

and then copying the resulting files to the correct locations in the tools repo.


mmcclimon:
The 8.2 tests are all failing here; the fix for that is coming later today.

autarch (approved):
LGTM!

PR #996, opened 2026-04-23
TOOLS-4184 Use simple MarshalExtJSON in mongoexport

tdq45gj:
This commit reverts a change made in TOOLS-3611 for mongoexport. Mongoexport will run the simple `MarshalExtJSON` without validation for BSON-JSON-BSON round tripping. This will result in faster export but might produce JSON files that are not restorable by mongoimport.

### DIFF
--- a/mongoexport/json.go
+++ b/mongoexport/json.go
@@ -10,7 +10,6 @@ import (
 	"bytes"
 	"io"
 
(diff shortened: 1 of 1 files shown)


FGasper (approved):
LGTM. Thanks!

PR #1010, opened 2026-07-01
timeseries tests: skip for now pending TOOLS-4228

mmcclimon:
Recent CVE fixes in the server have broken these tests; it's not clear if the server needs to fix these or we do, but they're just causing noise in the meantime.

Also: unbreak a different timeseries test that was broken by the same strict validation. We were unmarshaling a JSON blob into a map and then inserting it, but doing so can affect the order, which stricter validation rejected.


mmcclimon:
@autarch I'm happy to put this in two separate PRs if you'd rather.

autarch (approved):
LGTM!

PR #1012, opened 2026-07-06
Tools 4089 server 9.0 support

mattb-mdb:
Adds mongo-tools support for MongoDB 9.0, including CI coverage, build/download plumbing, and fixes for 9.0 behavior changes around timeseries collections and index metadata.

### Fixes for 9.0 behavior changes
- **mongorestore**: 9.0 reports collation: {locale: "simple"} on indexes that were created without an explicit collation, and rejects a v:1 index that carries any collation. mongorestore now strips this redundant simple collation for v:1 indexes only, so v:1 indexes dumped from 9.0+ restore correctly (v≥2 collations are preserved).
- **mongorestore (timeseries)**: demultiplex archives using the source server version from the prelude, so pre-8.3 system.buckets and 8.3+ viewless-timeseries archives land on the correct namespace; skip unsupported legacy system.buckets oplog replay on viewless-timeseries servers.
- **mongoexport**: reject system.buckets export before the existence check on 8.3+.
### Build & release
- Raise max supported server version to 9.0; fall back to the requested version when the downloads feed lacks it (unreleased alphas).
- Add debian13 and suse16 platforms.
### CI
- Add the .9.0 task suite (pinned to 9.0.0-alpha2) and debian13/suse16 variants; bump max_server_version; add mise/go install steps.
- Vendor 9.0 ShardingTest/ReplSetTest shell libs (the 8.1 libs can't run a 9.0 cluster), preserving the existing minimal-vendoring pattern and the mongo-tools skipValidation hook.
- Verified green across rhel88, rhel88-race, and windows; the mongorestore collation fix was also reproduced and confirmed against a real 9.0.0-alpha2 server.


mongorestore/restore.go
```diff
@@ -523,6 +529,56 @@ func (restore *MongoRestore) convertLegacyIndexes(
...
+		return v == 1
+	case int32:
+		return v == 1
+	case int64:
+		return v == 1
+	case float64:
+		return v == 1
+	}
+	return false
+}
```
autarch, line 566: I think you can use the `safecast` package to do this a bit more simply. Just safecast it to an int, I think.

autarch (approved):
LGTM % one small thing

PR #1069, opened 2026-08-05
TOOLS-4294 linuxRelease: fix push task for weird architectures

mmcclimon:
This was broken by 9d47eccb (TOOLS-4102 Use `mise` to execute `go` as well as other dev tools, 2026-07-07); on ppc64le and s390x architectures, we don't use `mise`, and so the "dist" task name on those platforms is actually "dist-without-mise". The code in the `push` task was looking for the hard-coded "dist" name, and so was failing with the new name.


autarch (approved):
LGTM!

PR #1071, opened 2026-08-05
TOOLS-4296 Annotate HTTP and S3 errors when downloading a server package

mattb-mdb:
downloadBinaries did not check the response status, so an S3 error document was written to disk as if it were the server archive. The failure only surfaced much later, when extracting the "archive" failed, and said nothing about why the download had not worked.

This change checks the status up front and report what came back. S3 serves a small XML document describing the error, so parse that and include its code, message and request ID. When the code is one that means the request itself was rejected, say that the credentials the presigned URL was signed with are at fault rather than the build, which is what an Evergreen artifact URL looks like after an AWS credential rotation.

### DIFF
--- a/release/release.go
+++ b/release/release.go
@@ -10,6 +10,8 @@ import (
 	"crypto/sha1"
 	"crypto/sha256"
 	"encoding/json"
+	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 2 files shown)


release/release.go
```diff
@@ -1498,6 +1500,58 @@ func downloadMongodAndShell(v string) {
...
+	// An error body is a short XML or HTML document. Anything longer is not something we
+	// want in a log line, and we already have the status to report.
+	body, err := io.ReadAll(io.LimitReader(res.Body, 4096))
+	if err != nil {
+		return fmt.Errorf("got %s and could not read the response body: %w", res.Status, err)
+	}
+
+	var s3Err s3ErrorResponse
+	if xml.Unmarshal(body, &s3Err) != nil || s3Err.Code == "" {
+		return fmt.Errorf("got %s: %s", res.Status, strings.TrimSpace(string(body)))
```
autarch, line 1540: A couple things here. First, I think we should separate the two checks into two errors (so two `if` blocks). Second, we should avoid using a colon in error messages, _except_ when we're wrapping another error.

release/release.go
```diff
@@ -1498,6 +1500,58 @@ func downloadMongodAndShell(v string) {
...
+	}
+
+	msg := fmt.Sprintf("got %s from S3: %s: %s", res.Status, s3Err.Code, s3Err.Message)
```
autarch, line 1543: Ditto re: colons

autarch (approved):
LGTM % nits re: error checks and messages.

PR #1072, opened 2026-08-05
TOOLS-4293 Restart mongod in oplog_replay_local_rs.js so the fake oplog is visible

mattb-mdb:
This test fakes a replica set oplog by inserting directly into local.oplog.rs on a standalone mongod. The server sets the timestamp up to which the oplog is visible at startup, and those direct inserts never advance it. As of MongoDB 9.0 a forward scan of the oplog stops at that timestamp, so mongodump saw an empty oplog and the test failed.

Restarting the mongod makes the server set the visibility timestamp from the last entry in the oplog. The restart reuses the same port and dbpath, so the fake oplog survives it, and the connections the test opened beforehand are re-established afterwards.

The restart goes through ToolTest.stop() and startDB(): the suites that run this test get ToolTest from the mongo shell rather than from jstests/libs/servers_misc.js, and MongoRunner.stopMongod cannot be used because toolTest.m is a connection with no 'port' property.

### DIFF
--- a/test/qa-tests/jstests/restore/oplog_replay_local_rs.js
+++ b/test/qa-tests/jstests/restore/oplog_replay_local_rs.js
@@ -36,6 +36,26 @@
     assert.eq(1, r.nInserted, "insert failed");
   }
 
+  // The server sets the timestamp up to which the oplog is visible when it starts up,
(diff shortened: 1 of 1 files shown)


autarch (approved):
LGTM!

FGasper (approved):
LGTM

PR #1073, opened 2026-08-05
TOOLS-4263 Convert mongorestore collection-options and metadata tests to Go

autarch:
Adds `integration/dumprestore/options_roundtrip_test.go`: covering how mongorestore handles collection options and dump metadata.

JS -> Go mapping (all Go tests live in `integration/dumprestore/options_roundtrip_test.go`):

- `test/legacy42/jstests/tool/dumprestore8.js` -> `TestCappedCollectionRoundTrip` (`integration/dumprestore/options_roundtrip_test.go`) - capped collection options and secondary indexes survive a dump/restore round trip. The fixture is a plain collection with two secondary indexes plus a capped collection that has already evicted documents, and it is restored four ways to cover each granularity: from the dump root, from a single database directory into a new database name, with `--nsInclude db.*`, and from a single BSON file into a new collection name. Each case checks document counts, index names, the capped options, and, most importantly, that the restored collection still evicts, by inserting past the cap and asserting the count does not grow.
- `test/qa-tests/jstests/restore/extended_json_metadata.js` -> `TestRestoreExtendedJSONMetadata` (`integration/dumprestore/options_roundtrip_test.go`) - metadata written by very old versions of mongodump used legacy extended JSON types for numeric values, so the test hand-writes metadata in that shape, with `{"$numberLong": ...}` for the capped `size` and for an index's `key`, `min`, `max`, and `bits`, plus `v: 1` indexes and an `ns` field, then asserts the capped options are restored and the index keys decode to plain values (`_id -> 1`, `pos -> "2d"`) rather than to documents or strings. Modern mongodump cannot produce such a dump, so the fixture is built at runtime instead of being checked in, so now-unused `testdata/dump_extended_json_options` fixture is deleted.
- `test/qa-tests/jstests/restore/no_options_restore.js` -> `TestNoOptionsRestore` (`integration/dumprestore/options_roundtrip_test.go`) - a default restore preserves collection options while `--noOptionsRestore` discards them, exercised against a fixture of one capped collection, one collection with a validator, and one collection with no options, all holding the same documents. In the stripped cases the restored collections are compared against the options an actual no-options collection reports on the same server, rather than against a hardcoded empty document, because servers add default option fields over time.
[...]

PR #1074, opened 2026-08-06
TOOLS-4263 Convert mongorestore index round-trip tests to Go

autarch:
Adds `integration/dumprestore/index_roundtrip_test.go`: covering how mongorestore round-trips index specifications.

JS -> Go mapping (all Go tests live in `integration/dumprestore/index_roundtrip_test.go`):

- `test/qa-tests/jstests/restore/indexes.js` -> `TestIndexRoundTrip` (`integration/dumprestore/index_roundtrip_test.go`) - one collection carrying every index type the JS covered (simple, sparse and unique, compound, compound with int64 key values, multikey via array data, text with a non-default `default_language`, and 2dsphere) survives a dump/restore round trip, restored both from the dump directory and from the BSON file directly. The JS only compared index counts, so a mangled text-index weight or a dropped sparse flag would have gone unnoticed; this compares the complete spec of every index before and after, with fixture preconditions asserting the int64 key types and the Spanish text language really landed so that we're testing what we think we're testing. Note that the JS passed `{language: 'spanish'}`, which is not the `createIndex` option name and was therefore ignored by the server, so the Spanish text index is new coverage.
- `test/legacy42/jstests/tool/dumprestore6.js` -> `TestRestoreLegacySystemIndexes` (`integration/dumprestore/index_roundtrip_test.go`) - a pre-2.6 dump stores index specs in a `system.indexes.bson` file rather than a per-collection metadata file, and mongorestore falls back to that file when a database directory has no metadata. The test builds such a dump at runtime, deliberately naming a different database in the specs' `ns` fields than the restore target so that mongorestore has to take the target from `--db`, then asserts the document and both indexes are created and the legacy v1 specs are converted to the current index version.
- `test/qa-tests/jstests/restore/index_version_roundtrip.js` -> `TestIndexVersionRoundTrip` (`integration/dumprestore/index_roundtrip_test.go`) - `--keepIndexVersion` restores each `_id` index at its original version, while a restore without the flag gets the server's default version instead. Coverage is limited to `_id` indexes because the version can only be set through the create command's `idIndex` argument: a v1 secondary index is silently upgraded to the current version at creation on supported servers, so there is no secondary index version to round-trip. The secondary indexes are still created and asserted to survive both restores, along with the document counts.
- `test/qa-tests/jstests/restore/ordered_partial_index.js` -> `TestOrderedPartialIndex` (`integration/dumprestore/index_roundtrip_test.go`) - an index whose `partialFilterExpression` has ten fields keeps that document's field order through the metadata file, since reordering it would produce an index the server treats as different from the one that was dumped. The fields are written in a deliberately unsorted order and the restored spec is read back as a `bson.D`, so both a sorted reordering and a randomized one are detectable; an ascending field list would have matched its own most likely corruption.


autarch:
* **#1143** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1143?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1144](https://github.com/mongodb/mongo-tools/pull/1144) <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1144?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1157](https://github.com/mongodb/mongo-tools/pull/1157) <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1157?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1142** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1142?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1141** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1141?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1140** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1140?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1139** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1139?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1138** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1138?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1137** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1136** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1135** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1135?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1134** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1134?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1133** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1133?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1132** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1132?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1131** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1131?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1130** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1130?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1129** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1129?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1104** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1104?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1102** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1102?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1098** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1098?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1097** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1097?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1096** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1093** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1093?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1091** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1091?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1074** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1074?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1073** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1073?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1148** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1148?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

integration/dumprestore/index_roundtrip_test.go
```diff
@@ -0,0 +1,468 @@
...
+		{
+			Keys:    bson.D{{"textField", "text"}},
+			Options: options.Index().SetDefaultLanguage("spanish"),
+		},
+		{Keys: bson.D{{"geoField", "2dsphere"}}},
+	})
+	s.Require().NoError(err, "can create the fixture indexes")
+
+	docs := make([]any, 0, indexFixtureDocCount)
+	for i := range 5 {
```
mmcclimon: This feels very strange to me, to have a loop that inserts 3 docs each time. I spent a long time trying to figure out why `5` was different than `indexFixtureDocCount`, because I read the append statement as inserting one deeply-nested doc and not 3 individual docs.

integration/dumprestore/index_roundtrip_test.go
```diff
@@ -0,0 +1,468 @@
...
+	s.Require().EqualValues(
+		indexFixtureDocCount,
+		s.docCount(coll),
+		"the fixture inserted every document",
+	)
+	specs := s.indexSpecs(coll)
+	s.Require().Len(
+		specs,
+		indexFixtureIndexCount,
+		"the fixture created every index, so comparing specs is not vacuous",
```
mmcclimon: phew, I'd hate for the test to be vacuous. (What a weird phrasing.)

integration/dumprestore/index_roundtrip_test.go
```diff
@@ -0,0 +1,468 @@
...
+// directory or the collection's bson file directly.
+func (s *DumpRestoreSuite) TestIndexRoundTrip() {
+	s.Run("restore from the dump directory", s.testIndexRoundTripFromDumpDir)
+	s.Run("restore from the bson file", s.testIndexRoundTripFromBSONFile)
+}
+
+func (s *DumpRestoreSuite) testIndexRoundTripFromDumpDir() {
+	testDB := s.database("indexes_dump_dir")
+	coll := s.createIndexFixture(testDB)
+	specsBefore := s.indexSpecs(coll)
```
mmcclimon, line 24: (nit) The placement of helper functions in this file doesn't match our usual style; `s.createIndexFixture` is just below, but `s.indexSpecs` is way at the bottom of the file.

integration/dumprestore/index_roundtrip_test.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	s.Assert().Len(specsAfter, len(specsBefore), "no index is restored that was not dumped")
+}
+
+func (s *DumpRestoreSuite) indexSpecsByName(specs []bson.D) map[string]bson.D {
+	specsByName := make(map[string]bson.D, len(specs))
+	for _, spec := range specs {
+		name, ok := optionValue(spec, "name").(string)
+		s.Require().True(ok, "every index spec reports a name")
```
mmcclimon: This is a weird assertion message; this assertion is only testing that this single spec reports a name.

integration/dumprestore/index_roundtrip_test.go
```diff
@@ -0,0 +1,468 @@
...
+	{"v1coll", 1},
+	{"v2coll", 2},
+}
+
+// createIDIndexVersionFixture creates two collections whose _id indexes differ
+// only in their index version. The version can only be set through the create
+// command's idIndex argument: a v1 secondary index is silently upgraded to the
+// server's current version, so there is no secondary index version to
+// round-trip. The secondary index is still created, so the tests can check that
+// restoring does not lose it.
```
mmcclimon: Is that like, true? The JS test creates the other indexes with a `v` field as well. Maybe that doesn't actually do anything, but it seems like we're missing coverage here; even if we do create the other index with `v:1` and it's auto-upgraded on create, we should dump/restore it correctly.

integration/dumprestore/index_roundtrip_test.go
```diff
@@ -0,0 +1,468 @@
...
+	s.Assert().Equal(
+		map[string]int32{"v1coll": 2, "v2coll": 2},
+		s.idIndexVersions(testDB),
+		"without --keepIndexVersion every _id index gets the server's default version",
+	)
+}
+
+// testLegacySystemIndexes restores a dump in the pre-2.6 layout, where index
+// specs live in a system.indexes.bson file instead of a per-collection metadata
+// file. mongorestore falls back to that file when a database directory has no
```
mmcclimon: Do we even still need this? We don't support cross-version dump/restore, and _definitely_ not from v2.6!

But it's fine to keep this test I guess, but we should consider removing the code that implements this stuff.

integration/dumprestore/index_roundtrip_test.go
```diff
@@ -0,0 +1,468 @@
...
+	return specsByName
+}
+
+// TestIndexVersionRoundTrip checks which version a restored index ends up at:
+// --keepIndexVersion carries the dumped version through, without it the server
+// applies its own default, and a legacy version in the dump is converted.
+func (s *DumpRestoreSuite) TestIndexVersionRoundTrip() {
+	s.Run("keepIndexVersion preserves the index version", s.testKeepIndexVersion)
+	s.Run("the server default version is used otherwise", s.testDefaultIndexVersion)
+	s.Run("a legacy system.indexes dump is converted", s.testLegacySystemIndexes)
```
mmcclimon: The commit message says that this is a top-level test, but it's actually a subtest of TestIndexVersionRoundtrip. Probably it should be a top-level test. (Probably _all_ of these could just be top-level tests, because they get nothing out of being subtests, but whatever.)

mmcclimon (approved):
LGTM % some things you can take or leave (all the same flavor of comments I made on the previous one).

integration/dumprestore/index_roundtrip_test.go
autarch: Lol, fixed.

integration/dumprestore/index_roundtrip_test.go
autarch: I think this is clearer now.

integration/dumprestore/index_roundtrip_test.go
autarch: Fixed.

integration/dumprestore/index_roundtrip_test.go
autarch: Fixed

integration/dumprestore/index_roundtrip_test.go
autarch: We're not 100% consistent about putting stuff right after first use or at the end when it's used in multiple places. I kind of prefer to have things that are called in many places at the end so it's after the _last_ caller rather than the first. But this was fixed, regardless.

integration/dumprestore/index_roundtrip_test.go
autarch: I made a new ticket to remove this legacy support stuff - [https://jira.mongodb.org/browse/TOOLS-4350](https://jira.mongodb.org/browse/TOOLS-4350). So for now I'll merge this test but we can remove it later as part of that work.

integration/dumprestore/index_roundtrip_test.go
autarch: Fixed by having it actually create `v:1` secondary indexes, which are still supported even in 9.0!

PR #1094, opened 2026-08-07
TOOLS-4309 Fail mongodump --oplog on the upgradeDowngradeViewlessTimeseries entry

mattb-mdb:
`mongorestore` cannot replay the `upgradeDowngradeViewlessTimeseries` entry — see #1095, which landed that interception — so an oplog window spanning a viewless timeseries conversion is not restorable. There is no point dumping one.

## Changes

`mongodump/oplog_dump.go` — reject the entry in `oplogDocumentValidator`, alongside the existing `renameCollection`, `importCollection`, and resharding checks, so the failure happens at dump time with a message that names the cause rather than at restore time.

This is partly redundant with the `admin.system.version` check, since `setFCV` writes that namespace: a dump window containing the whole FCV transition already fails today, just with a confusing message about `admin.system.version`. The new check covers a window that falls between the transitional and final FCV writes — plausible on a deployment with many timeseries collections, where the conversion phase is long.

The string literal is used rather than exporting `upgradeDowngradeViewlessTimeseriesCmd` from `mongorestore`, both to match the surrounding style and to avoid moving it into `common`, which other MongoDB projects consume.

## Tests

`TestErrorOnViewlessTimeseriesConversion` (`mongodump/oplog_dump_test.go`) — feeds a synthetic conversion entry to the validator and asserts the specific error. Synthetic, so it runs as a unit test on any server version.

### DIFF
--- a/mongodump/mongodump_test.go
+++ b/mongodump/mongodump_test.go
@@ -22,6 +22,7 @@ import (
 	"regexp"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/mongodb/mongo-tools/common"
 	"github.com/mongodb/mongo-tools/common/archive"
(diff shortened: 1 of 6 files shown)


mongodump/mongodump_test.go
```diff
@@ -2562,3 +2563,228 @@ func TestBrokenPipe(t *testing.T) {
...
+
+	err = <-dumpErr
+
+	// A failed setFCV means the conversion never happened, so there is nothing to assert.
+	require.NoError(t, setFCVErr, "setFCV should succeed while a dump is running")
+
+	if err != nil {
+		// The server interrupts reads whose collection is converted underneath them, and
+		// mongodump surfaces that as a dump failure. Losing the race is the expected outcome;
+		// what matters is that it is reported rather than silently truncating the dump.
```
tdq45gj, line 2650: If reads for that collection are done and then the time-series upgrade/downgrade oplog appeared, would mongodump create a dump containing the upgrade/downgrade oplog entry that is corrupted (not restorable by mongorestore)?

tdq45gj (commented):
The changes look good to me. I just have one question about a scenario that might cause corrupted dump files.

mongodump/mongodump_test.go
mattb-mdb: The dump should error out as above but using --oplogLimit should allow handling of the special case unless I am mistaken?

tdq45gj: Yeah they might be able to skip that oplog in restore. I think my point is - could we add a check in mongodump to return an error assuming it's straightforward?

tdq45gj (approved):
LGTM, thanks!

PR #1099, opened 2026-08-10
TOOLS-4312 Hint $natural on unfiltered mongodump queries

mattb-mdb:
SERVER-127688 (9.0) adds a maxEstimatedScanBytes server parameter that rejects queries at plan time when the plan requires an unbounded COLLSCAN on a collection larger than the configured threshold. An unfiltered mongodump query is exactly that shape, so with the knob enabled mongodump cannot dump any collection over the threshold.

An explicit $natural hint tells the server the full scan is intentional, which exempts the query from the rejection. It does not change the plan the server would have chosen for an unfiltered query, and it is metered separately so the operator keeps visibility into large scans. We must not send it when there is a filter, which could otherwise be served by an index.

The pre-dump count needs the same treatment: DeferredQuery.Count falls through to CountDocuments for views, and aggregations are rejected by the knob too.

### DIFF
--- a/common/db/query.go
+++ b/common/db/query.go
@@ -39,6 +39,9 @@ func (q *DeferredQuery) Count(isView bool) (int, error) {
 	}
 
 	opt := mopt.Count()
+	if q.Hint != nil {
+		opt.SetHint(q.Hint)
+	}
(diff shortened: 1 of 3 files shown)


autarch (approved):
LGTM!

PR #1100, opened 2026-08-10
TOOLS-4313 Hint $natural on unfiltered, unsorted mongoexport queries

mattb-mdb:
SERVER-127688 (9.0) adds a maxEstimatedScanBytes server parameter that rejects queries at plan time when the plan requires an unbounded COLLSCAN on a collection larger than the configured threshold. An export with neither a query nor a sort is exactly that shape, so with the knob enabled mongoexport cannot export any collection over the threshold.

mongoexport builds its own find options rather than going through common/db.DeferredQuery, so the equivalent mongodump fix in TOOLS-4312 does not cover it. There is no count path to fix here: getCount only ever uses EstimatedDocumentCount, which reads metadata rather than planning a scan.

We must not send the hint when there is a query or a sort, either of which could be served by an index. That leaves a sort on an unindexed field still rejected, which is deliberate: that plan is a collection scan plus a blocking sort, which is the workload the guardrail exists to catch.

### DIFF
--- a/mongoexport/mongoexport.go
+++ b/mongoexport/mongoexport.go
@@ -365,6 +365,15 @@ func (exp *MongoExport) getCursor() (*mongo.Cursor, error) {
 		findOpts.SetProjection(makeFieldSelector(exp.OutputOpts.Fields))
 	}
 
+	// An export with neither a query nor a sort is a deliberate full collection scan. Saying so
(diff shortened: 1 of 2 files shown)


autarch (approved):
LGTM!

PR #1101, opened 2026-08-10
evergreen: stop testing on server 8.1

mmcclimon:
This was an Atlas-only rapid release, and all the Atlas clusters have since been upgraded to (at least) 8.3. Since it wasn't released publicly, and since testing against 8.1 is getting increasingly annoying (because of various infrastructure issues), let's just cut testing for that version entirely.

We could probably also remove 8.2 testing, but since that version was publicly released, doing that right now is harder to justify.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -1511,99 +1511,6 @@ tasks:
         vars:
           target: test:awsauth ${testArgs}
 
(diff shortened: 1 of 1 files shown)


autarch (approved):
LGTM!

PR #1106, opened 2026-08-12
evergreen: restore missing AWS env vars for release

mmcclimon:
These got lost in the shuffle in 96909627 (TOOLS-4238 Move more inline shell code from the Evergreen config to scripts), 2026-07-16) and 926381b0 (TOOLS-4238 Finish moving inline shell that uses `_set_shell_env` to scripts), 2026-07-16), and we need them for the push task to work correctly on release.

### DIFF
--- a/common.yml
+++ b/common.yml
@@ -512,6 +512,8 @@ functions:
         args:
           - "./scripts/upload-release.sh"
         env:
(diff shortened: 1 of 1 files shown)


mattb-mdb (approved):
LGTM!

PR #1109, opened 2026-08-12
platform: tweak linux versions

mmcclimon:
We don't want to release for 9.0 for Ubuntu 20.04 (it's EOL), or 7.0 for Ubuntu 26.04. Also, SLES 16 support should be 9.0-only for now.


autarch (approved):
LGTM!

PR #1112, opened 2026-08-12
platform: fix additional failing platforms

mmcclimon:
In the previous commit I missed ubuntu 20.04 for x86, and we also need to special-case 8.2, which is EOL and so we can't push to those versions.


autarch (approved):
LGTM!

drichmdb (approved):

PR #1113, opened 2026-08-12
release: fix server 8.2 skip

mmcclimon:
This was clearly wrong the first time because it would also skip 8.0; oops! The right fix here is to remove 8.2 from linuxRepoVersionsStable.


mattb-mdb (approved):
LGTM

PR #1114, opened 2026-08-12
TOOLS-4314 100.18.0 Release CHANGELOG.md

mattb-mdb:
This PR contains the release notes for the 100.18.0 release changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,102 @@
 # Database Tools Changelog
 
+## 100.18.0
+
+_Released 2026-08-12_
+
+We are pleased to announce version 100.18.0 of the MongoDB Database Tools.
+
+The highlight for this release is MongoDB Server 9.0 support. This release also adds builds for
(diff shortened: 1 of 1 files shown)


mmcclimon (commented):
I think this is fine, but a second opinion would be good before merging, because: I wonder if some of these tasks should be marked internal-only and not be in the release notes. For example, "Use mise to manage our dev tools" probably doesn't need to be in the changelog, because it doesn't matter at all to users of the tools. Likewise, some of the bugs ("Scripts must call go via $GO_EXEC_PREFIX") doesn't actually reflect a bug in any released version of the tools, so we probably don't actually want to put it in the changelog (especially since the scripts in question are only called by internal toling in CI).

That's my own opinion though and not an official one, so I don't want to say one way the other "definitely remove them" or "definitely keep them".

rohshar (changes requested):
I agree with michael, let's remove the internal stuff that isn't useful to clients

rohshar (approved):

PR #1126, opened 2026-08-19
TOOLS-4080 Add a merge-queue variant so we can have GitHub block on this

autarch:
The previous attempt at adding merge queue support wasn't complete. We need a single name we can have GitHub block on, rather than having it block on an ever-changing list of tasks.

This is the same thing we do for Mongosync.

It also includes a test to make sure that the merge queue variant never depends on tasks that are _not_ run for a PR. If we were to let that happen, then the PR would never be mergeable.

### DIFF
--- a/evergreen/evergreen_test.go
+++ b/evergreen/evergreen_test.go
@@ -2,11 +2,12 @@ package evergreen
 
 import (
 	"os"
-	"path/filepath"
-	"runtime"
+	"os/exec"
+	"regexp"
 	"slices"
 	"testing"
 
+	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/mongodb/mongo-tools/common/testtype"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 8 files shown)


autarch:
* **#1127** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1127?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1126** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1126?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1126?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

evergreen/generator/main.go
```diff
@@ -17,11 +17,5 @@ func main() {
 		panic(err)
 	}
 
-	mq, err := c.MergeQueueAliasesYAML()
```
mmcclimon, line 20: I...did not know we had a generator for Tools evergreen config?! Fascinating. (From very early in your tenure here.)

evergreen/evergreen_test.go
```diff
@@ -18,127 +19,196 @@ type aliasEntry struct {
...
+}
+
+// sortedNames renders a set of scheduled tasks as sorted "variant/task" strings,
+// so that a failure names the offending tasks in a stable order.
+func sortedNames(tasks mapset.Set[variantTask]) []string {
+	names := make([]string, 0, tasks.Cardinality())
+	for _, vt := range tasks.ToSlice() {
+		names = append(names, vt.variant+"/"+vt.task)
+	}
+	sort.Strings(names)
```
mmcclimon: We usually use `slices.Sort` in new code.

mmcclimon (approved):
I think this is fine, though the tests aren't passing on it right now. Once they do, LGTM.

evergreen/generator/main.go
autarch: It generates one little tiny slice of the config. But working on the tools recently, I think it would benefit from switching to the full generator approach like we did for mongosync.

evergreen/evergreen_test.go
autarch: Fixed

PR #1127, opened 2026-08-19
Fix broken code for finding latest version we are testing with

autarch:
This particular code is not called right now, but it could be in the future.

### DIFF
--- a/evergreen/evergreen.go
+++ b/evergreen/evergreen.go
@@ -114,42 +114,6 @@ func (v *Variant) includesLatest() bool {
 	return false
 }
 
-var versionRE = regexp.MustCompile(`\.(\d+)\.(\d+)$`)
-
-func (v *Variant) mostRecentServerVersion() (string, error) {
-	var versions [][2]int
-	for _, t := range v.Tasks {
-		if matches := versionRE.FindStringSubmatch(t.Name); len(matches) != 0 {
-			maj, err := strconv.Atoi(matches[1])
-			if err != nil {
-				return "", err
-			}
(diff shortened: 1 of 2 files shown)


autarch:
* **#1127** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1127?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1126** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1126?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

evergreen/evergreen.go
```diff
@@ -328,3 +292,39 @@ func (c *Config) integrationTestAliases() ([]alias, error) {
...
+	sort.SliceStable(versions, func(i, j int) bool {
+		majI := versions[i][0]
+		majJ := versions[j][0]
+		minI := versions[i][1]
+		minJ := versions[j][1]
+
+		if majI != majJ {
+			return majI > majJ
+		}
+		return minI > minJ
```
autarch, line 326: I moved this func so the whole thing shows up as a diff, but the only real change is the logic on these lines to change from `<` to `>`, which reverses the sorting so the highest numbers come first.

nickweinberger (approved):
This was indeed broken...Thanks!

PR #1128, opened 2026-08-20
TOOLS-4315 Throttle inserts in TestIgnoreMongoDBInternal

tdq45gj:
This commit attempts to fix oplog rollover error caused by `TestIgnoreMongoDBInternal` in two ways: 
1) reduce size per document in the test by switching the _id type from ObjectID to int32
2) throttle the write loop by adding a 10ms delay between each insert


autarch (approved):
LGTM but I do wonder if there's a more deterministic way to write this test, probably by using a failpoint.

tdq45gj:
@autarch The last time I looked at the failpoints in the mongodump/mongorestore, they're also implemented with sleep. We need better failpoints in the tools first.

PR #1150, opened 2026-08-21
TOOLS-4338 Report the real error when the oplog overflow check fails to read

autarch:
`mongodump --oplog` could fail with "oplog overflow: mongodump was unable to capture all new oplog entries during execution" when the oplog had not overflowed at all. `checkOplogTimestampExists` returned `(false, err)` when it could not read the oplog, and `Dump` tested `!exists` before `err`, so any read failure was discarded and reported to the user as possible data loss.

The read failure seen in CI was `CappedPositionLost` (server code 136) on the check's own query against `local.oplog.rs`. Scanning the front of the oplog in `$natural` order races the server's oplog truncation: the records being read can be deleted mid-scan, which fails the whole read even though the scan had already found its single document. That says nothing about whether the dump's starting timestamp survived. The dump that produced this failure ran for 87ms against a 40MB oplog, so no real rollover was possible.

`Dump` now checks the error before the boolean at both call sites, so a read failure is reported as itself rather than as an overflow, and `findOldestOplogEntry` retries a read that lost its position up to four times with exponential backoff.

The retry cannot mask real data loss. Oplog truncation only ever removes the oldest records: if the starting timestamp is actually gone, every retry still reads an oldest timestamp newer than that starting timestamp, and the overflow is still reported. Exhausting the attempts returns an error, so the check fails closed. The cost is that each retry slightly widens the window in which truncation can legitimately advance past the starting timestamp, which can turn a good dump into a reported overflow.

The new `FailOplogCheckRead` failpoint and `Failpoint.FireOnce` inject that error once so the retry can be tested without racing real truncation.

`getCurrentOplogTime` runs the same scan with `$natural: -1`. Reading from the newest end is far less exposed to truncation, and its caller already handles its error correctly, so it is left alone.

### DIFF
--- a/mongodump/mongodump.go
+++ b/mongodump/mongodump.go
@@ -472,13 +472,13 @@ func (dump *MongoDump) Dump() (err error) {
 		// we started still exist, so we know we haven't lost data)
 		log.Logvf(log.DebugLow, "checking if oplog entry %v still exists", dump.oplogStart)
 		exists, err := dump.checkOplogTimestampExists(dump.oplogStart)
+		if err != nil {
+			return fmt.Errorf("unable to check oplog for overflow: %w", err)
+		}
 		if !exists {
 			return fmt.Errorf(
 				"oplog overflow: mongodump was unable to capture all new oplog entries during execution",
 			)
-		}
-		if err != nil {
-			return fmt.Errorf("unable to check oplog for overflow: %v", err)
 		}
 		log.Logvf(log.DebugHigh, "oplog entry %v still exists", dump.oplogStart)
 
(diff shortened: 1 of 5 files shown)


autarch:
* **#1150** <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/mongodb/mongo-tools/1150?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

mongodump/oplog_dump.go
```diff
@@ -144,6 +132,102 @@ func (dump *MongoDump) checkOplogTimestampExists(ts bson.Timestamp) (bool, error
...
+		return nil
+	}
+
+	return mongo.CommandError{
+		Code: cappedPositionLostErrCode,
+		Name: "CappedPositionLost",
+		Message: "CollectionScan died due to position in capped collection being deleted" +
+			" (injected by the FailOplogCheckRead failpoint)",
+	}
+}
```
tdq45gj, line 225: This seem to belong to a particular test. Can we make the failpoint embed a callback function or an customized error and move the testing logic to the test?

tdq45gj (commented):
This is looking good other than the testing logic in `oplog_dump.go`.

mongodump/oplog_dump.go
autarch: Done. I changed it so the failpoint can be configured to return an optional error.

mongodump/mongodump.go
```diff
@@ -472,14 +472,14 @@ func (dump *MongoDump) Dump() (err error) {
 		// we started still exist, so we know we haven't lost data)
 		log.Logvf(log.DebugLow, "checking if oplog entry %v still exists", dump.oplogStart)
 		exists, err := dump.checkOplogTimestampExists(dump.oplogStart)
+		if err != nil {
+			return fmt.Errorf("unable to check oplog for overflow: %w", err)
+		}
```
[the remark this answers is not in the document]
autarch: Done

tdq45gj (approved):
LGTM

PR #1179, opened 2026-08-28
TOOLS-4352 Tighten validation of special, dump-internal collections

FGasper:


FGasper:
auggie review

autarch:
@FGasper I don't really understand exactly what this is doing. Can you add a PR description explaining the issue and how this fixes it?

autarch (changes requested):
I need some more context.

PR #1197, opened 2026-09-09
TOOLS-4368 passthrough config: bump mongosync pin

mmcclimon:
This includes some changes from a few months ago that never made it into this pin (oops), and gets the tests running on 8.3, along with some other cleanups. We're going to start working on this code more seriously, so this commit is just bringing it up to the current state before that work starts in earnest.


mmcclimon:
Passthrough patch on this commit: https://spruce.corp.mongodb.com/version/6aa16a336066e4000703e7a5/tasks

autarch (approved):
LGTM!!

